Issue #14, opened 2022-09-01
Uneccessary authorization attribute on RedirectController


DanRJ:
### Description of the bug

Anonymous apps are unable to use the returnUrl query parameters because RedirectController uses the [Authorize] attribute.
Users who click the X button will be redirected to login page instead of the application intended URL. 

I suggest we remove the [Authorize] attribute considering this should be available for anon users aswell

### Steps To Reproduce

1. Use returnUrl query parameter on a anonymous app
2. Click the X button
3. Get sent to login page instead of the URL defined in returnUrl query parameter

PR #15, opened 2022-09-01
Remove authorization on RedirectController endpoint

DanRJ:
Remove the Authorize attribute from RedirectController

## Description

Removing the Authorize attribute from RedirectController so anonymous Altinn 3 apps are able to also use the returnUrl query parameter.

## Related Issue(s)
- #14 

## Verification
- [ ] All tests run green

### DIFF
--- a/src/Altinn.App.Api/Controllers/RedirectController.cs
+++ b/src/Altinn.App.Api/Controllers/RedirectController.cs
@@ -16,6 +16,7 @@ namespace Altinn.App.Api.Controllers
     /// <summary>
     /// Controller for redirect and validation of URL
     /// </summary>
+    [AllowAnonymous]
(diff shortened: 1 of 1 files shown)


src/Altinn.App.Api/Controllers/RedirectController.cs
```diff
@@ -16,7 +16,6 @@ namespace Altinn.App.Api.Controllers
     /// <summary>
     /// Controller for redirect and validation of URL
     /// </summary>
```
tjololo, line 18: We should add [AllowAnonymous] to leave no doubt that this Controller is open
```suggestion
    /// </summary>
    [AllowAnonymous]
```

tjololo (changes requested):

tjololo (approved):
:godmode:

Issue #21, opened 2022-09-20
Generation of PDF fails when using dynamic options in repeating groups


lorang92:
### Description of the bug

Adding [dynamic options for repeating groups](https://docs.altinn.studio/app/development/data/options/#pass-query-parameters-when-fetching-options) fails when generating PDF.

This is due to our index-indicators are not "initialized" in the pdf-generator. Need to:
1. Find out how many indexes there are for the given group
2. "new up" options for each index

Reported in [slack (internal link)](https://altinndevops.slack.com/archives/CDU1S3NLW/p1663304818834969). Related stack trace:

```
fail: Altinn.App.Api.Controllers.ProcessController[0]
      Process next failed.
      Newtonsoft.Json.JsonException: Unexpected character while parsing path indexer: {
         at Newtonsoft.Json.Linq.JsonPath.JPath.ParseArrayIndexer(Char indexerCloseChar)
         at Newtonsoft.Json.Linq.JsonPath.JPath.ParseIndexer(Char indexerOpenChar, Boolean scan)
         at Newtonsoft.Json.Linq.JsonPath.JPath.ParsePath(List`1 filters, Int32 currentPartStartIndex, Boolean query)
         at Newtonsoft.Json.Linq.JsonPath.JPath.ParseMain()
         at Newtonsoft.Json.Linq.JsonPath.JPath..ctor(String expression)
         at Newtonsoft.Json.Linq.JToken.SelectToken(String path, JsonSelectSettings settings)
         at Altinn.App.Core.Features.Pdf.PdfService.GetComponentKeyValuePairs(JToken component, Object data)
         at Altinn.App.Core.Features.Pdf.PdfService.GetOptionsDictionary(String formLayout, String language, Object data, String instanceId)
         at Altinn.App.Core.Features.Pdf.PdfService.GenerateAndStoreReceiptPDF(Instance instance, String taskId, DataElement dataElement, Type dataElementModelType)
         at Altinn.App.Core.Implementation.DefaultTaskEvents.OnEndProcessTask(String endEvent, Instance instance)
```

### Steps To Reproduce

1. Add dynamic options for a repeating group
2. Try submitting app to start pdf generation
3. See error.

tjololo:
I have a fix for this ready for v7, just need to discuss how we merge this into main. Should be ready for review and testing tomorrow.

PR #22, opened 2022-09-27
Bugfix/21 pdfservice support repeatinggroups

tjololo:
## Description
Having dynamic options inside a repeating group caused the PDFService to fail.
This bugfix handles the replacement of index placeholders.

## Related Issue(s)
- #21 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Pdf/PdfOptionsMapping.cs
+++ b/src/Altinn.App.Core/Internal/Pdf/PdfOptionsMapping.cs
@@ -4,15 +4,24 @@ using Newtonsoft.Json.Linq;
 
 namespace Altinn.App.Core.Internal.Pdf;
 
+/// <summary>
+/// Default implementation of IPdfOptionsMapping
+/// </summary>
 public class PdfOptionsMapping: IPdfOptionsMapping
 {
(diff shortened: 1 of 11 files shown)


tjololo:
The code is not something I'm proud of and the test is really cryptic.
The test taks a formlayout from my test app and data and runs it through the mapping logic

src/Altinn.App.Core/Internal/Pdf/PdfOptionsMapping.cs
```diff
@@ -0,0 +1,178 @@
+using Altinn.App.Core.Features.Options;
+using Altinn.App.Core.Models;
+using Newtonsoft.Json.Linq;
+
+namespace Altinn.App.Core.Internal.Pdf;
+
+/// <summary>
+/// Default implementation of IPdfOptionsMapping
+/// </summary>
+public class PdfOptionsMapping: IPdfOptionsMapping
```
RonnyB71, line 10: Just a thought: Moving forward we could consider creating a set of domain classes representing the various config files we have as this kind of reading is "business logic".

RonnyB71 (approved):
🕹️

PR #17, opened 2022-09-09
Backend dynamics

ivarne:
I find pull requests better for discussion than a branch, so I open this early.  This is the backend implementation of the new Layout Expressions that are parallelly implemented in https://github.com/Altinn/app-frontend-react/tree/feat/layout-expressions

## Description
* Tools for working with Layout files in Json with C#
* Logic for evaluating expressions
* Common test suite with frontend. Tests are specified in a custom Json format. A test adapter is maintained in each repo.

## Related issues
- #9 (this implementation)
- Altinn/app-frontend-react/issues/355 (frontend implementation)

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #23, opened 2022-09-27
Backend dynamics2

### LINKED ISSUE #9: Layout expressions for hidden/required fields
### Description

In the layout expressions project, we are adding support for defining dynamic behaviour in form layouts. This includes the new `hidden` property for components, which can be used by app developers to define an expression to decide whether a component should be hidden or visible. Expressions can also be defined for the `required` property for a component.

This expression should also be read and evaluated on the backend, in order to:
- [ ] Iterate the layout, resolving components to their respective locations in the data model (via their bindings), supporting multiple instances of a single component (i.e. in repeating groups and nested repeating groups). 
- [ ] Clean submitted data, removing values for components that are hidden
- [ ] Simplify validation for required fields. A field can be marked as required, but if it's hidden, validation for this field should be skipped.

### Additional Information

- [The project workspace](https://github.com/orgs/Altinn/projects/22/)
- Altinn/app-frontend-react/issues/355 - frontend issue

ivarne:
Reopening #17 that was closed because it targeted a branch that was deleted.

## Related Issue(s)
- closes #9

## Verification
- [ ] All tests run green

## Documentation


RonnyB71:
After we had the first review we have moved the internal feature implementations to the Internal folder while the external features are in the Features folder. We should move the Expressions folder into the Internal folder to make it clear it's internal.

ivarne:
I’m not completely convinced, at least not if you think this should be marked `internal` instead of `public`. Seems like this would be really useful to have available in apps and other code, and because of the json in apps, it needs to be very stable.

src/Altinn.App.Core/Expressions/LayoutEvaluatorState.cs
```diff
@@ -0,0 +1,199 @@
...
+        // TODO: shaddowFields ...
+    }
+
+
+    /// <summary>
+    /// Get a hierarcy of the different contexts in the component model (remember to iterate <see cref="ComponentContext.ChildContexts" />)
+    /// </summary>
+    public IEnumerable<ComponentContext> GetComponentContexts()
+    {
+        if (_componentModel is null || _dataModel is null)
```
RonnyB71: I think these checks and exception throwing should go in the constructor to ensure the object is initialized in a valid state instead of deferring the check to when a method is called.

src/Altinn.App.Core/Models/Layout/Components/OptionsComponent.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// <summary>
+    /// Is the component referencing a secure code list (uses security context of the instance)
+    /// </summary>
+    public bool Secure { get; }
+
+    /// <summary>
+    /// Constructor
+    /// </summary>
+    public OptionsComponent(string id, string type, IReadOnlyDictionary<string, string>? dataModelBindings, Expression? hidden, Expression? required, Expression? readOnly, string? optionId, List<AppOption>? options, bool secure, IReadOnlyDictionary<string, string>? extra) :
```
RonnyB71: Does it make sense to group some of the parameters into a parameter object to reduce the number and make it a bit more readable? hidden, required, readonly could be candidates for grouping.

src/Altinn.App.Core/Models/Layout/BaseComponent.cs
```diff
@@ -0,0 +1,84 @@
...
+/// for components that don't have them.
+/// </remarks>
+public class BaseComponent
+{
+    /// <summary>
+    /// Constructor that ensures n
+    /// </summary>
+    public BaseComponent(string id, string type, IReadOnlyDictionary<string, string>? dataModelBindings, Expression? hidden, Expression? required, Expression? readOnly, IReadOnlyDictionary<string, string>? extra)
+    {
+        Id = id;
```
RonnyB71: Should we validate Id and Type in the constructor and throw an InvalidArgumentException if they are null or empty?

src/Altinn.App.Core/Models/Layout/Components/OptionsComponent.cs
ivarne: Maybe? but they will be separate variables in the parser, and separate properties on the Component class. I don’t see the huge benefit of grouping them before calling the constructor.

src/Altinn.App.Core/Models/Layout/BaseComponent.cs
```diff
@@ -0,0 +1,84 @@
...
+    }
+    /// <summary>
+    /// ID of the component (or pagename for pages)
+    /// </summary>
+    public string Id { get; }
+
+    /// <summary>
+    /// Get the page for the component
+    /// </summary>
+    public string Page
```
RonnyB71: Page | PageId | PageName 🤔 Page feels like it's a Page object, but it's just a id/name as a string. Might be a review-in-browser-thing and easier to reason about in and IDE. Actually another place I think PageRef was used, so we should be consistent on this.

src/Altinn.App.Core/Models/Layout/BaseComponent.cs
ivarne: I do that in the Json parser before calling the constructor, and they are declared as non nullable arguments.

src/Altinn.App.Core/Models/Layout/BaseComponent.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// <summary>
+    /// The group or page that this component is part of. NULL for page components
+    /// </summary>
+    public BaseComponent? Parent { get; internal set; }
+
+    /// <summary>
+    /// Extra properties that are not modelled explicitly as a class that inhertits from <see cref="BaseComponent" />
+    /// </summary>
+    public IReadOnlyDictionary<string, string>? Extra { get; }
```
RonnyB71: Extra | ExtraProperties | AdditionalProperties 🤔 consider the two latter for a bit more explicit statement of what the dictionary contains without having to read the comment.

src/Altinn.App.Core/Models/Layout/BaseComponent.cs
ivarne: Haha! Great catch! I got confused about this today🫣🫣.

src/Altinn.App.Core/Models/Layout/GroupComponent.cs
```diff
@@ -0,0 +1,29 @@
+using System.Text.Json;
+
+using Altinn.App.Core.Models.Expressions;
+
+namespace Altinn.App.Core.Models.Layout;
+
+/// <summary>
+/// Tag component to signify that this is a group component
+/// </summary>
+public class GroupComponent : BaseComponent
```
RonnyB71: Any particular reason why this isn't in the Components namespace with the other components?

ivarne: After trying to explain my reasoning, that seemed to be sound when I put them in separate namespaces, I hope you will accept the answer “no, not anymore”.

ivarne: [ad296dc](https://github.com/Altinn/app-lib-dotnet/pull/23/commits/ad296dc06c4c7617099a33451e282da243177168)

src/Altinn.App.Core/Models/Layout/BaseComponent.cs
ivarne: [1c80254](https://github.com/Altinn/app-lib-dotnet/pull/23/commits/1c80254b1a44be04783e85cd76cbab0f09876abf)

src/Altinn.App.Core/Models/Layout/BaseComponent.cs
ivarne: [b731c8c](https://github.com/Altinn/app-lib-dotnet/pull/23/commits/b731c8c85d54b5136b2639f60f7e1738be6595b7)

src/Altinn.App.Core/Expressions/LayoutEvaluatorState.cs
ivarne: I just dropped the null checks. They are declared as non nullable in the constructor, so the compiler will check, or if you disable nullable reference types, you deserve a null pointer exception.

The reason I did the check previously, was that I wanted the properties to be nullable, and have some way of figuring out what data is actually needed for a given evaluation. It turned out to be too complex, so I just gave up.
085b143

src/Altinn.App.Core/Models/Layout/Components/OptionsComponent.cs
ivarne: I made an attempt, but I'm not sure it improved readability to a huge extent. https://github.com/Altinn/app-lib-dotnet/commit/225027b2f200706c10473592a70bf69d66a8ff7d

Note that this was on a separate branch. Can include in this PR if you like it.

RonnyB71: I think the reduced number of parameters in the constructor made it more readable, although I was a bit confused at first since you used the component base class itself. I was more thinking about a simple parameter object if any parameters naturally belonged together and could be named thereafter eg. DisplayOptions (hidden, required, readonly) or something. Personally, I strive to keep the number of parameters to 3 or less, max 5, but here we have 10 which I hoped we could reduce a bit. But won't block any approval:)

RonnyB71 (approved):
🕹️

src/Altinn.App.Core/Models/Layout/Components/BaseComponent.cs
```diff
@@ -19,15 +19,15 @@ public class BaseComponent
     /// <summary>
     /// Constructor that ensures n
```
RonnyB71: Complete comment:)

RonnyB71 (approved):
🕹️

RonnyB71 (approved):
🕹️

src/Altinn.App.Core/Expressions/LayoutEvaluatorState.cs
RonnyB71: > I’m not completely convinced, at least not if you think this should be marked `internal` instead of `public`. Seems like this would be really useful to have available in apps and other code, and because of the json in apps, it needs to be very stable.

No, I don't mean the classes should be internal - just moved to the internal namespace together with the other features that is internal. The reasoning is partly to avoid all internal features on root but mostly to signal that this is internal stuff, and if you replace implementations from this namespace then you need to be sure of what you are doing.

RonnyB71:
> I’m not completely convinced, at least not if you think this should be marked internal instead of public. Seems like this would be really useful to have available in apps and other code, and because of the json in apps, it needs to be very stable.

No, I don't mean the classes should be internal - just moved to the internal namespace together with the other features that is internal. The reasoning is partly to avoid all internal features on root but mostly to signal that this is internal stuff, and if you replace implementations from this namespace then you need to be sure of what you are doing.

src/Altinn.App.Core/Models/Layout/Components/OptionsComponent.cs
ivarne: Given your vague reply, I think I'd like to keep the current constructor.

src/Altinn.App.Core/Models/Layout/Components/BaseComponent.cs
ivarne: This comment is on an old diff. Currently the comment is 
```csharp
/// <summary>
/// Constructor
/// </ summary>
```
Not sure how to really add value, but i changed it to `Constructor for <see cref="BaseComponent" />`?

src/Altinn.App.Core/Models/Layout/BaseComponent.cs
RonnyB71: Ok, but that's another class and only compile time checks. Isn't it better if the class that actually has the properties ensures they are correct and that it leaves the constructor with the object in valid state?

RonnyB71:
Created a [PR based on your branch](https://github.com/Altinn/app-lib-dotnet/pull/68) to run the SonarCloud analytics. Can you take a look at the code smells reported and check the test coverage if some critical paths are not covered.

src/Altinn.App.Core/Expressions/ExpressionEvaluator.cs
```diff
@@ -0,0 +1,307 @@
...
+        {
+            return defaultReturn;
+        }
+
+        return EvaluateExpression(state, expr, context) switch
+        {
+            true => true,
+            false => false,
+            null => defaultReturn,
+            _ => throw new Exception($"Return from evaluating \"{property}\" on \"{context.Component.Id}\" was not boolean (value)")
```
RonnyB71: We should either find a more specific exception or create a custom application exception.

src/Altinn.App.Core/Expressions/ExpressionEvaluator.cs
```diff
@@ -0,0 +1,307 @@
...
+            ExpressionFunctionEnum.concat => Concat(args),
+            ExpressionFunctionEnum.equals => EqualsImplementation(args),
+            ExpressionFunctionEnum.notEquals => !EqualsImplementation(args),
+            ExpressionFunctionEnum.greaterThanEq => GreaterThanEq(args),
+            ExpressionFunctionEnum.lessThan => LessThan(args),
+            ExpressionFunctionEnum.lessThanEq => LessThanEq(args),
+            ExpressionFunctionEnum.greaterThan => GreaterThan(args),
+            ExpressionFunctionEnum.and => And(args),
+            ExpressionFunctionEnum.or => Or(args),
+            // "<" => args.First() <  args.ElementAt(1),
```
RonnyB71: Should these be deleted?

src/Altinn.App.Core/Expressions/ExpressionEvaluator.cs
```diff
@@ -0,0 +1,307 @@
...
+                    _ => throw new Exception($"Expected boolean, got value \"{s}\""),
+                }
+            },
+            double s => s switch
+            {
+                1 => true,
+                0 => false,
+                _ => throw new Exception($"Expected boolean, got value {s}"),
+            },
+            _ => throw new Exception("TODO: FIxme"),
```
RonnyB71: Exception says it all:)

src/Altinn.App.Core/Expressions/LayoutEvaluator.cs
```diff
@@ -0,0 +1,116 @@
...
+                {
+                    if (state.GetModelData(binding, context) is null)
+                    {
+                        validationIssues.Add(new ValidationIssue()
+                        {
+                            Severity = ValidationIssueSeverity.Error,
+                            InstanceId = state.GetInstanceContext("instanceId").ToString(),
+                            DataElementId = dataElementId,
+                            Field = state.AddInidicies(binding, context),
+                            Description = "TODO required",
```
RonnyB71, line 108: Fix TODO?

src/Altinn.App.Core/Expressions/LayoutEvaluatorState.cs
```diff
@@ -0,0 +1,193 @@
...
+    /// <summary>
+    /// Constructor for LayoutEvaluatorState. Usually called via <see cref="LayoutEvaluatorStateInitializer" /> that can be fetched from dependency injection.
+    /// </summary>
+    public LayoutEvaluatorState(IDataModelAccessor dataModel, LayoutModel componentModel, FrontEndSettings frontEndSettings, Instance instance)
+    {
+        _dataModel = dataModel;
+        _componentModel = componentModel;
+        _frontEndSettings = frontEndSettings;
+        _instanceContext = instance;
+        // TODO: shaddowFields ...
```
RonnyB71: I guess this is for later when we actually implement shadow fields so we can just delete the TODO for now.

src/Altinn.App.Core/Features/Validation/ValidationAppSI.cs
```diff
@@ -186,21 +190,34 @@ public async Task<List<ValidationIssue>> ValidateDataElement(Instance instance,
                 dynamic data = await _dataService.GetFormData(
                     instanceGuid, modelType, instance.Org, app, instanceOwnerPartyId, Guid.Parse(dataElement.Id));
 
+                // Remove hidden data before validation
+                //TODO: Figure out the layout set id from task name
```
RonnyB71: Something to TODO or forgot to remove comment?

src/Altinn.App.Core/Helpers/DataModel/JsonDataModel.cs
```diff
@@ -0,0 +1,153 @@
...
+namespace Altinn.App.Core.Helpers.DataModel;
+
+/// <summary>
+/// Implementation of <see cref="IDataModelAccessor" /> for data models based on JsonElement (mainliy for testing )
+/// </summary>
+/// <remarks>
+/// This class is written to enable the use of shared tests (with frontend) where the datamodel is defined
+/// in json. It's hard to IL generate proper C# classes to use the normal <see cref="DataModel" /> in tests
+/// </remarks>
+public class JsonDataModel : IDataModelAccessor
```
RonnyB71: Move class to testproject.

ivarne: Done

src/Altinn.App.Core/Expressions/ExpressionEvaluator.cs
ivarne: Changed to  
```c#
_ => throw new ExpressionEvaluatorTypeErrorException("Unknown data type encountered in expression: " + arg.GetType().Name),
```

src/Altinn.App.Core/Expressions/ExpressionEvaluator.cs
ivarne: Done!

src/Altinn.App.Core/Expressions/ExpressionEvaluator.cs
ivarne: I created `ExpressionEvaluatorTypeErrorException` and `DataModelException` to comply with the coding standard.

src/Altinn.App.Core/Expressions/LayoutEvaluator.cs
ivarne: Yes, not sure what to write here. I have not figured out how validation really should work, and where this message is displayed.

I could possibly write something like 
```c#
Description = $"{context.Component.Id} is Required"
```
, but that might be a guid, that the user might not know anything about. Maybe I should use the `binding` (with indexes) instead, so that the error is tied to the datamodel, but that is already availible in the `Field` prop?

src/Altinn.App.Core/Models/Layout/PageComponentComverter.cs
```diff
@@ -0,0 +1,353 @@
...
+                default:
+                    // store extra fields as json
+                    additionalProperties[propertyName] = reader.SkipReturnString();
+                    break;
+            }
+        }
+        ThrowJsonExceptionIfNull(id);
+        ThrowJsonExceptionIfNull(type);
+
+        switch (type.ToLowerInvariant())
```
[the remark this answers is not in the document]
ivarne: This is depressing. In order to reduce CognitiveComplexity for SonarCloud, I had to replace
```c#
if (type is null)
{
    throw JsonException("type is null");
}
/// With
ThrowJsonExceptionIfNull(type);
```
I used the `[NotNull]` annotation to make sure Roslyn understood that `ThrowJsonExceptionIfNull` would not return if `type is null`, but CodeQL does not seem to rely on standard nullability analysis and fails here.

test/Altinn.App.Core.Tests/LayoutExpressions/CommonTests/TestContextList.cs
```diff
@@ -0,0 +1,111 @@
...
+                    data,
+                    new JsonSerializerOptions
+                    {
+                        ReadCommentHandling = JsonCommentHandling.Skip,
+                    })!;
+            }
+            catch (Exception e)
+            {
+                testCase.ParsingException = e;
+            }
```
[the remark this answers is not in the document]
ivarne: I think a generic catch is suitable for tests.

test/Altinn.App.Core.Tests/LayoutExpressions/CommonTests/TestFunctions.cs
```diff
@@ -0,0 +1,194 @@
...
+                    })!;
+            }
+            catch (Exception e)
+            {
+                using var jsonDocument = JsonDocument.Parse(data);
+
+                testCase.Name = jsonDocument.RootElement.GetProperty("name").GetString();
+                testCase.ExpectsFailure = jsonDocument.RootElement.TryGetProperty("expectsFailure", out var expectsFailure) ? expectsFailure.GetString() : null;
+                testCase.ParsingException = e;
+            }
```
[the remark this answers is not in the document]
ivarne: I think a generic catch is suitable for tests.

src/Altinn.App.Core/Implementation/DefaultTaskEvents.cs
```diff
@@ -118,16 +124,24 @@
...
+        Guid instanceGuid = Guid.Parse(instance.Id.Split("/")[1]);
+        List<DataType> dataTypesToLock = _appMetadata.DataTypes.FindAll(dt => dt.TaskId == endEvent);
+        try
+        {
+            await RemoveHiddenData(instance, instanceGuid, dataTypesToLock);
+        }
+        catch(Exception e)
+        {
+            _logger.LogError(e, "Failed to remove hidden data in task");
+        }
```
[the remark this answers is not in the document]
ivarne: I don't think any error in removing hidden data is worth failing the submission of the schema.

src/Altinn.App.Core/Features/Validation/ValidationAppSI.cs
```diff
@@ -183,24 +187,44 @@
...
+                var evaluationState = await _layoutEvaluatorStateInitializer.Init(instance, data, layoutSet?.Id);
+                // Remove hidden data before validation
+                try
+                {
+                    LayoutEvaluator.RemoveHiddenData(evaluationState);
+                }
+                catch (Exception e)
+                {
+                    _logger.LogError(e, "Failed to remove hidden data");
+                }
```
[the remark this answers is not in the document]
ivarne: Unexpeced errors in removing hidden data should not trigger validation failiure.

RonnyB71:
Closing this as it's now a duplicate of https://github.com/Altinn/app-lib-dotnet/pull/68

PR #18, opened 2022-09-12
Chore/90 appbase simplified

RonnyB71:
Created V7 of the nuget packages.

## Description
1. Merged Altinn.App.Common and Altinn.App.PlatformServices into Altinn.App.Core, now shipping:
  * Altinn.App.Api
  * Altinn.App.Core
  * Altinn.App.CodeGeneration

2. Removed the need for AppBase and App.cs

3. Simplified Program.cs

4. Reorganized namespaces

5. Replaced Stylecop with editorconfig settings.

## Related Issue(s)
- [#108](https://github.com/altinn/app-template-dotnet/issues/108)
- [#109](https://github.com/altinn/app-template-dotnet/issues/109)
- [#110](https://github.com/altinn/app-template-dotnet/issues/110)
- [#111](https://github.com/altinn/app-template-dotnet/issues/111)
- [#112](https://github.com/altinn/app-template-dotnet/issues/112)

## Verification
- [x] **Your** code builds clean without any errors or warnings
- [x] Manual testing done (required)
- [x] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #24, opened 2022-09-27
Fix a few nullability warnings.

ivarne:
I'm having problems finding the warnings in my code, so now that the interface branch is merged to main in #18, I think its time to start fixing the obvious ones at least.


ivarne:
Number of warnings in app-lib-dotnet reduced from **419** to **227**. Half of the remaining issues are `cs8618`, where non-nullabe properties requires a constructor to set a non-null value. They are really annoying when deserializing objects from JSON, and I typically just add ` = default!;` to fix them, but it needs to be done with care.

RonnyB71:
> Number of warnings in app-lib-dotnet reduced from **419** to **227**. Half of the remaining issues are `cs8618`, where non-nullabe properties requires a constructor to set a non-null value. They are really annoying when deserializing objects from JSON, and I typically just add ` = default!;` to fix them, but it needs to be done with care.

Great job @ivarne 👍 I see there is a conflict, but I think that's related to changes cone by @tjololo yesterday and I assume he would like a quick look on this as well to get some nit picking eyes on this:)

src/Altinn.App.Core/Implementation/DefaultTaskEvents.cs
```diff
@@ -114,16 +117,15 @@ public async Task OnEndProcessTask(string endEvent, Instance instance)
...
-            bool generatePdf = dataType.AppLogic?.ClassRef != null && dataType.EnablePdfCreation;
-
             foreach (DataElement dataElement in instance.Data.FindAll(de => de.DataType == dataType.Id))
             {
                 dataElement.Locked = true;
                 _logger.LogInformation($"Locking data element {dataElement.Id} of dataType {dataType.Id}.");
                 Task updateData = _dataClient.Update(instance, dataElement);
 
-                if (generatePdf)
+                if (dataType.AppLogic?.ClassRef != null && dataType.EnablePdfCreation)
```
tjololo: Not sure how the compiler would actually do here, but are we potentially evaluating this boolean for each DataElement for each DataType instead of just for each DataType?

src/Altinn.App.Core/Internal/Language/ApplicationLanguage.cs
```diff
@@ -34,17 +34,16 @@ public async Task<List<Models.ApplicationLanguage>> GetApplicationLanguages()
             var directoryInfo = new DirectoryInfo(pathTextsResourceFolder);
             var textResourceFilesInDirectory = directoryInfo.GetFiles();
             var applicationLanguages = new List<Models.ApplicationLanguage>();
+            JsonSerializerOptions options = new() { PropertyNamingPolicy = JsonNamingPolicy.CamelCase };
```
tjololo, line 37: Nice catch!

tjololo (commented):
Great work! Just one question/nitpick from me. Once my curiosity and the conflict is resolved we should get this into main

src/Altinn.App.Core/Implementation/DefaultTaskEvents.cs
ivarne: Yes, but Altinn has way more important optimizations that could be done that would actually improve something.

I agree this looks strange, but it was the simplest way to get inference to work for nullability checks. An obvious alternative would be to add ‘.!’ to suppress the warnings on the next lines.

tjololo: Shouldn't we opt for the `!.` approach? Then we would not introduces another, although small, performance reduction and the code should still be null safe as the if statement still checks that `dataType.AppLogic?.ClassRef` isn't null?

ivarne: Both approaches works, but I find it hard to reason about .! when the condition is far away as in this case.

SandGrainOne: I'm disapointed that the analyser is unable to see, through the original boolean variable, that the variable cannot be null when used inside the block.

tjololo: I agree that it's hard to reason about !. when the condition is far away, but then again the if statement is harder to read with these changes, in my opinion. I vote for the !. approach since this make the code as a whole more readable and we avoid a small performance hit. 

Maybe we should file a featurerequest/bug to the compiler team? 😄

ivarne: I have reverted this change. I'd really like the to get rid of the other almost 200 warnings.

src/Altinn.App.Core/Internal/Language/ApplicationLanguage.cs
ivarne: This is at least a small performance boost, as `System.Text.Json` caches some of it's reflection results on the `JsonSerializerOptions` instance.

tjololo (approved):
Great work 👍

Issue #25, opened 2022-09-28
Add extension point for abandon process task


tjololo:
### Description

Through https://github.com/Altinn/altinn-studio/issues/2058 and https://github.com/Altinn/app-frontend-react/issues/432 it will be possible to move back to a Data task from Confirmation.
To enable service owner developers to write custom logic that should happen if a task is abandoned we should make it easy the write custom logic when a taks is abandoned

### Additional Information

Option 1: extend https://github.com/Altinn/app-lib-dotnet/blob/main/src/Altinn.App.Core/Features/ITaskProcessor.cs with the method: `public Task ProcessTaskAbandon(string taskId, Instance instance);`

### Tasks

- [ ] Extend ITaskProcessor
- [ ] Implement new method in NullImplementation

### Acceptance Criterias

- [ ] Service owner developers are able to write custom logic when a task is abandoned in their app.

PR #26, opened 2022-09-28
Add new method to handle abandoned task events

tjololo:
## Description
When a process task is abandoned service owners are now required to handle these in the new method ProcessTaskAbandon in ITaskProcessor.
Previous solution made it impossible for service owners to differentiate between a task being completed or abandoned.

Update to the integrations tests in app-template-dotnet is coming

## Related Issue(s)
- #25 

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Core/Features/ITaskProcessor.cs
```diff
@@ -13,4 +13,12 @@ public interface ITaskProcessor
...
-}
\ No newline at end of file
+    
+    /// <summary>
+    /// Method for defining custom process on TaskAbandoned event.
+    /// </summary>
+    /// <param name="taskId"></param>
+    /// <param name="instance"></param>
+    /// <returns></returns>
+    public Task ProcessTaskAbandon(string taskId, Instance instance);
```
RonnyB71: This would be a breaking change if the ITaskProcessor wasn't introduced in v7 which isn't out yet. But could this have been a new interface IAbandonTask? Or do we think that is too fine grained?

tjololo: Yes, would have been a breaking change if not. Smaller interfaces is a good idea, if the developer implementing both interfaces he can choose to just have one class implement both.
If we extract this to IAbandonTask we should probably rename ITaskProcessor to IEndTask

ivarne: There is now default [implementations](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/default-interface-methods-versions), so adding new methods, isn’t really a big problem any more.

implementing multiple small interfaces in the same class, means you need two service registrations (one for each method used).

tjololo: We opted for two interfaces. The both describe the same method.
Changed the calling code to use IEnumerable instead, this makes it possible to define one class with logic that should run in both cases just by extending both interfaces and registering two services. If custom logic for one of the process events is needed the developer can create a class implementing only one of the interfaces and register this service also

ivarne: If I’m to implement two interfaces in the same class, it would be great if the methods in the interfaces would have different names. Two `ProcessEvent` methods is kind of awkward.

tjololo: With the latest changes you would have to create two classes if the logic for Abandon differs from End. 
If the same code is to be executed on both you can use the same class (implementing both interfaces) register the services and the same code is executed for both events

tjololo: here is an example app using the interfaces: https://altinn.studio/repos/ttd/custom-view-confirm-back/src/branch/master/App

tjololo: Given it a fair amount of thought and I starting to disagree with my last change (interfaces defining the same method signature). The idea behind having two interfaces is to spare the developers having to implement methods they aren't using, seeing as most app today does not have the need for abandon hook. In my opinion two interfaces is the best way to solve that.
The initial idea behind similar method signatures was to have the possibility to create common logic, but that would come at the cost of being able to implement both interfaces in one class. Not sure if common logic for both events is a use case we would see alot or not. If we opt for different signatures the developer could implement it in one class and easily call one method from the other.
What do you think @RonnyB71 and @ivarne ?

ivarne: For performance it’s a minute benefit to have small interfaces that only gets instantiated when actually required.

For discovering the available hooks, it’s better for app devs if we group multiple hooks in sensible interfaces with default implementations, so you only override what you need.

My highest priority is that we must solve similar problems in a similar way.

tjololo: The main theme of v7 is smaller interfaces to solve specific tasks. So i fell this is in the theme of the other changes we have made. 
I will push a change where the method signatures will differ again so the interfaces can be implemented in one class.
My suggestion is `IProcessTaskEnd.End(taskId, instance)` and `IProcessTaskAbandon.Abandon(taskId, instance)`

RonnyB71 (approved):

src/Altinn.App.Core/Features/ITaskProcessor.cs
tjololo: The main reason for separating these is that the entire process engine is due for some rework and I don't think abandon has ha natural home in that rework.

ivarne: If you think it might make sense to one day remove IProcessTaskAbanon while keeping IProcessTaskEnd, it makes a lot of sense to have them separate.

PR #28, opened 2022-09-30
Added update of instance after a task end or abandon is performed

TheTechArch:
Fixed saving of instance


tjololo (approved):
:godmode:

src/Altinn.App.Core/Implementation/ProcessChangeHandler.cs
```diff
@@ -163,7 +163,7 @@ private async Task<Instance> UpdateProcessAndDispatchEvents(ProcessChangeContext
 
             // need to update the instance process and then the instance in case appbase has changed it, e.g. endEvent sets status.archived
             Instance updatedInstance = await _instanceClient.UpdateProcess(processChangeContext.Instance);
-
+
```
ivarne:
```suggestion

```

PR #34, opened 2022-10-04
More null waring fixes 222 => 193

ivarne:
More cleanup separated into 4 commits (should probably be reviewed separately)
* [Move initialization to construction.](https://github.com/Altinn/app-lib-dotnet/commit/2663356e4938dfed70876b6f75a21afd7f175632)
* [Add nullable annotations where they should be](https://github.com/Altinn/app-lib-dotnet/commit/445e67841727b75d89330e7a243d83372100889e)
* [Add constructor to fix nullable warnings in test](https://github.com/Altinn/app-lib-dotnet/commit/b6168070ff15be812324f3ce4f7dbfc52e54000c)
* [Delete class without references](https://github.com/Altinn/app-lib-dotnet/commit/d86cc3ed7c959c1e14de02d24852c5f7a8f14701) (I hope this is OK, I could not find any reference to this class anywhere, so it seems to be a leftover.)


src/Altinn.App.Core/Implementation/AppResourcesSI.cs
```diff
@@ -256,14 +256,14 @@ public string GetModelJsonSchema(string modelId)
...
             {
-                filedata = File.ReadAllText(filename, Encoding.UTF8);
+                var filedata = File.ReadAllText(filename, Encoding.UTF8);
+                LayoutSettings layoutSettings = JsonConvert.DeserializeObject<LayoutSettings>(filedata)!;
+                return layoutSettings;
             }
 
-            LayoutSettings layoutSettings = JsonConvert.DeserializeObject<LayoutSettings>(filedata);
-            return layoutSettings;
+            throw new Exception($"Could not find file in {filename}");
```
tjololo:
```suggestion
            throw new FileNotFoundException($"Could not find layoutsettings file: {filename}");
```
Throw a more acurate exception

src/Altinn.App.Core/Internal/AppModel/DefaultAppModel.cs
```diff
@@ -3,24 +3,30 @@
...
-        return Activator.CreateInstance(GetModelType(classRef));
+        return Activator.CreateInstance(GetModelType(classRef))!;
     }
 
+    /// <inheritdoc />
     public Type GetModelType(string classRef)
     {
         _logger.LogInformation($"GetAppModelType {classRef}");
-        return Assembly.GetEntryAssembly().GetType(classRef);
+        return Assembly.GetEntryAssembly()?.GetType(classRef) ?? throw new Exception("App model of type {classRef} does not exist");
```
tjololo:
```suggestion
        return Assembly.GetEntryAssembly()?.GetType(classRef, true);
```

src/Altinn.App.Core/Models/Components.cs
```diff
@@ -8,6 +8,6 @@ public class Components
         /// <summary>
         /// Exclude from pdf
         /// </summary>
-        public List<string> ExcludeFromPdf { get; set; }
+        public List<string>? ExcludeFromPdf { get; set; }
```
tjololo, line 11:
```suggestion
        public List<string> ExcludeFromPdf { get; set; } = new();
```

src/Altinn.App.Core/Models/LayoutSets.cs
```diff
@@ -8,6 +8,6 @@ public class LayoutSets
         /// <summary>
         /// Sets
         /// </summary>
-        public List<LayoutSet> Sets { get; set; }
+        public List<LayoutSet>? Sets { get; set; }
```
tjololo, line 11:
```suggestion
        public List<LayoutSet> Sets { get; set; } = new();
```

src/Altinn.App.Core/Models/Pages.cs
```diff
@@ -8,11 +8,11 @@ public class Pages
         /// <summary>
         /// Order
         /// </summary>
-        public List<string> Order { get; set; }
+        public List<string>? Order { get; set; }
```
tjololo, line 11:
```suggestion
        public List<string> Order { get; set; } = new();
```

src/Altinn.App.Core/Models/Pages.cs
```diff
@@ -8,11 +8,11 @@ public class Pages
...
 
         /// <summary>
         /// Exclude from pdf
         /// </summary>
-        public List<string> ExcludeFromPdf { get; set; }
+        public List<string>? ExcludeFromPdf { get; set; }
```
tjololo, line 16:
```suggestion
        public List<string> ExcludeFromPdf { get; set; } = new();
```

tjololo (changes requested):
Overall nice work, some small changes.
Throw explicit exceptions instead of System.Exception
Lists should initialised to empty lists

src/Altinn.App.Core/Internal/AppModel/DefaultAppModel.cs
ivarne: I'm generally in favor of throwing my own exceptions that explain what happens (especially when the error is triggered by a config error), but for your approach to work, we at least need to suppress the warnings with `!`
```suggestion
        return Assembly.GetEntryAssembly()!.GetType(classRef, true)!;
```

src/Altinn.App.Core/Models/Components.cs
ivarne: I sometimes prefer this approach too, but this class is the target of Json Deserialization of a file that users will edit in VSCode. I think it is better to include the nullability in the classes, instead of pretending that we don't support json that we actually need to support for backwards compatibility.

src/Altinn.App.Core/Internal/AppModel/DefaultAppModel.cs
tjololo: I agree that throwing explicit exceptions is the best, but in this case GetType(classRef, true) throws a very explicit exception:
```
System.TypeLoadException: Could not load type 'Altinn.App.Models.Skjemas' from assembly 'Altinn.App, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
         at System.Reflection.RuntimeAssembly.GetType(QCallAssembly assembly, String name, Boolean throwOnError, Boolean ignoreCase, ObjectHandleOnStack type, ObjectHandleOnStack keepAlive, ObjectHandleOnStack assemblyLoadContext)
         at System.Reflection.RuntimeAssembly.GetType(String name, Boolean throwOnError, Boolean ignoreCase)
         at System.Reflection.Assembly.GetType(String name, Boolean throwOnError)
... Omitted for brevity.....
```
I think we don't need to handle cases where EntryAssembly is null, if we do we should throw an exception that explains that the EntryAssembly is null.
So I agree on your suggestion to my suggestion :)

src/Altinn.App.Core/Models/Components.cs
tjololo: Ah, I missed that one! I will mark these as resolved.

src/Altinn.App.Core/Internal/AppModel/DefaultAppModel.cs
ivarne: Yea, that message was pretty good.

tjololo (approved):
Great! lets get it onto main

PR #35, opened 2022-10-04
Fix 10 more warnings 197 -> 187

ivarne:

PR #42, opened 2022-10-11
Change IInstantiationValidator.Validate to be nullable return value

ivarne:
Pretty obvous change, as the docs specify to return null.

> `The validation result object (null if no errors)`


ivarne:
Sorry, I see this was a duplicate of part of #35, but this is causing warnings in updated apps, so somewhat more important than the other ones.

tjololo (approved):
Indeed an obvious change

Issue #49, opened 2022-09-19
Register event subscription in app


RonnyB71:
### Description

The application needs to be able to subscribe to "note-to-selv" events during startup ie. evnts that the app itself creates to remind itself to check the status on previous external message deliveries. Technically any type of event should be possible to subscribe to.

### Tasks

- [x] Add support for subscribing to a custom "note-to-self" event on application start or basically any type of custom event from the app.

### Acceptance Criterias
- [x] The subscription is successfully registered in the event system
- [x] The app is authenticated as it self towards the event system

PR #80, opened 2022-11-06
Fix event subscription and eFormidling status check.

RonnyB71:
This change checks to see if an implementation for IEventSubscription is already added to the service collection before adding it.
Add more logging to the EventSubscriptionClient and other areas.

## Related Issue(s)
- #49 

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Core/Internal/Events/EventHandlerResolver.cs
```diff
@@ -16,6 +16,11 @@ public EventHandlerResolver(IEnumerable<IEventHandler> eventHandlers)
         /// <inheritDoc/>
```
tjololo:
```suggestion
        /// <inheritdoc/>
```

src/Altinn.App.Api/Controllers/EventsReceiverController.cs
```diff
@@ -62,11 +69,11 @@ public async Task<ActionResult> Post([FromQuery] string code, [FromBody] CloudEv
             }
             catch (NotImplementedException)
             {
-                return BadRequest();
+                return BadRequest($"No eventhandler found that supports {cloudEvent.Type}");
             }
             catch (Exception ex)
```
tjololo, line 74: Catching and handling all exceptions in controllers makes sense to me, should we dismiss the CodeQL alert with won't fix?

tjololo (approved):
:godmode:

src/Altinn.App.Api/Controllers/EventsReceiverController.cs
RonnyB71: Yes, in this particular case I wanted a more explicit logging that we would get with the generic error handling which also would return a 500 but with less info logged.

PR #95, opened 2022-11-18
Backend dynamics cleanup

ivarne:
No hurry to relase, but this is required when apps start using dotnet 7 version of `System.Text.Json`

[Use AyncLocal instead of ConditionalWeakTable to set PageName](https://github.com/Altinn/app-lib-dotnet/commit/cbf7975a1e397b43a73bfd615c84485c4c9b4995) 

The json for a page does not contain the page name, so it needs to be
passed in a different way to the page constructor. ConditionalWeakTable
with the JsonSerializerOptions does not seem to work in dotnet 7.

I also happened to notice that `LayoutModelConverter` was only used in the Tests project, so I moved it.


tjololo (approved):
:godmode:

olemartinorg:
@tjololo Anything blocking this from merging? :blush: I don't think @ivarne can push the button.

PR #97, opened 2022-11-22
Expressions: Adding more tests, fixing double.TryParse on Linux

olemartinorg:
## Description
Syncing tests with frontend, fixing broken tests on Linux (re: double TryParse invariant), adding more null-tests

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/pull/685

## Verification
- [x] All tests run green

## Documentation


ivarne:
Great to see you were able to understand my code in a way that allowed you to make updates!

It seems to me that you made two test files equal. 

https://github.com/Altinn/app-lib-dotnet/blob/chore/expr-new-tests/test/Altinn.App.Core.Tests/LayoutExpressions/CommonTests/shared-tests/functions/component/hidden-in-group-other-row.json
vs 
https://github.com/Altinn/app-lib-dotnet/blob/chore/expr-new-tests/test/Altinn.App.Core.Tests/LayoutExpressions/CommonTests/shared-tests/functions/component/hidden-in-group-hidden-other-row.json

ivarne:
That `ParseDouble` logic really needs a lot of tests! Not just between backend/front end, but also between browsers.

olemartinorg:
@ivarne Oh, well spotted! I just changed whitespace in that one test, so they were already duplicate apart from some indentation. Deleted it and renamed the remaining one to better indicate what it does.

Also, the tests were failing on my machine (possibly some wonky default locale for dotnet?). Strange that they weren't failing when running on githubs linux machines. I think they're fine enough as-is, there are a lot of implicit tests in the `greaterThan` function tests that also test decimal casting - and we're using essentially using the same regex to verify that numbers can be cast.

ivarne:
👍 🚀

tjololo (approved):
Leaning on the expert @ivarne, but looks good to me also.

I must commend you both on the work and the test setup once again! Easy to read what is expected from the code by reading the tests 👏

PR #98, opened 2022-11-24
App validator (Request for review)

ivarne:
This is a first draft on an app validator to ensure all json files parse, and that references are valid.

Some of the functionality might be better to do in frontend instead, but we need the infrastructure in backend, because some of the checks are not frontend relevant.
* `applicationmetadata.json` `appLogic.classRef` is not possible to validate on frontend
* the C# model must match the json schema model. 
* I also implemented some of the reference validations that has hurt our productivity badly. I don't think duplication is really a big issue here, as apps with these issues will never be valid.

See also
* https://github.com/Altinn/app-frontend-react/issues/648

A big advantage of validation in C# is that I can print file name and location, to get a clickable link in the vsCode terminal to the exact location of the error.
image

### Solution description
New implementation of `IAppResources` that depends on the new `AppFilesBytes` so that all files can be loaded async in a hosted service on startup. That way accessing resources don't rely on reading from disk (I do json parsing for each return, to ensure that modifications to the results don't get preserved.) I have not yet changed the default registration of `IAppResources`, but tested that it works for some of our apps.

New `AppFilesSyncHostedService` that is a hosted service that loads `AppFilesBytes` on startup before the app is ready to serve http requests.

New `AppFilesSyncHostedServiceDebug` that runs validations and updates `AppFilesBytes` on startup and every time any `json`, `xml` or `bpmn` file is changed. Prints validation errors to the console.

### For discussion
How do we report errors? (terminal, 503 with content so that frontend can show the error)

 Should the app crash, or continue on an error?


ivarne:
PS: This also includes the commits for https://github.com/Altinn/app-lib-dotnet/pull/95

olemartinorg:
Fantastic!! :partying_face: :heart_eyes: 

I haven't looked properly at the code, but from the description it seems we agree on a lot here. If you didn't see my [comment yesterday](https://github.com/Altinn/app-frontend-react/issues/648#issuecomment-1326179061), I'm thinking that each project validates the app on their own, and that we collect all validations, merge them, and present them to the user. I'm leaning towards making a JsonSchema for validation messages that each project can use. Something like:

```json
{
  "Some headline/category of validation errors": [
    {
      "class": "duplicate-component-id",
      "title": "Component ID is a duplicate, the same ID was found in page3",
      "description": "These should be unique for every component, across all pages in the same process task",
      "file": "App/ui/layouts/page2.json", // Optional, frontend doesn't know this
      "line": 55, // Optional
      "context": "..." // Optional, backend shouldn't need to fill this in, but it could contain a multi-line string with a formatted representation of the area around where the error occurred in the file
    },
    ...
  ],
  "Other headline/category": [...]
}
```

That way, we could:
1. Render the validation messages locally (for this project that means in the terminal, potentially while crashing the app)
2. Pass the validations to some other system (frontend via a 503, Studio via a command-line tool that outputs the JSON) that can merge the messages along with other messages and present nicely to the user.

What do you think?

ivarne:
I think my [AppValidationError](https://github.com/Altinn/app-lib-dotnet/blob/5ece3526a62a9cda8e28e9f1cd13dbd766fc791b/src/Altinn.App.Core/Internal/AppValidation/AppValidationError.cs) is pretty close to your suggestion.

Adding a middleware that responds with the error for all requests when a static list of errors isn’t empty would be easy, but then we obviously can’t stop the app on error.

olemartinorg:
Ah, great - I did not see that part. We should probably consider adding severity information, then. One or more fatal errors should probably refuse to start the app, but maybe a regular error shouldn't. Same with warnings (and possibly notices).

tjololo:
Great!
Love the idea of fail fast and tell the developer whats wrong.
One thing I started to think about: could we fail even earlier/faster? If this is analysis of static files present at compile time, could we write an analyser that leverages the same checks and giving feedback when we try to compile an application with bad configuration?

ivarne:
@tjololo That was my first thinking, and would be a reasonable way to also ensure docker builds fail, and give very nice red squiggles in json files in VSCode. A few reasons made me reconsider.

1. Running C# inside the compilation process is restrictive and including libraries (for json/jsonSchema) is problematic because you might get into versioning issues (you run in a host process with their dependencies). It might be true (as Microsoft says) that this is a tooling issue, but I don't really see how to solve the compatibility issues of publishing code that should run in process can really be fixed.
2. Users might not want to restart their app every time they change a json file (that doesn't really affect compilation). It is much smoother to just refresh the browser to see the changes.
3. The same code might be used in `studio`, or other places where a compile interface would make integration much harder.

PS: Most developers will use `dotnet run`, so triggering the error on startup isn't much slower for those. Docker build can be failed by running the app to detect errors (and run possible test projects). The remaining benefit would be the VSCode squiggles.

tjololo:
Valid points that I didn't think of!
To have the same validation in studio would it make sense to extract the validation logic into a separate nuget that both app and studio use?

ivarne:
> To have the same validation in studio would it make sense to extract the validation logic into a separate nuget that both app and studio use?

Maybe. Or studio can just depend on `Altinn.App.Core` to get everything (exept the parts of `Altinn.App.Api` that registrers controllers and stuff).

tjololo:
In regards to the question "how do we fail" I have always thought that if the app can not run as expected and we know it, the application should log the error and then crash hard. Id this happens at startup we need to make sure the logs are flushed to application insights as not to leave those who deploy to tt02 in the dark as to why the crash.

Could you make the test pass again?

ivarne:
@tjololo The tests kind of passes, but I dissagree with CodeQL that `catch (Exception e)` is bad (in code that isn't guarded by the request pipeline), and I can't figure out a way to parse JsonDocuments in a loop, and still dispose them with the `using` keyword.

tjololo:
Sorry, thought I had posted a reply, but apparently it was forgotten! 🙈 
As we are checking if the config files are valid it makes sense to me to catch a generic exception to.

We are getting more pull requests that is handling caching of static files, but with variations in implementation. If the caching implementation you have introduced isn't strictly necessary to solve the validation I think we need to take a step back in our team and solve the caching solution in consistent way. We also need to keep the memory footprint in mind as having applications that are used once a week or  less using memory to cache static files would affect the cost.

ivarne:
Sorry, for opening so many pull requests (especially like this one that only represents a partial solution in the hope that you would engage in some sort of discussion to make this solution better). I'll close this and work on alternative approaches for app validation instead.

PR #103, opened 2022-12-07
Bug: Ensure that layouts are loaded from the correct location

ivarne:
Seems like the full tests was lacking, so the file path somehow ended up being wrong.

This issue prevents validation and submission of apps, when `RemoveHiddenDataPreview` is active, so it would be nice if a new version can be made available on nuget pretty quick.


tjololo (approved):

tjololo:
I will merge it and release v7.2.0-preview.5 asap

PR #110, opened 2022-12-09
Fix a few minor bugs in backend dynamics

ivarne:
### 1 Fix a typo in the parsing code
`OptionId` is wrong and is changed to `OptionsId`

### 2 Fix a logic error where the context was checked instead of the target context,


tjololo (approved):
Do you need a new preview version?

ivarne:
> Do you need a new preview version?

Yes, we do!

PR #113, opened 2022-12-13
Ensure that ValidationIssue.Field respects [JsonPropertyName]

ivarne:
When there is a mismatch between the C# property name and the `[JsonPropertyName()]`
image
And one of the nested `[System.ComponentModel.DataAnnotations]` rules fails, the `Field` property of `ValidationIssue` is wrong, and frontend can't attach the error to the correct component.

## Description

Some xsd definitions contain property names with illegal characters that can't be used in properties (without ugly escaping)

To solve that conflict, additional properties is added for json and xml serialziation to include the illegal characters. This needs to be respected everywhere.

In net70, this is easily fixed with a configuration change, https://learn.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-7.0#use-json-property-names-in-validation-errors, but in net60, we need to manually translate the keys.

## Related Issue(s)
- not sure if this has previously been reported.

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Features/Validation/ValidationAppSI.cs
+++ b/src/Altinn.App.Core/Features/Validation/ValidationAppSI.cs
@@ -221,7 +221,7 @@ namespace Altinn.App.Core.Features.Validation
                 // Add the validation messages from System.ComponentModel.DataAnnotations and IInstanceValidator to the return list
                 if (!validationResults.IsValid)
                 {
-                    messages.AddRange(MapModelStateToIssueList(actionContext.ModelState, instance, dataElement.Id));
+                    messages.AddRange(MapModelStateToIssueList(actionContext.ModelState, instance, dataElement.Id, data));
                 }
 
             }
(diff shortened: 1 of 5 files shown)


src/Altinn.App.Core/Features/Validation/ValidationAppSI.cs
```diff
@@ -232,7 +232,8 @@ public async Task<List<ValidationIssue>> ValidateDataElement(Instance instance,
         private List<ValidationIssue> MapModelStateToIssueList(
             ModelStateDictionary modelState,
             Instance instance,
-            string dataElementId)
+            string dataElementId,
+            object data)
```
tjololo: Change method signature to only require what is needed in the method
```suggestion
            Type data)
```

tjololo (changes requested):
We should change the signature of MapModelStateToIssueList to only require the Type of the data object as we don't need anything else in the method

ivarne:
@tjololo Totally agree ❤️ 
I just initially thought the logic would be dependent on the value, not just the type, and then I fixed it for only one level up.

tjololo (approved):
Looks good 👍

PR #150, opened 2023-01-03
Update shared tests and require maxCount for a `Group` component to be repeating

ivarne:
Previously a component was considered a repeating group, even if `maxCount` was not set, but only if the `group` `dataModelBinding` was included. This is wrong, and not the way schemas are interpreted in frontend.

I also included some fixes for code style in tests, and removed some tests that was marked as `[Skiped]` as they will never be active and creates noise in the console when testing.

## Verification
- [ ] All tests run green


RonnyB71 (approved):
🕹️

Issue #153, opened 2023-01-09
Support new PDF service in app lib nugets


RonnyB71:
### Description

The new pdf service is based on chromium

### Additional Information

Epic: https://github.com/Altinn/app-lib-dotnet/issues/63

### Tasks

- [x] Expose new api for frontend to call where we check if there is any custom logic to hide components from PDF (IPdfFormatter). The api should return an object with two lists (pages and components) including the id of the components to be hidden from PDF. #157 
- [x] Call the new PDF service from within the app lib code. Should be possible to feature toggle on/off
- [x] Implement a new PDF client

### Acceptance Criterias

- [x] One PDF generated per datamodel (if more than one)
- [x] It should be easy to disable PDF generation

ivarne:
Do we really want to keep the old “hide component in pdf” logic for the new pdf generator. This new api will require updating of the nuggets anyway, so we have the option to just remove the interface instead.

bjosttveit:
We could remove it, but keeping it makes it easier for existing apps to adopt the new PDF service. It also provides a dynamic way of specifying excluded pages/components as opposed to `Settings.json` which is completely static.

RonnyB71:
Valid question. I'm not sure if we can remove it, we need to make sure we actually cover how this interface has been used in various apps. I imagine you could have some rules for hiding that we don't support directly with the new solution.

PR #157, opened 2023-01-10
PDF format API

bjosttveit:
## Description

Added a new controller with a single endpoint for getting the pdf formatting. Tested manually on the `frontend-test` app. Borrowed a lot from `PdfService`.

## Related Issue(s)
- #153 
- https://github.com/Altinn/app-frontend-react/issues/818

## Verification
- [x] All tests run green

## Documentation


tjololo (changes requested):
I think it would be wise to rename the Controller to `PdfController.cs` so other API regarding PDF have a natural home/grouping in this controller. This would not change the API but make it easier to implement new pdf APIs in the future.

We are moving away from `Newtonsoft.Json` and replacing it with `System.Text.Json` we should use `System.Text.Json` in new code.

src/Altinn.App.Api/Controllers/PdfFormatController.cs
```diff
@@ -0,0 +1,128 @@
...
+    [ApiController]
+    [Route("{org}/{app}/instances/{instanceOwnerPartyId:int}/{instanceGuid:guid}/data/{dataGuid:guid}/pdfformat")]
+    public class PdfFormatController : ControllerBase
+    {
+        private readonly IInstance _instanceClient;
+        private readonly IPdfFormatter _pdfFormatter;
+        private readonly IAppResources _resources;
+        private readonly IAppModel _appModel;
+        private readonly IData _dataClient;
+        private readonly ILogger _logger;
```
tjololo: No logging done, remove it until we actually do some logging

bjosttveit:
> I think it would be wise to rename the Controller to `PdfController.cs` so other API regarding PDF have a natural home/grouping in this controller. This would not change the API but make it easier to implement new pdf APIs in the future.

In that case, should I also change the route of the controller to be more general? maybe something like:

`{org}/{app}/instances/{instanceOwnerPartyId:int}/{instanceGuid:guid}/data/{dataGuid:guid}/pdf`

And then change the route of `GetPdfFormat()` to something like:

`{org}/{app}/instances/{instanceOwnerPartyId:int}/{instanceGuid:guid}/data/{dataGuid:guid}/pdf/format`

tjololo:
> > I think it would be wise to rename the Controller to `PdfController.cs` so other API regarding PDF have a natural home/grouping in this controller. This would not change the API but make it easier to implement new pdf APIs in the future.
> 
> In that case, should I also change the route of the controller to be more general? maybe something like:
> 
> `{org}/{app}/instances/{instanceOwnerPartyId:int}/{instanceGuid:guid}/data/{dataGuid:guid}/pdf`
> 
> And then change the route of `GetPdfFormat()` to something like:
> 
> `{org}/{app}/instances/{instanceOwnerPartyId:int}/{instanceGuid:guid}/data/{dataGuid:guid}/pdf/format`

I totally agree, nice catch!

bjosttveit:
Some of the controllers like `ValidateController` do not have a route for the controller, and each individual endpoint specify the entire route, is that maybe more appropriate? Since we do not know whether future APIs will require all of the same parameters?

src/Altinn.App.Api/Controllers/PdfController.cs
```diff
@@ -0,0 +1,122 @@
...
+            var result = new
+            {
+                ExcludedPages = layoutSettings.Pages.ExcludeFromPdf,
+                ExcludedComponents = layoutSettings.Components.ExcludeFromPdf,
+                PageOrder = layoutSettings.Pages.Order,
+            };
+            return Ok(result);
+        }
+    }
+}
```
tjololo: Add trailing linebreak
```suggestion
}

```

tjololo (approved):
:godmode:

PR #173, opened 2023-01-27
Add new DynamicsPdfFormatter

ivarne:
Read the layout files and hide the relevant pages/components.
We needed this for a big app with lots of dynamics and thought it would be nice to share.

Activation is simple as it is just another service to registrer in `Program.cs`.
```C#
using Altinn.App.Core.Features.Pdf;
...
    services.AddTransient<IPdfFormatterInstance, DynamicsPdfFormatter>();
```

## Description
This ~is~ (was intended to be) a simple `IPdfFormatter` implementation that uses the dynamics infrastructure to figure out what the page/component ids that would be `hidden` in using dynamic `hidden` expressions. Because expressions has `instanceContext` that requires the instance to be available, it required a new interface (`IPdfFormatterInstance`) and some cleanup of the code that calls the pdf formatter.

## Changes
1. New interface `IPdfFormatterInstance` that is a copy of `IPdfFormatter`, but includes the instance, so it can be used for determining what to hide.
2. New `IPdfFormatterInstance` implementation, `DynamicsPdfFormatter` that reads dynamic expressions and hide `hidden` pages and components.
3. Use the shared json deserialization functionality from `AppResourcesSI`, instead of deserializing json in `PdfController` and `PdfService`.
4. Improve `AppResourcesSI` to use a shared `JsonSerializerOptions` instance that accepts comments and trailing commas.
5. Remove `NullPdfFormatter` and make the constructor injection inject optional services instead `IPdfFormatter? pdfFormatter = default,`

If this is somewhat hard to write sensible tests for this functionality (we really need an app and pdf generation to see), so I'll wait for feedback before writing tests.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (will be provided if this is approved)

### DIFF
--- a/src/Altinn.App.Api/Controllers/PdfController.cs
+++ b/src/Altinn.App.Api/Controllers/PdfController.cs
@@ -1,7 +1,6 @@
-using System;
-using System.Linq;
+#nullable enable
+
 using System.Text.Json;
-using System.Threading.Tasks;
 using Altinn.App.Core.Features;
 using Altinn.App.Core.Interface;
 using Altinn.App.Core.Internal.AppModel;
@@ -66,13 +65,13 @@ namespace Altinn.App.Api.Controllers
                 return NotFound("Did not find instance");
             }
 
-            string taskId = instance.Process?.CurrentTask?.ElementId;
+            string? taskId = instance.Process?.CurrentTask?.ElementId;
             if (taskId == null)
             {
                 return Conflict("Instance does not have a valid currentTask");
             }
 
-            DataElement dataElement = instance.Data.FirstOrDefault(d => d.Id == dataGuid.ToString());
+            DataElement? dataElement = instance.Data.FirstOrDefault(d => d.Id == dataGuid.ToString());
             if (dataElement == null)
             {
                 return NotFound("Did not find data element");
@@ -84,20 +83,20 @@ namespace Altinn.App.Api.Controllers
             JsonSerializerOptions options = new() { PropertyNamingPolicy = JsonNamingPolicy.CamelCase };
 
             string layoutSetsString = _resources.GetLayoutSets();
-            LayoutSets layoutSets = null;
-            LayoutSet layoutSet = null;
+            LayoutSets? layoutSets = null;
+            LayoutSet? layoutSet = null;
             if (!string.IsNullOrEmpty(layoutSetsString))
             {
-                layoutSets = JsonSerializer.Deserialize<LayoutSets>(layoutSetsString, options);
+                layoutSets = JsonSerializer.Deserialize<LayoutSets>(layoutSetsString, options)!;
                 layoutSet = layoutSets.Sets?.FirstOrDefault(t => t.DataType.Equals(dataElement.DataType) && t.Tasks.Contains(taskId));
             }
 
-            string layoutSettingsFileContent = layoutSet == null ? _resources.GetLayoutSettingsString() : _resources.GetLayoutSettingsStringForSet(layoutSet.Id);
+            string? layoutSettingsFileContent = layoutSet == null ? _resources.GetLayoutSettingsString() : _resources.GetLayoutSettingsStringForSet(layoutSet.Id);
 
-            LayoutSettings layoutSettings = null;
+            LayoutSettings? layoutSettings = null;
             if (!string.IsNullOrEmpty(layoutSettingsFileContent))
             {
-                layoutSettings = JsonSerializer.Deserialize<LayoutSettings>(layoutSettingsFileContent, options);
+                layoutSettings = JsonSerializer.Deserialize<LayoutSettings>(layoutSettingsFileContent, options)!;
             }
 
             // Ensure layoutsettings are initialized in FormatPdf
@@ -113,8 +112,8 @@ namespace Altinn.App.Api.Controllers
 
             var result = new
             {
-                ExcludedPages = layoutSettings.Pages.ExcludeFromPdf,
-                ExcludedComponents = layoutSettings.Components.ExcludeFromPdf,
+                ExcludedPages = layoutSettings?.Pages?.ExcludeFromPdf ?? new List<string>(),
+                ExcludedComponents = layoutSettings?.Components?.ExcludeFromPdf ?? new List<string>(),
             };
             return Ok(result);
         }
--- a/src/Altinn.App.Api/Controllers/PdfController.cs
+++ b/src/Altinn.App.Api/Controllers/PdfController.cs
@@ -108,7 +108,7 @@ namespace Altinn.App.Api.Controllers
 
             object data = await _dataClient.GetFormData(instanceGuid, dataType, org, app, instanceOwnerPartyId, new Guid(dataElement.Id));
 
-            layoutSettings = await _pdfFormatter.FormatPdf(layoutSettings, data, instance);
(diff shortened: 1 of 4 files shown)


RonnyB71:
I think the new IPdfFormatterInstance interface is better compared to the old as it takes in the instance as well. My question is if we should consider replacing the old instead of creatin a new. This would mean a breaking change, but easy to implement as it's only adding another parameter to the method but is called from a place where we have the instance available.

Another thing to consider is that, although not registered by default, the DynamicsPdfFormatter implementation will be redundant for new applications. The new PDF service will render a PDF view and evaluate the dynamic expressions on frontend. So this implementation should probably only be a part of nuget packages targeting the old PDF service. That said, we still intend to keep the IPdfFormatter as it opens up for other logic for hiding components for the PDF generator.

ivarne:
> My question is if we should consider replacing the old instead of creatin a new

Who knows? It's a change that requires app code to change thus should be accompanied with a major version bump. An alternative to changing the old, or adding a new interface would be to have a new function in the interface with a default implementation `SetInstance(Instnace instance)`. That would be source level backwards compatible, but look a little awkward.

After thinking more about release schedule and the new PDF engine, the best solution is probably just to add the `DynamicsPdfFormatter` implementation to some sort of cookbook, and document that `["instanceContext"]` expressions does not work. They are (probably?) not very common.

ivarne:
I made another attempt at introducing this with much less changes required.

This time I added a new method to the interface with an empty default implementation. This ensures that existing implementations work, and new implementations can receive the instance and use it if required.
```c#
public interface IPdfFormatter
{
    Task<LayoutSettings> FormatPdf(LayoutSettings layoutSettings, object data);
    void SetInstance(Instance instance) {}
 ```

https://github.com/Altinn/app-lib-dotnet/compare/main...ivarne:app-lib-dotnet:simpler_dynamics_in_old_pdf

**Comment by RonnyB71** [comment]
What about adding the new new method to the current `IPdfFormatter` interface like this:

```csharp
Task<LayoutSettings> FormatPdf(LayoutSettings layoutSettings, object data, Instance instance) { throw new NotImplementedException("FormatPdf with instance parameter is not implemented!"); }
```

That way we would keep the interface and only those in need of the new method would have to implement it as the default implementation throws an `NotImplementedException`.

ivarne:
@RonnyB71 That would lead to the interface having 2 methods, where the old one is required and the new one is optional. A typical implementation will need the old method.

```c#
public class CustomPdfFormatter : IPdfFormatter
{
    public async Task<LayoutSettings> FormatPdf(LayoutSettings layoutSettings, object data)
    {
        // this is not used
        throw new NotImplementedException();
    }
    public async Task<LayoutSettings> FormatPdf(LayoutSettings layoutSettings, object data, Instance instance)
    {
        // Implementation
        // ...
        return layoutSettings;
    }
}
```

A better solution would be to just give both methods default implementations, so implementors can chose the method they want to override. The drawback that made me discard this option is that it does not enforce that implementors actually override any method (before it fails runtime). Maybe that isn't a huge problem in practice?
```c#
public interface IPdfFormatter
{
    [Obsolete] // Not sure how this works in an interface?
    async Task<LayoutSettings> FormatPdf(LayoutSettings layoutSettings, object data)
    {
        // One of the methods needs to be overriden
        throw new NotImplementedException();
    }
    async Task<LayoutSettings> FormatPdf(LayoutSettings layoutSettings, object data, Instance instance)
    {
        // Call the older method as a default implementation
        return await FormatPdf(layoutSettings, data);
    }
}
```

RonnyB71:
> A better solution would be to just give both methods default implementations, so implementors can chose the method they want to override. The drawback that made me discard this option is that it does not enforce that implementors actually override any method (before it fails runtime). Maybe that isn't a huge problem in practice?

Yes, it would lead to two methods. But I'm thinking that as a temporary solution until for instance V8 is out and we can have breaking changes. The we would remove the old and leave the new. At the same time we have the option to signal this by deprecating the old, allow for a transition period and eventually just remove the old.

In stead of throwing an exception we could also just have an empty default implementation in the interface. That way everything should work an those who would like to use the instance variant can easily do so. Perhaps moving the "default" implementation from NullPdfFormatter to a default implementation on the interface?

ivarne:
Why do we even have the `NullPdfFormatter`? It's confusing and useless and DI works fine with optional dependencies.

I've drawn up a few options here, all of which works for my purposes, so we just need to make a decision.
1. Create a new interface `IPdfFormatterInstance` (and leave `IPdfFormatter` as is). (**This PR**)
2. Create a new method `SetInstance` on `IPdfFormatter` with a default "do nothing" implementation, so that implementors can record the instance and use it when `FormatPdf` is called. (**Linked branch**)
3. Create a new method `FormatPdf(LayoutSettings layoutSettings, object data, Instance instance)` on `IPdfFormatter`, with a default implementation that calls the old method without instance, and change all app-lib-dotnet code to call the new method.
    1. Default implementation of the old method that throws `NotImplementedException`, so that implementations only need to implement one of the methods. (**Suggested interface in my comment**)
    2. No default implementation of the old method so that implementations need an empty implementation of the old method, even if they use the new signature. (**sample implementation shown**)

Is there any option I missed?

RonnyB71:
> Why do we even have the `NullPdfFormatter`? It's confusing and useless and DI works fine with optional dependencies.
> 
> I've drawn up a few options here, all of which works for my purposes, so we just need to make a decision.
> 
> 1. Create a new interface `IPdfFormatterInstance` (and leave `IPdfFormatter` as is). (**This PR**)
> 2. Create a new method `SetInstance` on `IPdfFormatter` with a default "do nothing" implementation, so that implementors can record the instance and use it when `FormatPdf` is called. (**Linked branch**)
> 3. Create a new method `FormatPdf(LayoutSettings layoutSettings, object data, Instance instance)` on `IPdfFormatter`, with a default implementation that calls the old method without instance, and change all app-lib-dotnet code to call the new method.
>    
>    1. Default implementation of the old method that throws `NotImplementedException`, so that implementations only need to implement one of the methods. (**Suggested interface in my comment**)
>    2. No default implementation of the old method so that implementations need an empty implementation of the old method, even if they use the new signature. (**sample implementation shown**)
> 
> Is there any option I missed?

The `NullPdfFormatter` was added to avoid null checks everywhere and so that the application can always expect one. It's similar to how we ensure that a list is always initialized even if it's empty so that you safely can do a for-each on the list. 

I think the third options is the best way moving forward as it takes us a step in the direction we would like to go eventually dropping the old method and leaving us with the new.

ivarne:
@RonnyB71 Alternaltive 3 has `i` and `ii` options. Would you make your pick, so I can update the PR?

RonnyB71:
> @RonnyB71 Alternaltive 3 has `i` and `ii` options. Would you make your pick, so I can update the PR?

Option `i` sounds like a good way to go.

ivarne:
@RonnyB71 I have updated now.

(Hold on. We need to add LayoutSet to the interface as well. It is obviously required if we have an app with multiple LayoutSets)

ivarne:
Updated with support for `LayoutSet?`. We don't use it yet, so I did not realize there might be a need, and it's silly to leave out when changing the interface.

RonnyB71:
Looks good:) I'll do some testing before I'll merge.

RonnyB71 (approved):
🕹️

PR #175, opened 2023-01-27
Use more relaxed Json parsing in AppResourcesSi.cs

ivarne:
Sometimes people want to use trailing commas or comments in their json files, and I can't really see why that should be a failiure that requires fixing. This fix is also part of https://github.com/Altinn/app-lib-dotnet/pull/173, but seems like it is worth a standalone review.

## Verification
- [ ] All tests run green


RonnyB71 (approved):
🕹️

PR #176, opened 2023-01-30
Nullsafe active instances endpoint

ivarne:
`{org}/{app}/instances/{id}/active` fails with a nullpointerexeption when the lookup for `instance.lastChangedBy` fails. This causes the frontend to crash and never load the app if one of the lookups for active instances fails.
```json
"onEntry": {
    "show": "select-instance"
}
```

## Question
Currently unknown values in `instance.lastChangedBy` results in an empty result.

image

Would it look better to just show the raw value?

image

**Note**: This PR has 2 commits, one significant and the other updating `InstancesController.cs` to `#nullable enable` in an effort to update the project when I touch files.

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -135,7 +136,7 @@ namespace Altinn.App.Api.Controllers
                 Instance instance = await _instanceClient.GetInstance(app, org, instanceOwnerPartyId, instanceGuid);
                 SelfLinkHelper.SetInstanceAppSelfLinks(instance, Request);
 
-                string userOrgClaim = User.GetOrg();
+                string? userOrgClaim = User.GetOrg();
 
                 if (userOrgClaim == null || !org.Equals(userOrgClaim, StringComparison.InvariantCultureIgnoreCase))
                 {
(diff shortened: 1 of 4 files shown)


ivarne:
Seems like @codeql didn't like that I only added nullability checks and didn't remove unused variables and combined if statements. What to do?

RonnyB71:
@Febakke can you take a look at Ivar's question above regarding returning the user id instead of an empty string?

RonnyB71 (approved):
Awaiting feedback from UX regarding question on user id vs empty string.

ivarne:
If https://github.com/Altinn/altinn-studio/pull/9684 is accepted, it might not be very common for the lookup to fail for valid org numbers.

`LastChangedBy` seems to be either an org number 9 digits (for maskinporten authentication of owner orgs), or an userId with unknown number of digits. It would probably only be useful to show org numbers, as UserId is generally unknown for end users. I don't think showing the raw value as fallback makes much sense.

Febakke:
@RonnyB71 
When does this occur?
Im guessing that both possibilities would just read as "Something technical went wrong" by the user. The best would probably be to have some text explaining what the error is. "We had trouble fetching the name of this user" If it is possible to know what happened that is?

RonnyB71:
> @RonnyB71 When does this occur? Im guessing that both possibilities would just read as "Something technical went wrong" by the user. The best would probably be to have some text explaining what the error is. "We had trouble fetching the name of this user" If it is possible to know what happened that is?

Not sure. I'll merge this as it doesn't introduce a change of current behavior.

PR #183, opened 2023-02-06
Added API for fetching footer layout

bjosttveit:
## Description

Added API for fetching footer layout file. Currently a static file, could be defined in layout-sets when that becomes standard. Is it necessary with tests for something as simple as serving one static file?

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/issues/110
- https://github.com/Altinn/app-frontend-react/issues/538

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/ResourceController.cs
+++ b/src/Altinn.App.Api/Controllers/ResourceController.cs
@@ -258,10 +258,15 @@ namespace Altinn.App.Api.Controllers
         /// <param name="app">The application name</param>
         /// <returns>The footer layout in the form of a string.</returns>
         [HttpGet]
-        [Route("{org}/{app}/api/footer")]
+        [Route("{org}/{app}/api/v1/footer")]
         public ActionResult GetFooterLayout(string org, string app)
         {
             string layout = _appResourceService.GetFooter();
+            if (layout is null)
+            {
+                return NoContent();
+            }
+            
             return Ok(layout);
         }
     }
--- a/src/Altinn.App.Api/Controllers/ResourceController.cs
+++ b/src/Altinn.App.Api/Controllers/ResourceController.cs
@@ -250,5 +250,19 @@ namespace Altinn.App.Api.Controllers
 
             return new FileContentResult(fileContent, MimeTypeMap.GetMimeType(".json"));
         }
+
+        /// <summary>
+        /// Get the footer layout
+        /// </summary>
+        /// <param name="org">The application owner short name</param>
+        /// <param name="app">The application name</param>
+        /// <returns>The footer layout in the form of a string.</returns>
+        [HttpGet]
+        [Route("{org}/{app}/api/footer")]
+        public ActionResult GetFooterLayout(string org, string app)
+        {
+            string layout = _appResourceService.GetFooter();
+            return Ok(layout);
+        }
     }
 }
--- a/src/Altinn.App.Api/Controllers/ResourceController.cs
+++ b/src/Altinn.App.Api/Controllers/ResourceController.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Globalization;
 using System.IO;
@@ -70,7 +71,7 @@ namespace Altinn.App.Api.Controllers
         [HttpGet]
         public IActionResult GetRuntimeResource(string id)
         {
-            byte[] fileContent = _appResourceService.GetRuntimeResource(id);
+            byte[]? fileContent = _appResourceService.GetRuntimeResource(id);
 
             if (fileContent != null)
             {
@@ -178,7 +179,7 @@ namespace Altinn.App.Api.Controllers
         [Route("{org}/{app}/api/layoutsettings")]
         public ActionResult GetLayoutSettings(string org, string app)
         {
-            string settings = _appResourceService.GetLayoutSettingsString();
+            string? settings = _appResourceService.GetLayoutSettingsString();
             return Ok(settings);
         }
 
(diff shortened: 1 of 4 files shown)


src/Altinn.App.Core/Interface/IAppResources.cs
```diff
@@ -115,6 +115,12 @@ public interface IAppResources
         /// <returns>The layout sets</returns>
         string GetLayoutSets();
 
+        /// <summary>
+        /// Gets the footer layout
+        /// </summary>
+        /// <returns>The footer layout</returns>
+        string GetFooter();
```
ivarne:
```suggestion
        string? GetFooter();
```

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
```diff
@@ -460,5 +460,18 @@ private byte[] ReadFileContentsFromLegalPath(string legalPath, string filePath)
 
             return null;
         }
+
+        /// <inheritdoc />
+        public string GetFooter()
```
ivarne:
```suggestion
        public string? GetFooter()
```

src/Altinn.App.Api/Controllers/ResourceController.cs
```diff
@@ -250,5 +250,19 @@ public ActionResult GetRuleConfiguration(string org, string app, string id)
...
+        /// </summary>
+        /// <param name="org">The application owner short name</param>
+        /// <param name="app">The application name</param>
+        /// <returns>The footer layout in the form of a string.</returns>
+        [HttpGet]
+        [Route("{org}/{app}/api/footer")]
+        public ActionResult GetFooterLayout(string org, string app)
+        {
+            string layout = _appResourceService.GetFooter();
+            return Ok(layout);
```
ivarne, line 271:
```suggestion
            if (layout is null)
            {
                return NoContent()
            }
            
            return Ok(layout);
```

ivarne (changes requested):
I added some suggestions on how to better handle nullable types.

See also comments in https://github.com/Altinn/app-frontend-react/pull/871#issuecomment-1419096067

src/Altinn.App.Api/Controllers/ResourceController.cs
bjosttveit: I think it already sends a `204` when returning `Ok(null)`.

ivarne: Maybe? It would make kind of sense.

ivarne: It doesn't return a 204 for `Ok(null)`.

src/Altinn.App.Api/Controllers/ResourceController.cs
```diff
@@ -250,5 +250,24 @@ public ActionResult GetRuleConfiguration(string org, string app, string id)
...
+        /// Get the footer layout
+        /// </summary>
+        /// <param name="org">The application owner short name</param>
+        /// <param name="app">The application name</param>
+        /// <returns>The footer layout in the form of a string.</returns>
+        [HttpGet]
+        [Route("{org}/{app}/api/v1/footer")]
+        public ActionResult GetFooterLayout(string org, string app)
+        {
+            string layout = _appResourceService.GetFooter();
```
ivarne: Sorry for missing this in the previous batch. I tend to prefer `var`
```suggestion
            string? layout = _appResourceService.GetFooter();
```

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
```diff
@@ -460,5 +460,18 @@ private byte[] ReadFileContentsFromLegalPath(string legalPath, string filePath)
...
         }
+
+        /// <inheritdoc />
+        public string? GetFooter()
+        {
+            string filename = Path.Join(_settings.AppBasePath, _settings.UiFolder, _settings.FooterFileName);
+            string filedata = null;
+            if (File.Exists(filename))
+            {
+                filedata = File.ReadAllText(filename, Encoding.UTF8);
```
tjololo: We should consider making this async.
```suggestion
                filedata = await File. ReadAllTextAsync(filename, Encoding.UTF8);
```

tjololo (commented):
Most of the other methods in AppResourecsSI isn't async, but we should leverage async methods where available.

ivarne:
@tjololo Or we should just load every file async at startup, and then return them synchronously from memory without any changes to the interface.

olemartinorg:
@ivarne All modern operating systems cache file reads in volatile memory, don't they? I mean, putting this on the heap just serves to reserve that memory so that your app memory usage will be larger than it needs to be (and that object can't easily be evicted if not used, when other apps could need the memory for temporary increased throughput). Imagine the irony of that page of memory being underused and having to be swapped out to disk by the OS... IMO, we shouldn't be aggressively caching things in memory (especially if you're not parsing the file or doing anything computationally intensive with it), especially not if we don't have performance boost figures that defend the increased memory usage.

But, uhh, maybe I'm just overstepping my bounds as a frontend-developer! :sweat_smile:

tjololo:
I'm agreeing with @olemartinorg on this one. Aggressively caching all these files in memory would lead to excessive consumption of memory. One instance of the application container would not seem so bad, but when every app should have at least two replicas and a service owner could have an excess of 100 apps the numbers starts to add up.
If not all applications are continuously used the memory could have been used by other applications in need of it.

tjololo (approved):

PR #185, opened 2023-02-07
Populate InstanceOwner the same way for statelss apps

ivarne:
Currently stateless apps doesn't populate `instance.InstanceOwner` the same way as normal apps does. This means that logic like `["instanceContext","instanceOwnerType"]` and in `dataProcessRead` you can't use `instance.InstanceOwner.OrganisationNumber` for a lookup, as you would do with a normal instance.

## Description
Obviously I wanted to reuse the `InstancesController.SetInstanceOwnerProps` method in `StatelessDataController`, so I rewrote and moved it to `src/Altinn.App.Core/Helpers/InstantiationHelper.cs` as a common reference.

Next I disliked reading the `party` header value from the request object, instead of just as a parameter to the action with a `[FromHeader(Name = "party")] string partyFromHeader` annotation. The tests proved great to discover some edge cases, and I had to rewrite lots of them. One of the tests looked at what happens if a user adds multiple headers. To test that with my change, I had to use a `WebApplicationFactory`, because normal unit testing of controllers doesn't allow duplicate values to a string.

I also looked at the frontend code, and it seems like the backend supports much more than what is actually used. Fronend always specify `party: partyid:123` in a header, but backend also supports
* No header (use party id of the current user)
* `party: org:[orgnummer]`
* `party: person:[ssn]`
 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -1,11 +1,7 @@
 #nullable enable
-using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
+
 using System.Net;
 using System.Text;
-using System.Threading.Tasks;
 using Altinn.App.Api.Helpers.RequestHandling;
 using Altinn.App.Api.Infrastructure.Filters;
 using Altinn.App.Api.Mappers;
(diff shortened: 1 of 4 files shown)


tjololo:
Sorry for the delayed review, had some busy days.
The extra partyId (org/person) I belive is used by API consumers and needs to be there even if frontend does not use them.
Could you have a look at the CodeQL warnings?

ivarne:
> The extra partyId (org/person) I belive is used by API consumers and needs to be there even if frontend does not use them.

I don't really see the use case for API consumers to use the stateless api with the (as far as I can see) undocumented `party` header to lookup some information, like if this was a normal API. Maybe I'm missing something? Anyway, I did preserve this weird behaviour, so it should work as before. Statelss apps seems to only be used by a single app in production, so I think there is still time to suggest reductions in the api surface.

tjololo:
True that! For now lets keep the API for stateless similar to stateful apps even if it makes no sens to use it as an API...

tjololo (approved):

ivarne:
> For now lets keep the API for stateless similar to stateful apps even if it makes no sense to use it as an API...

For stateful apps the API is totally different (you need to post an instance object with either partyId, ssn or orgnumber set in the `InstanceOwner` property). I don't really see any reason for one to use an object and the other a simple header, but that is a different issue. The [docs for stateles](https://docs.altinn.studio/nb/api/apps/stateless-data-elements/) is also missing the essential information about the `party` header, but I guess it kind of works without if you're not representing someone else when using the api.

PR #186, opened 2023-02-07
Added null check to `PdfOptionsMapping`

### LINKED ISSUE #184: Fields without data causes NullReferenceException in PdfOptionsMapping during PDF-generation
### Description of the bug

The method GetMappingValues in PdfOptoinsMapping encounters a NullReferenceException on this line `selectedDatas.Add(jsonData.SelectToken(select).ToString());` when `jsonData.SelectToken(select)` returns `null `and `ToString()` is called on `null`.
This occurs in our application during PDF generation and seems to be related using dynamic options, the keyValuePairs, on checkboxes.

The for-loop in GetMappingValues loops through a set of fields in the data model, if one of these fields contain no data the NullReferenceException occurs.
In our case this occurs because all fields will not always have a value. We can not circumvent this by filling dummy data in relevant fields as some of these have a restriction on legal values which causes validation errors. 
This error first occured in v5 of the Altinn Nuget packages and is still present in v7.

Relevant part of JSON-datamodel:
```
  "soknad": {
    "transportorOrgnummer": "979599498",
    "soknadtype": "new",
    "nyGaranti": {
      "garantinummer": "1",
      "garantitype": "TURVOGN",
      "belop": "1000000",
      "gyldigFraDato": "2023-02-02T12:00:00+01:00",
      "loyvedekning": 0,
      "skjemaBekreft": "confirm"
    },
    "endreGaranti": null,
    "erstattGaranti": null,
    "erstattAndreGarantister": null,
    "skjemaBekreft": null,
    "loyvedekning": null
  }
```
  
Relevant part of layout JSON-file:
```
   {
	"id": "ErstattEgenGarantiCheckboxes",
	"type": "Checkboxes",
	"textResourceBindings": {
	  "description": "Velg hvilke garantier du ønsker å erstatte",
	  "title": "Erstatt garantier"
	},
	"dataModelBindings": {
	  "simpleBinding": "soknad.erstattGaranti.gammelGarantinummer"
	},
	"optionsId": "transportTillatelserErstattEgenGaranti",
	"secure": true,
	"mapping": {
	  "soknad.soknadtype": "soknadtype",
	  "soknad.erstattGaranti.garantitype": "valgtGarantitype",
	  "soknad.transportorOrgnummer": "transportOrgnummer",
	  "orgNoGarantist": "garantistOrgnummer"
	},
	"required": true
  }
```
  
`soknad.endreGaranti` is `null `here which causes the exception in `selectedDatas.Add(jsonData.SelectToken(select).ToString());`

### Steps To Reproduce

Can be reproduced on branch PDF_NullReferenceException of svv/transportloyvegarantier.
1. Start the application and open transportloyvegarantier with the user Sophie Salt representing DDG Fitness.
2. Fill in a 9 digit number in the field "Organisasjonsnummer til transportør"
3. Scroll down and click the button "Ny garanti"
4. Fill al the fields, upload a PDF file, and checkbox at the bottom
5. Click "Send inn"

Can probably be reproduced in any app by adding radio buttons that uses dynamic options in a similar way.

### Additional Information

_No response_

bjosttveit:
## Description

Added a null check, tested on the app mentioned in the linked issue.

## Related Issue(s)
- closes https://github.com/Altinn/app-lib-dotnet/issues/184
- https://altinn.slack.com/archives/C02ESDT071T/p1675343538806029

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Pdf/PdfOptionsMapping.cs
+++ b/src/Altinn.App.Core/Internal/Pdf/PdfOptionsMapping.cs
@@ -153,7 +153,10 @@ public class PdfOptionsMapping : IPdfOptionsMapping
             }
             else
             {
-                selectedDatas.Add(jsonData.SelectToken(select).ToString());
+                JToken? token = jsonData.SelectToken(select);
+                if (token != null) {
+                    selectedDatas.Add(token.ToString());
+                }
             }
         }
 


src/Altinn.App.Core/Internal/Pdf/PdfOptionsMapping.cs
```diff
@@ -153,7 +153,10 @@ private static List<string> GetMappingValues(JObject jsonData, KeyValuePair<stri
             }
             else
             {
-                selectedDatas.Add(jsonData.SelectToken(select).ToString());
+                JToken? token = jsonData.SelectToken(select);
+                if (token != null) {
+                    selectedDatas.Add(token.ToString());
```
ivarne, line 158: Not related to this PR, but I'm not looking forward to the day someone discovers that `.ToString()` on a number is dependent on the thread locale settings to figure out if the number is `0.1` or `0,1`

olemartinorg: Been there! :cry: I once spent an entire day trying to figure out why a calendar application wasn't rendering at all. No error messages, no hints. Dissecting commits revealed the cause was one commit (where the locale changed), and the cause of the rendering bug was some dynamically generated CSS that now contained commas instead of periods. Took me quite a long time squinting to find those pixels that marks the difference between a comma and a period. :joy: Learned a lot that day.

RonnyB71 (approved):
🕹️

Issue #63, opened 2022-10-25
PDF from HTML


RonnyB71:
## Description

This epic is part of the roadmap issue https://github.com/Altinn/altinn-roadmap/issues/38 which is rewrite of today's PDF generator.

The architecture is discussed here https://github.com/Altinn/altinn-pdf/issues/16, but in short it's based on the app calling a PDF microservice providing it with a url in the app. The PDF microservice calls back to the app using the provided url, reads the generated HTML and converts to PDF which is returned to the app.

## Primary goal
* Out-of-the-box generation of a PDF that is visually presentable and accessible without the need to configure anything

## In scope
* One PDF per data model will be generated (as today)
* A default PDF generation should "just work" out of the box using an autogenerated summary page based on the layout of the data process step
  * The page needs to be printer friendly
  * Dynamics rules must be evaluated as part of the PDF generation
  * Current exclude from PDF should still work and should be evaluated after any dynamic rules
  * The PDF should have a first-page/header with information from the instance (id, timestamp, name of app, name of service owner, name of user, list of attachments, contact information to the service owner) and any other information relevant to the user.
  * The PDF should adhere to Elmers requirements for a receipt ref. https://brukskvalitet.brreg.no/elmer3/#Innsending
* Override the default generated summary page by either
  * Specify a summary page with a custom layout
* Masking of sensitive data should be supported by the components and respected by the PDF generation
* It should be possible to disable PDF generation
* Override the default generated summary page by either
  * Specify a summary page with a custom layout

### Other Requirements
- [ ] Long response times while generating PDFs needs to be handled with regards to the user
- [ ] Robust handling of services being slow/down eg. if external services are used as dynamic options

### Out of scope
* Override the default generated summary page by either
  * Specify a summary page with a custom layout
  * Use the Confirmation process step with a custom layout
* PDF generation in the app should be implemented as a BPMN Service task visible in the BPMN process
  * Navigation back in the process (after the PDF has been generated) should be supported, ie. a new PDF should be generated and presumably the existing deleted
* Asynchronous PDF generation - while this probably is the way we would like to go at some point - it introduces some big changes on the platform side
  * Ques would need to be introduced as part of the solution
  * The PDF service would need to be able to authenticate on behalf of it-self, and not the user when calling the app
  * The app needs to be able to get user data authenticated as the app and not the user. 
* Multiple data models within the same PDF
  * Frontend can only render data from one model at a time
* Hosting PDF service using Azure Container Apps
[...]

PR #193, opened 2023-02-14
Improve settings in apps

ivarne:
The sum of these changes is to reduce the amount of boilerplate config that is required for an app to work. This is especially important because we introduce a the new pdf functionality and want to make updating as simple as possible, so we can spend time testing and verifying, instead of fiddeling and doing code review on config changes.

This pr (and following work )will reduce the content of https://docs.altinn.studio/nb/app/development/ux/pdf/local-development/ to just add the following in `appsettings.json`

```json
"FeatureManagement": {
    "NewPdfGeneration":  true
}
```

## Hostname
The setting `GeneralSettings.HostName` was previously used with just part of full hostname, and used for cookies. When we later try to use it to get the self url (for events and pdf), we need to add logic to the app for production to add `https://{org}.apps.` in front of the hostname. I'm proposing a new config that includes the full external url for the app `PlatformSettings.ExternalAppBaseUrl`. As I really hate configuring the same info multiple times, I added support for replacements through the helper method `PlatformSettings.FormattedExternalAppBaseUrl`. One might do this replacements at app startup and drop the need for a helper method. Not sure what you prefer?

## PdfSettings.AppPdfPageUriTemplate
This setting seemed rather ill considered, becuase it includes both the internal url used for app routing `#/instance/{instanceId}` that obviously should be the same in any environment (only set once), and a base url that needs to be different in different environments. My suggestion is to move the base url part to PlatformSettings (different between environments), and have a single config in the PdfSettings for just the internal route.

## Defaults in code for all settings
The app template contains defaults for lots of settings. In an effort to reduce the size of apps (mostly for review purposes), these defaults that I can't really see when I would override from the default value set in the template.
* `GeneralSettings.SoftValidationPrefix = "*WARNING*";`
* `GeneralSettings.FixedValidationPrefix = "*FIXED*"`
*  `GeneralSettings.HostName = "local.altinn.cloud";` (this is presumably overridden in tt02/prod)
*  `GeneralSettings.AltinnPartyCookieName = "AltinnPartyId";`
*  `GeneralSettings.TemplateLocation` (removed, as I found no trace of usage)

## Required changes in platform
This PR creates 2 new configurations in platformSettings that needs to be overridden using environment variables when running tt02/prod
* `PlatformSettings.BrowserPdfEndpoint` (moved from `PdfGeneratorSettings.ServiceEndpointUri`). This is obviously part of the platform, and one of the settings you need to configure for an app to work in an environment.
* `PlatformSettings.ExternalAppBaseUrl` (extracted from part of `AppPdfPageUriTemplate`, and is an extention of the config required for `HostName`.) This allows apps to not hard code assumptions about how the production urls will look in the future.

## Related Issue(s)
- #63 

## Verification
- [x] All tests run green


ivarne:
Sorry for the extra force push. I have not changed anything. I have completed a suggestion for how we can allow all (or a limited set) configs to use `{org}` and `{app}` replacements without requiring the replacements to be performed at all usage sites. https://github.com/ivarne/app-lib-dotnet/compare/pdf_settings...ivarne:app-lib-dotnet:config-override

RonnyB71:
Looks good:) Would like to keep the PdfGeneratorSetting in the pdf namespace where it was. Moving forward we are trying to keep the releated non-generic classes in the same namespace.

ivarne:
@RonnyB71 I moved the file that matches the configuration back to the `Altinn.App.Core.Internal.Pdf` namespace.

RonnyB71 (dismissed):
🕹️

ivarne:
@RonnyB71 Any hint about what the issues was?

I quite enjoy seeing colleagues struggle for a day with an app that doesn’t work properly, and then being able to fix it in 2 minutes by looking at some boilerplate code/config that they copied wrong, but Altinn has enough such cases already, so I really don’t need more traps to prove myself valuable.

src/Altinn.App.Core/Configuration/PlatformSettings.cs
```diff
@@ -9,37 +12,66 @@ public class PlatformSettings
...
         /// <summary>
         /// Gets or sets the the url for the PDF API endpoint.
         /// </summary>
-        public string ApiPdfEndpoint { get; set; }
+        public string ApiPdfEndpoint { get; set; } = "http://localhost:5070/api/v1/";
+
+        /// <summary>
+        /// Gets or sets tue url for the new browser based PDF API endpoint.
+        /// </summary>
+        public string BrowserPdfEndpoint { get; set; } = "http://localhost:5300/pdf";
```
RonnyB71:
```suggestion
        public string ApiPdf2Endpoint { get; set; } = "http://localhost:5300/pdf";
```
ApiPdf2Endpoint to be consistent with previous naming and indicate that this is a new version of the pdf service.

src/Altinn.App.Core/Configuration/PlatformSettings.cs
```diff
@@ -9,37 +12,66 @@ public class PlatformSettings
...
+        /// The externally accesible base url for the app with trailing /
+        /// </summary>
+        /// <remarks>
+        /// This setting offers the following replacemnts
+        /// <br />
+        /// {hostName}: GeneralSettings::Hostname<br />
+        /// {org}: Org from applicationmetadata.json<br />
+        /// {app}: App from applicationmetadata.json<br />
+        /// </remarks>
+        public string ExternalAppBaseUrl { get; set; } = "http://{hostName}/{org}/{app}/";
```
RonnyB71: Would like this to be moved to GeneralSettings as it's not a platform thing. And then the convenience method below should follow.

src/Altinn.App.Core/Configuration/PlatformSettings.cs
ivarne: Seems very reasonable. Version numbers is probably better than trying to figure out proper naming. We should probably be more consistent to always use pdf2. At least LocalTest uses pdf vs pdf-service, which is rather confusing.

src/Altinn.App.Core/Configuration/PlatformSettings.cs
ivarne: That depends on how you read platform. I placed it here because the app platform needs to override this. I’m about to suggest a way to move the remaining settings that gets overridden here too.

RonnyB71: The way we read platform from the app perspective is the underlying services provided to us.

RonnyB71:
> @RonnyB71 Any hint about what the issues was?
> 
> I quite enjoy seeing colleagues struggle for a day with an app that doesn’t work properly, and then being able to fix it in 2 minutes by looking at some boilerplate code/config that they copied wrong, but Altinn has enough such cases already, so I really don’t need more traps to prove myself valuable.

I only had a short time to create a package and test in on an app. Just concluded that it failed and didn't have time to look into it. So don't know if it was me, my app or your changes. Giving it another go after we land this and will of course give you a hint if I find anything.

src/Altinn.App.Core/Configuration/PlatformSettings.cs
ivarne: Moved.

RonnyB71 (approved):
🕹️

PR #194, opened 2023-02-14
Fix default `AppPdfPageUriTemplate`

bjosttveit:
## Description

I enabled the new PDF-generator in `frontend-test` and deployed to tt02. Then I noticed that the PDF did not use the PDF-view to generate, just the regular form. The default URL was missing `?pdf=1`.

I think the example in the docs is also incorrect, as the URL in the "default settings" was different from what is here.
See: https://github.com/Altinn/altinn-studio-docs/pull/847


ivarne:
Please see https://github.com/Altinn/app-lib-dotnet/pull/193. `AppPdfPageUriTemplate` is two parts with different scope (base url, different between environments and path which is constant between all envrionments).

RonnyB71 (approved):

PR #201, opened 2023-03-02
Extract metadata methods from IAppResources to separate interface and add feature flags section (resolves #188)

### LINKED ISSUE #188: FeatureSet
### Description

We have a concrete need in `app-frontend-react` to treat older and newer versions of the backend differently depending on which features it supports. This because our releases are asynchronous, and some of our (frontend) features depend on corresponding implementations on your end.

The way we expect to solve this is by ~~requesting `api/v1/featureset` (an open API on your end)~~, which should respond with a JSON object indicating which features exist (and are enabled) on the backend. An example of such a reply would be:

```json
{
  "multiPartSave": true,
  "footer": true
}
```

When the endpoint is implemented, it should start by replying with enabled features like:
- `"footer": true` when #183 is merged
- `"multiPartSave": true` when #187 is implemented

### Additional Information

- Altinn/app-frontend-react/issues/810 (the idea originated in a comment here)
- Altinn/app-frontend-react/pull/899 (our client-side implementation of #187)
- Altinn/app-frontend-react/pull/871 (an example where this would be needed)
- #183
- #187

tjololo:
## Description
Application metadata response extends Application model from Altinn.Platform.Storage.Interface.Models and adds list of supported and enabled features necessary for frontend to support multiple backends

## Related Issue(s)
Fixes #188 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -13,6 +13,7 @@ using Altinn.App.Core.Features;
 using Altinn.App.Core.Helpers;
 using Altinn.App.Core.Helpers.Serialization;
 using Altinn.App.Core.Interface;
+using Altinn.App.Core.Internal.App;
 using Altinn.App.Core.Internal.AppModel;
 using Altinn.App.Core.Models;
 using Altinn.App.Core.Models.Validation;
@@ -23,12 +24,10 @@ using Altinn.Common.PEP.Models;
 using Altinn.Platform.Profile.Models;
 using Altinn.Platform.Register.Models;
 using Altinn.Platform.Storage.Interface.Models;
-
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.Primitives;
-
 using Newtonsoft.Json;
 
 namespace Altinn.App.Api.Controllers
@@ -52,7 +51,7 @@ namespace Altinn.App.Api.Controllers
         private readonly IEvents _eventsService;
         private readonly IProfile _profileClientClient;
 
-        private readonly IAppResources _appResourcesService;
+        private readonly IAppMetadata _appMetadata;
         private readonly IAppModel _appModel;
         private readonly IInstantiationProcessor _instantiationProcessor;
         private readonly IInstantiationValidator _instantiationValidator;
@@ -71,7 +70,7 @@ namespace Altinn.App.Api.Controllers
             IRegister registerClient,
             IInstance instanceClient,
             IData dataClient,
-            IAppResources appResourcesService,
+            IAppMetadata appMetadata,
             IAppModel appModel,
             IInstantiationProcessor instantiationProcessor,
             IInstantiationValidator instantiationValidator,
@@ -85,7 +84,7 @@ namespace Altinn.App.Api.Controllers
             _logger = logger;
             _instanceClient = instanceClient;
             _dataClient = dataClient;
-            _appResourcesService = appResourcesService;
+            _appMetadata = appMetadata;
             _registerClient = registerClient;
             _appModel = appModel;
             _instantiationProcessor = instantiationProcessor;
@@ -175,7 +174,7 @@ namespace Altinn.App.Api.Controllers
                 return BadRequest("The path parameter 'app' cannot be empty");
             }
 
-            Application application = _appResourcesService.GetApplication();
+            Application? application = await _appMetadata.GetApplicationMetadata();
             if (application == null)
             {
                 return NotFound($"AppId {org}/{app} was not found");
@@ -332,9 +331,9 @@ namespace Altinn.App.Api.Controllers
         [ProducesResponseType(StatusCodes.Status400BadRequest)]
         [RequestSizeLimit(RequestSizeLimit)]
         public async Task<ActionResult<Instance>> PostSimplified(
-        [FromRoute] string org,
-        [FromRoute] string app,
-        [FromBody] InstansiationInstance instansiationInstance)
+            [FromRoute] string org,
+            [FromRoute] string app,
+            [FromBody] InstansiationInstance instansiationInstance)
         {
             if (string.IsNullOrEmpty(org))
             {
@@ -346,7 +345,7 @@ namespace Altinn.App.Api.Controllers
                 return BadRequest("The path parameter 'app' cannot be empty");
             }
 
-            Application application = _appResourcesService.GetApplication();
+            Application? application = await _appMetadata.GetApplicationMetadata();
             if (application == null)
             {
                 return NotFound($"AppId {org}/{app} was not found");
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -174,11 +174,7 @@ namespace Altinn.App.Api.Controllers
                 return BadRequest("The path parameter 'app' cannot be empty");
             }
 
-            Application? application = await _appMetadata.GetApplicationMetadata();
-            if (application == null)
-            {
-                return NotFound($"AppId {org}/{app} was not found");
-            }
+            ApplicationMetadata application = await _appMetadata.GetApplicationMetadata();
 
             MultipartRequestReader parsedRequest = new MultipartRequestReader(Request);
             await parsedRequest.Read();
@@ -345,11 +341,7 @@ namespace Altinn.App.Api.Controllers
                 return BadRequest("The path parameter 'app' cannot be empty");
             }
 
-            Application? application = await _appMetadata.GetApplicationMetadata();
-            if (application == null)
-            {
-                return NotFound($"AppId {org}/{app} was not found");
-            }
+            ApplicationMetadata application = await _appMetadata.GetApplicationMetadata();
 
             bool copySourceInstance = !string.IsNullOrEmpty(instansiationInstance.SourceInstanceId);
             if (copySourceInstance && application.CopyInstanceSettings?.Enabled != true)
@@ -483,10 +475,10 @@ namespace Altinn.App.Api.Controllers
             return Created(url, instance);
         }
 
-        private async Task CopyDataFromSourceInstance(Application application, Instance targetInstance, Instance sourceInstance)
+        private async Task CopyDataFromSourceInstance(ApplicationMetadata application, Instance targetInstance, Instance sourceInstance)
         {
             string org = application.Org;
-            string app = application.Id.Split("/")[1];
+            string app = application.App;
             int instanceOwnerPartyId = int.Parse(targetInstance.InstanceOwner.PartyId);
 
             string[] sourceSplit = sourceInstance.Id.Split("/");
@@ -792,7 +784,7 @@ namespace Altinn.App.Api.Controllers
             }
         }
 
-        private async Task StorePrefillParts(Instance instance, Application appInfo, List<RequestPart> parts)
+        private async Task StorePrefillParts(Instance instance, ApplicationMetadata appInfo, List<RequestPart> parts)
         {
             Guid instanceGuid = Guid.Parse(instance.Id.Split("/")[1]);
             int instanceOwnerIdAsInt = int.Parse(instance.InstanceOwner.PartyId);
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -478,7 +478,7 @@ namespace Altinn.App.Api.Controllers
         private async Task CopyDataFromSourceInstance(ApplicationMetadata application, Instance targetInstance, Instance sourceInstance)
         {
             string org = application.Org;
-            string app = application.App;
+            string app = application.AppIdentifier.App;
             int instanceOwnerPartyId = int.Parse(targetInstance.InstanceOwner.PartyId);
 
             string[] sourceSplit = sourceInstance.Id.Split("/");
--- a/src/Altinn.App.Core/Implementation/AppResourcesSI.cs
+++ b/src/Altinn.App.Core/Implementation/AppResourcesSI.cs
@@ -109,14 +109,21 @@ namespace Altinn.App.Core.Implementation
         /// <inheritdoc />
         public Application GetApplication()
         {
-            var task = Task.Run<ApplicationMetadata?>(async () => await _appMetadata.GetApplicationMetadata());
-            task.Wait();
-            if (task.IsCompletedSuccessfully)
+            try
             {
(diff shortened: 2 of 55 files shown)


tjololo:
Alot of nullability checks due to the fact that if read of applicationmetadata.json fails we return null.
Have created an issue #202 for this

src/Altinn.App.Core/Infrastructure/Clients/Register/PersonClient.cs
```diff
@@ -64,16 +63,17 @@ public class PersonClient : IPersonRetriever
         /// <inheritdoc/>
         public async Task<Person?> GetPerson(string nationalIdentityNumber, string lastName, CancellationToken ct)
         {
-            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, $"persons");
-
-            await AddAuthHeaders(request);
+            using(var request = new HttpRequestMessage(HttpMethod.Get, $"persons"))
+            {
```
ivarne: There is now new syntax for a using declaration that creates less indentation https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement
```suggestion
            using var request = new HttpRequestMessage(HttpMethod.Get, $"persons");
```

Just skip the `(` after using and end the line with a `;`

tjololo: Nice! Thanks, great readability improvement!

ivarne: UUps, Should be sure to create a full example when using the suggest feature.

tjololo:
> Alot of nullability checks due to the fact that if read of applicationmetadata.json fails we return null. Have created an issue #202 for this

I have tested renaming and supplying invalid applicationmetdata.json files in an application. 
Missing file result in crash at startup. Missconfigures (eg: string in an int field) result in crash when instance is created. 
I think we can safely remove the nullable from GetApplicationMetdata and have it throw and Exception instead. 
This would remove the null checks that only occurs when the application crashes either way.

What do you think @RonnyB71 ?

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
```diff
@@ -104,66 +109,49 @@ public byte[] GetText(string org, string app, string textResource)
...
             }
             catch (Exception ex)
             {
-                _logger.LogError("Something went wrong when fetching application metadata. {0}", ex);
-                return null;
+                _logger.LogError(ex, "Something went wrong fetching application metadata");
+                return null;   
             }
+            _logger.LogError("Something went wrong fetching application metadata");
+            return null;
```
tjololo: @RonnyB71 should we just ignore these exceptions here as well? Would change how the class works but we are no longer returning null on a non-nullable return type

RonnyB71: Yes, I think so. Does not make sense to swallow the exception here and let it blow up somewhere else.

src/Altinn.App.Core/Internal/App/IAppFeatures.cs
```diff
@@ -0,0 +1,14 @@
...
+    /// <summary>
+    /// Interface reporting features needed by frontend and their status to support multiple versions of the backend
+    /// </summary>
+    public interface IAppFeatures
+    {
+        /// <summary>
+        /// Fetch frontend features that are supported by this backend
+        /// </summary>
+        /// <returns>List of frontend features enabled/disabled for this backend</returns>
+        public Task<Dictionary<string, bool>> GetEnabledFeatures();
```
RonnyB71: Does it get enabled features or enabled/disabled features - perhaps rename to GetFeatures?

src/Altinn.App.Core/Internal/App/AppMetadata.cs
```diff
@@ -0,0 +1,118 @@
...
+            try
+            {
+                if (File.Exists(filename))
+                {
+                    return await File.ReadAllTextAsync(filename, Encoding.UTF8);
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Something went wrong when fetching BPMNProcess");
```
RonnyB71: Should we do the same for BPMN as for ApplicationMetadata? The app will crash without it so might as well let it crash and burn.

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
RonnyB71: And let's create version 8 just to signal that we are changing behavior even though it would crash in a different place anyway.

src/Altinn.App.Core/Internal/App/AppFeatures.cs
```diff
@@ -0,0 +1,24 @@
+namespace Altinn.App.Core.Internal.App
+{
+    /// <summary>
+    /// Default implementation of IAppFeatures
+    /// </summary>
+    public class AppFeatures : IAppFeatures
```
RonnyB71: We should perhaps document the difference between this feature handling and the FeatureManagement use somewhere in the code as it's there it might be confusing for others (and future me).

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
```diff
@@ -104,66 +109,49 @@ public byte[] GetText(string org, string app, string textResource)
...
-
-                _application = JsonConvert.DeserializeObject<Application>(filedata)!;
-                return _application;
             }
             catch (Exception ex)
             {
-                _logger.LogError("Something went wrong when fetching application metadata. {0}", ex);
-                return null;
+                _logger.LogError(ex, "Something went wrong fetching application metadata");
+                return null;
```
RonnyB71: Let it crash and burn?

src/Altinn.App.Core/Internal/App/AppMetadata.cs
ivarne: We should do an app validation on startup to check everything as I suggested in https://github.com/Altinn/app-lib-dotnet/pull/98

src/Altinn.App.Core/Models/ApplicationMetadata.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// <summary>
+        /// List of features and status (enabled/disabled)
+        /// </summary>
+        [JsonProperty(PropertyName = "features")]
+        public Dictionary<string, bool>? Features { get; set; }
+
+        /// <summary>
+        /// Name of the application
+        /// </summary>
+        public string App { get; set; } = "";
```
RonnyB71: Should we consider using AppIdentifier instead of continuing to require both app and org in signatures or taking the whole ApplicationMetadata object when you only needs those two in combination?

src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -484,10 +475,10 @@ public class InstancesController : ControllerBase
             return Created(url, instance);
         }
 
-        private async Task CopyDataFromSourceInstance(Application application, Instance targetInstance, Instance sourceInstance)
+        private async Task CopyDataFromSourceInstance(ApplicationMetadata application, Instance targetInstance, Instance sourceInstance)
```
RonnyB71, line 478: An example where I think a dependency on ApplicationMetadata is too broad since we only need the app/org or AppIdentifier.

src/Altinn.App.Core/Internal/App/AppMetadata.cs
tjololo: I totally agree we should add some sort of startup validation. If we could make it easy to extend the startup validation we could tackle the files one at a time and also give each app developer the possibility to extend what checks are triggered at startup.

src/Altinn.App.Core/Internal/App/AppFeatures.cs
tjololo: Should we consider renaming our implementation to capabilities to make the likelihood of confusing it with FeatureManagement?
@olemartinorg if we decide to rename this, does it seem like a better name in appmetadata? we are reporting if the backend has the capability X not a feature?

src/Altinn.App.Core/Internal/App/IAppFeatures.cs
tjololo: Agreed, will change it as soon as we conclude if we should rename features to capabilities

src/Altinn.App.Core/Internal/App/AppFeatures.cs
olemartinorg: Fine by me - but we'd need to change it in the next release of app-frontend in that case (the multipart-save feature/capability is already in production).

Let's consider the two cases we have now:
- Footer. This is a new capability the backend magically gets in a new version.
- Multipart saving (should probably be named "changelist in data write" or something on the backend), which is a capability you should be able to turn on/off in apps (and which is on by default for new apps, and will be the only option in some future major version).

From that, I think 'features' sound better, because features can be toggled and capabilities sounds like they cannot be toggled. The difference is that `FeatureManagement` lets an app developer control the backend/frontend features of their app, while `AppFeatures` lets the backend toggle features in app-frontend. So maybe `FrontendFeatures` is a better name? 🤷

tjololo: Agreed, was only thinking of footer when I came up with capabilities. FrontendFeatures sound better as our intent is to report what frontend features the backend supports and has enabled. Any objections to FronendFeatures @RonnyB71 ?

src/Altinn.App.Core/Internal/App/AppMetadata.cs
RonnyB71: Sounds  good. But I guess we would like to get this out of the door and not add any more changes? Still good to just throw the exception where it happens.

src/Altinn.App.Core/Internal/App/AppFeatures.cs
RonnyB71: Yes. I would like it to be FrontendFeatures ;-)

tjololo: Renamed it to FrontendFeatures

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
tjololo: throwing ApplicationConfigException, the interface is promising a value so I think we don't need to update the major version. Either way the application will crash

src/Altinn.App.Core/Internal/App/AppMetadata.cs
tjololo: Created an issue to track validation on startup #208 to not make this PR even larger

src/Altinn.App.Core/Internal/App/AppMetadata.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                if (File.Exists(filename))
+                {
+                    return await File.ReadAllTextAsync(filename, Encoding.UTF8);
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Something went wrong when fetching XACML Policy");
+            }
```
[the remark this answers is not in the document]
RonnyB71: Perhaps add a comment why we allow swallowing this exception and include the file path in the  message as well.

src/Altinn.App.Core/Internal/App/AppMetadata.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                return _application;
+            }
+
+            string filename = _settings.AppBasePath + _settings.ConfigurationFolder + _settings.ApplicationMetadataFileName;
+            try
+            {
+                if (File.Exists(filename))
+                {
+                    JsonSerializerOptions jsonSerializerOptions = new JsonSerializerOptions()
```
RonnyB71, line 47: Ideally we should have these as default settings thus avoiding having to repeat them, but I'm not sure if it's possible. Just something we could keep in mind if we end up repeating these all over.

src/Altinn.App.Core/Internal/App/AppMetadata.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <exception cref="System.IO.FileNotFoundException">Thrown if applicationmetadata.json file not found</exception>
+        public async Task<ApplicationMetadata> GetApplicationMetadata()
+        {
+            // Cache application metadata
+            if (_application != null)
+            {
+                return _application;
+            }
+
+            string filename = _settings.AppBasePath + _settings.ConfigurationFolder + _settings.ApplicationMetadataFileName;
```
RonnyB71: Not you code, but a Path.Join would be good here.

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
```diff
@@ -104,67 +109,52 @@ public byte[] GetText(string org, string app, string textResource)
...
-                if (File.Exists(filename))
-                {
-                    return File.ReadAllText(filename, Encoding.UTF8);
-                }
-            }
-            catch (Exception ex)
-            {
-                _logger.LogError("Something went wrong when fetching XACML Policy. {0}", ex);
+                return task.Result;
             }
```
RonnyB71: Same here, worth a comment on why we allow swallowing this particular exception while the others are thrown.

RonnyB71 (approved):

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
tjololo: Crash and burn implemented in latest version :)

src/Altinn.App.Core/Internal/App/AppMetadata.cs
tjololo: Nice catch, replacing all similar lines in AppMetadata with Path.Join

src/Altinn.App.Core/Internal/App/AppMetadata.cs
tjololo: Agreed, have removed the generic exception part as I didn't affect anything authorisation wise

src/Altinn.App.Api/Controllers/InstancesController.cs
tjololo: Agreed, I belive this is a weakness in the Application model, we have and Id containing org and appname (org/appname) but also a separate field for org. These could potentially become out of sync and return different orgs.
I have a suggestion for adding AppIdentifier to ApplicationMetadata class instead of string App

src/Altinn.App.Core/Models/ApplicationMetadata.cs
tjololo: Made some changes and added AppIdentifier instead of string App. Override set of Id to ensure AppIdentifier is updated if Id is updated

PR #217, opened 2023-03-29
Fix TODO required

ivarne:
Return a sensible description. Validation currently does not know the language, but frontend can probably safely ignore required from backend, because it calculates dynamics and required in js.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Expressions/LayoutEvaluator.cs
+++ b/src/Altinn.App.Core/Internal/Expressions/LayoutEvaluator.cs
@@ -99,13 +99,14 @@ public static class LayoutEvaluator
                 {
                     if (state.GetModelData(binding, context) is null)
                     {
+                        var field = state.AddInidicies(binding, context);
                         validationIssues.Add(new ValidationIssue()
                         {
                             Severity = ValidationIssueSeverity.Error,
                             InstanceId = state.GetInstanceContext("instanceId").ToString(),
                             DataElementId = dataElementId,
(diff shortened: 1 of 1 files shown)


src/Altinn.App.Core/Internal/Expressions/LayoutEvaluator.cs
```diff
@@ -99,18 +99,19 @@ private static void RunLayoutValidationsForRequiredRecurs(List<ValidationIssue>
...
+                        var field = state.AddInidicies(binding, context);
                         validationIssues.Add(new ValidationIssue()
                         {
                             Severity = ValidationIssueSeverity.Error,
                             InstanceId = state.GetInstanceContext("instanceId").ToString(),
                             DataElementId = dataElementId,
-                            Field = state.AddInidicies(binding, context),
-                            Description = "TODO required",
+                            Field = field,
+                            Description = $"{field} is required in component with id {context.Component.Id}",
```
RonnyB71, line 109: If I remember correctly we use a text resource id in the description field in order to be able to localize the error messages.

ivarne: Yes, but I hope that in the future we’ll use `code` for that purpose. (At least if it says `required`)

see https://github.com/Altinn/app-frontend-react/issues/1062

RonnyB71 (approved):

PR #219, opened 2023-03-29
Avoid Task.Run when calling methods from IAppMetadata

RonnyB71:
## Description
This changes the way we call async methods from IAppMetadata within synchronous methods in AppResourcesSi. The newly introduced way was based on Task.Run which performance test showed to be a critical bottleneck.

## Verification
- [x] All tests run green

## Documentation


SandGrainOne (approved):

ivarne:
This would have been so much simpler if we could just load all these files into memory on startup.

ivarne:
I always thought async over sync escape hatches as`.Result` was really bad to have in  production code, because you risk thread pool issues and deadlocks?

PR #224, opened 2023-03-30
Add GET endpoint to trigger copy of instance

SandGrainOne:
## Description
Adding a new GET endpoint in InstanceController. The new endpoint is intended to be an easy to access function to create a new instance based on an existing instance. Data is copied from one to the other based on CopyInstanceSettings.

## Related Issue(s)
- https://github.com/Altinn/altinn-platform/issues/355

## Verification
- [ ] All tests run green

## Documentation


ivarne:
Doesn’t mutating apis usually try to avoid using the GET method?

tjololo:
> Doesn’t mutating apis usually try to avoid using the GET method?

To explain the GET suggestion it is due to technical challenges in Altinn2. We need to have a discussion in the team as to how we should solve this.
I think it's a bad idea to have technical dept in an legacy system result in technical dept in a new system.

ivarne:
There are two standard ways this could work without xsrf issues.
1. "altinn2" does an authenticated backchannel call to the app to an instantiation endpoint that accepts an existing instance to copy from and then redirects the user to the correct url for the newly created instance.
2. "altinn2" redirects the user to a get endpoint that does no mutation, but then asks the user to trigger a Post with the action.

I get that the first might be limited by altinn2 capabilities, but the manual confirmation step is just as simple on the altinn 2 side. It should probably be implemented in app-frontend-react, but could just be a simple form a user might click submit on.
Just redirecting a user to a GET endpoint that does the mutation was much worse before "same-orgin" cookies, but still don't feel like a safe thing to do.

src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -475,66 +462,100 @@ public class InstancesController : ControllerBase
...
+        /// <param name="org">Unique identifier of the organisation responsible for the app</param>
+        /// <param name="app">Application identifier which is unique within an organisation</param>
+        /// <param name="instanceOwnerPartyId">Unique id of the party that is the owner of the instance</param>
+        /// <param name="instanceGuid">Unique id to identify the instance</param>
+        /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+        /// <remarks>
+        /// The endpoint will return a redirect to the new instance if the copy operation was successful.
+        /// </remarks>
+        [Authorize]
+        [HttpGet("{instanceOwnerPartyId:int}/{instanceGuid:guid}/copy")]
```
tjololo: Should prefix this path with legacy to make it obvious that this should not be used.
Exclude this from OpenAPI specs and mark the method as obsolete.

tjololo (changes requested):
Not happy with mutating GET endpoint, but as the old system has limitations we would need to accept technical debt somewhere. By accepting the dept here we would not introduce dept in both frontend and backend

src/Altinn.App.Api/Controllers/InstancesController.cs
tjololo: New endpoint would be: `/legacy/instances/<org>/<app>/<ownerPartyID>/<instanceGUID>/copy`

ivarne:
See also the suggestion in https://github.com/Altinn/app-frontend-react/issues/286

> The workflow should be that triggering copy in SBL should send a request to app frontend. That again builds the request to app backend and renders the newly created instance.

Personally I think frontend should ask the user for a confirmation that they want to copy the instance (unless altinn2 is able to provide a signed assertion), so that there isn't any risk that third party websites can trigger copies with user cookies without user involvement.

src/Altinn.App.Api/Controllers/InstancesController.cs
SandGrainOne: Made it: `/<org>/<app>/legacy/instances/<ownerPartyID>/<instanceGUID>/copy`

tjololo (approved):
:godmode:

PR #225, opened 2023-04-03
Removing / for health endpoint filter

acn-sbuad:
## Description
Health requests not filtered out before telemetry is sent to AI due to error in config. 
endpoint is not completed with a `/` at the end

image

we need this result set to be empty. 

## Related Issue(s)
- #226 

## Verification
~~- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)~~
- [x] All tests run green

### DIFF
--- a/src/Altinn.App.Api/Infrastructure/Telemetry/HealthTelemetryFilter.cs
+++ b/src/Altinn.App.Api/Infrastructure/Telemetry/HealthTelemetryFilter.cs
@@ -37,7 +37,7 @@ namespace Altinn.App.Api.Infrastructure.Telemetry
         {
             RequestTelemetry request = item as RequestTelemetry;
 
-            if (request != null && request.Url.ToString().EndsWith("/health/"))
+            if (request != null && request.Url.ToString().EndsWith("/health"))
             {
                 return true;
             }


acn-sbuad:
Closing PR as this might not be the resolution we need. Creating a bug and leaving it up to team apps to solve when they have time :)

acn-sbuad:
Re-opening the PR after checking the helm charts for apps. This should fix it.

tjololo (approved):

src/Altinn.App.Api/Infrastructure/Telemetry/HealthTelemetryFilter.cs
```diff
@@ -37,7 +37,7 @@ private static bool ExcludeItemTelemetry(ITelemetry item)
         {
             RequestTelemetry request = item as RequestTelemetry;
 
-            if (request != null && request.Url.ToString().EndsWith("/health/"))
+            if (request != null && request.Url.ToString().EndsWith("/health"))
```
ivarne, line 40: One might consider looking at the start of the path, instead of the end, so that we can possibly add more health probes in the future (maybe one to run often and one that checks that storage or other services are availible, and correctly configured.) 
```suggestion
            if (request?.Url.AbsolutePath.StartsWith("/health") == true)
```

acn-sbuad: If more health checks are added in the future wouldn't we also want to filter these checks out from the telemetry logs? Leaving as is for now.

ivarne: That's exactly why I suggested to look at the start of the `AbsolutePath`, instead of the end of the `Url.ToString()`. If you look at the start, you ignore a full prefix, if you look at the end, you will (wrongly) match urls like `api/instances?dontLogThis=/health`. Not sure it's a big issue in reality, but I wrote a comment when i noticed the issue.

ivarne: Oh..

I assumed new endpoints would be `/health/storageConnection` or something. It could obviously be `storageConnection/health`, and then it would be "correct" to look at the end.

PR #227, opened 2023-04-13
Reading out and passing on selected language to pdf generator

RonnyB71:
## Description
If the user has selected another language than the one specified in the profile this language was not passed on to the pdf generator. This passes on the selected language if overridden or the profile language. If nothing is set the falls back to nb.

## Related Issue(s)
- [#1073](https://github.com/Altinn/app-frontend-react/issues/1073)

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Pdf/PdfService.cs
+++ b/src/Altinn.App.Core/Internal/Pdf/PdfService.cs
@@ -13,6 +13,8 @@ using Altinn.Platform.Storage.Interface.Models;
 
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
+using Microsoft.IdentityModel.Tokens;
 using Newtonsoft.Json;
 
 namespace Altinn.App.Core.Internal.Pdf;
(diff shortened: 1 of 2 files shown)


src/Altinn.App.Core/Internal/Pdf/PdfService.cs
```diff
@@ -243,6 +268,22 @@ private async Task<string> GetLanguage()
...
+        if (_httpContextAccessor.HttpContext != null)
+        {
+            bool hasQueryLanguage = _httpContextAccessor.HttpContext.Request.Query.TryGetValue("lang", out StringValues queryLanguage);
+            if (hasQueryLanguage)
+            {
+                return queryLanguage.ToString();
+            }
+        }
+
+        return string.Empty;
```
ivarne, line 284: With nullable reference types, you can safely return `null` and the compiler will ensure that the null is handled properly.

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
```diff
@@ -243,6 +268,22 @@ private async Task<string> GetLanguage()
         return language;
     }
 
+    private string GetOverriddenLanguage()
+    {
+        // If the user has overridden the language from his profile by selecting a
+        // language directly in the form, then use the selected language.
+        if (_httpContextAccessor.HttpContext != null)
+        {
+            bool hasQueryLanguage = _httpContextAccessor.HttpContext.Request.Query.TryGetValue("lang", out StringValues queryLanguage);
```
ivarne, line 277: Could we add this to the controller in the `/next` endpoint, so that swagger will pick this up for documentation?

tjololo: I think I'm agreeing with @ivarne here, unless it would confuse api users as to where the language is used?

tjololo (approved):

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
RonnyB71: Good point, both of you. It's an optional and very specific usage of the parameter and nothing to do with the process navigation, but still. I'll try to document it properly so we can have it in swagger without to much confusion.

Issue #156, opened 2023-01-09
Extended attachment validation


ghost:
### Description
We want to be able to access the metadata about the file and the file as a stream, when the file is uploaded.
The information should be sent as soon as the file is uploaded in the app, so that it is possible to send validation feedback to the client. 

**Validation examples**
* Validate that the file's MIME type is what the extension and content-type claims the file to be
* Image resolution is too low based on the files metadata
* PDF is allowed, but not version 1.5 is this might contain executable scripts

### Additional Information
Checking a file's mime type based on it's binary data is based on analyzing the byte stream for known signatures. We should look into using a library for this. One alternative is for .Net is https://github.com/MediatedCommunications/Mime-Detective

File sizes might be large ie. in the 50 MB to 200 MB range. 

### Tasks
- [x] Identify library to sniff for MIME type

### Acceptance Criterias
- [x] Validate the actual MIME type of the file should be a standard solution enabled with configuration
- [x] The solution should allow for metadata parsers for known file types to be added by registering an implementation of a defined interface, if no metadata parsers exists the minimum metadata available should be the MIME type (if identified)
- [x] Metadata should be added to a context object as name value/pairs making it possible to add validations against them eg. the MIME type should be image/png AND the width should be more than 2000 and the height more than 1500.
- [ ] Validation messages should be localized into the appropriate language and support text resources

PR #229, opened 2023-04-17
Support for extended file analysis

RonnyB71:
## Description
This feature adds support for doing stream based file analysis on attachments uploaded to an instance. Changes in this repository is adding interface/api support, while the actual implementations is the Altinn.FileAnalyzers pacakge. First implementation is supporting extended mime type analysis.

## Related Issue(s)
- #156  

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -137,9 +153,33 @@ public class DataController : ControllerBase
...
+                    List<ValidationIssue> validationIssues = new();
+                    if (FileValidationEnabledForDataType(dataTypeFromMetadata))
+                    {
+                        (fileValidationSuccess, validationIssues) = await _fileValidationService.Validate(dataTypeFromMetadata, fileAnalysisResults);
+                    }
+
+                    if (!fileValidationSuccess)
                     {
-                        return errorResponse;
+                        return new BadRequestObjectResult(GetErrorDetails(errors));
```
tjololo: Checking for !fileValidationSuccess but building response from ValidationIssues produced by CompliesWithDataRestrictions ?

src/Altinn.App.Api/Helpers/RequestHandling/DataRestrictionValidation.cs
```diff
@@ -19,77 +18,126 @@ public static class DataRestrictionValidation
         /// </summary>
         /// <param name="request">the original http request</param>
         /// <param name="dataType">datatype the files is beeing uploaded to</param>
-        /// <param name="errorResponse">Null if validation passed, error response if not</param>
         /// <returns>true with errorResponse = null if all is ok, false with errorResponse if not</returns>
-        public static bool CompliesWithDataRestrictions(HttpRequest request, DataType? dataType, out ActionResult? errorResponse)
+        public static (bool Success, List<ValidationIssue> Errors) CompliesWithDataRestrictions(HttpRequest request, DataType? dataType)
```
tjololo, line 22: `<return>` part of xml comment needs a small update as errors is not null, but an empty List on Success

tjololo, line 22: The return type is list, but it seems like we always return after adding an error so the list would always only contain one item. Is there any reason for not returning all errors found not just the first found?

src/Altinn.App.Core/Extensions/HttpContextExtensions.cs
tjololo, line 1: 🤩

src/Altinn.App.Core/Features/FileAnalyzis/FileAnalysisResult.cs
```diff
@@ -0,0 +1,44 @@
...
+        public string AnalyserId { get; internal set; }
+
+        /// <summary>
+        /// The name of the analysed file.
+        /// </summary>
+        public string? Filename { get; set; }
+
+        /// <summary>
+        /// The file extension(s) without the . i.e. pdf | png | docx
+        /// Some files might have multiple extensions.
```
tjololo: Always thought a file only had one extension, how can a file have multiple?

src/Altinn.App.Core/Internal/App/FrontendFeatures.cs
```diff
@@ -10,9 +13,14 @@ public class FrontendFeatures : IFrontendFeatures
...
         /// Default implementation of IFrontendFeatures
         /// </summary>
-        public FrontendFeatures()
+        public FrontendFeatures(IFeatureManager featureManager)
         {
             features.Add("footer", true);
+
+            if (featureManager.IsEnabledAsync(FeatureFlags.JsonObjectInDataResponse).Result)
+            {
+                features.Add("json_object_in_data_response", true);
```
tjololo: Had a chat with the frontend devs when I added the frontend feature for process actions. And we decided on camelCase. We should choose one type of casing. Any reason for not using camelCase?.

This could also be changed so that we always return the feature, but set the enabled value to false, thoughts?

tjololo (commented):
As we discussed the file analyser test doubles should be made simpler as we test the actual analysers in the library providing them.
Nice to see the integrationtests 🤩  at the same time we are getting another set of large mocks, we should create a issue for improving this so we dont need to have all these mocks (unit integration tests, localtest and others)

src/Altinn.App.Core/Features/FileAnalyzis/FileAnalysisResult.cs
RonnyB71: Unclear comment, but a mime-type can be associated with multiple extensions like .jpeg and .jpg

src/Altinn.App.Core/Internal/App/FrontendFeatures.cs
RonnyB71: Nope, think I did this because I thought I saw you did it somewhere else and that this was what you agreed upon. I like camelCase :-D 

Good point - I'll change.

src/Altinn.App.Core/Features/FileAnalyzis/FileAnalysisResult.cs
RonnyB71: I'll clarify the comment.

src/Altinn.App.Api/Controllers/DataController.cs
RonnyB71: Nice catch, copy/paste error - will fix.

src/Altinn.App.Api/Helpers/RequestHandling/DataRestrictionValidation.cs
RonnyB71: Mostly to be compatible with how it was previously when validating data restrictions since we then returned immediately with one error. It was changed to list to support multiple errors as part of the new file analysis feature. Suggest we keep it like this in this round, an change the behavior when we change the result to be of type ProblemDetails/ValidationProblem.

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -137,9 +153,34 @@
...
+                    if (!validationRestrictionSuccess)
+                    {
+                        return new BadRequestObjectResult(GetErrorDetails(errors));
+                    }
+
+                    IEnumerable<FileAnalysisResult> fileAnalysisResults = new List<FileAnalysisResult>();
+                    if (FileAnalysisEnabledForDataType(dataTypeFromMetadata))
+                    {
+                        StreamContent streamContent = Request.CreateContentStream();
+                        Stream fileStream = new MemoryStream();
```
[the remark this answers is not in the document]
tjololo: Possible to dispose this?

tjololo (approved):
Have a look at the warning from codeQL otherwise this looks good to me

PR #231, opened 2023-04-18
Introduce new method in IData for updating BinaryData (#228)

### LINKED ISSUE #228: Add methods in IData interface that does not take HttpRequest as input
### Description

IData.UpdateBinaryData takes HttpRequest with the binary data as input, we should remove the need for passing the HttpRequest down from controllers and only require the data we actually need to store the data.

The data we currently extract from the HttpRequest today seems to be filename and content type.

This would make the controller responsible for extracting the parts from the httprequest and as a side effect make it much easier to use the IData service to insert and update data in storage.

Old methods requiring httprequest should me marked as deprecated and remove in an later major release

### Additional Information

* Create new methods where only methods with HttpRequest as input paramas exists
* Deprecate and move away from using the old methods

tjololo:
Introduce new method in IData for updating BinaryData without a HttpRequest as input. Old method marked as obsolete

## Related Issue(s)
Fixes #228 

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
```diff
@@ -350,6 +351,30 @@ public async Task<DataElement> UpdateBinaryData(string org, string app, int inst
             _logger.LogError($"Updating attachment {dataGuid} for instance {instanceGuid} failed with status code {response.StatusCode}");
             throw await PlatformHttpException.CreateAsync(response);
         }
+        
+        /// <inheritdoc />
+        public async Task<DataElement> UpdateBinaryData(InstanceIdentifier instanceIdentifier, string? contentType, string filename, Guid dataGuid, Stream stream)
+        {
+            string apiUrl = $"{_platformSettings.ApiStorageEndpoint}instances/{instanceIdentifier}/data/{dataGuid}";
+            string token = JwtTokenUtil.GetTokenFromContext(_httpContextAccessor.HttpContext, _settings.RuntimeCookieName);
```
RonnyB71: Consider using IUserTokenProvider instead to avoid those HttpContext dependencies.

RonnyB71 (approved):
🕹️

PR #233, opened 2023-04-26
Feature/10108 remove shadow fields

nkylstad:
## Description

Add support for removing shadow fields `OnEndProcessTask`. 

- Uses a configurable prefix on the data model fields to determine which fields to remove. 
- Allows for configuration of whether data element should be updated directly or if a copy (where shadow fields are removed) should be created.

## Related Issue(s)
- https://github.com/Altinn/altinn-studio/issues/10108

## Related PRs
_The code will not work until this PR is merged and updated nuget published_
- https://github.com/Altinn/altinn-storage/pull/194

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Core/Implementation/DefaultTaskEvents.cs
```diff
@@ -256,6 +268,44 @@ private async Task RemoveHiddenData(Instance instance, Guid instanceGuid, List<D
...
+                        Modifiers = { modifier.ModifyPrefixInfo }
+                    },
+                    DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+                };
+                
+                string serializedData = JsonSerializer.Serialize(data, options);
+                if (dataType.AppLogic.ShadowFields.SaveToDataType != null) {
+                    var saveToDataType = dataTypesToLock.Find(dt => dt.Id == dataType.AppLogic.ShadowFields.SaveToDataType);
+                    Type saveToModelType = _appModel.GetModelType(saveToDataType.AppLogic.ClassRef);
+                    var updatedData = Newtonsoft.Json.JsonConvert.DeserializeObject(serializedData, saveToModelType ?? modelType);
```
tjololo: Is there a reason for using Newtonsoft instead of System.Text.Json here?

src/Altinn.App.Core/Helpers/ShadowFieldsConverter.cs
tjololo, line 1: Missing some /// docs comments here

tjololo (commented):
Looks good, is there any chance of adding some tests?

src/Altinn.App.Core/Implementation/DefaultTaskEvents.cs
nkylstad: Only that it is consistent with our deserialization elsewhere in this project (as far as I could tell). Had to use `System.Text.Json` for the actual removal of fields, as Newtonsoft did not seem to support that kind of operation.

tjololo: We are migrating away from Newtonsoft, but haven't gotten around to removing it from the entire project. New code is, unless there are good reasons, leveraging System.Text.Json and when we touch code using Newtonsoft we migrate it to System.Text.Json.

tjololo (approved):

PR #236, opened 2023-05-05
Improve handling of non existing source instance

SandGrainOne:
## Description
Turns out that Storage returns 403 if the given instance id results in a 404 from Cosmos DB. I've added slightly better handling of this specific case in the new Copy instance logic, but I'm not confident it's the best solution.

## Related Issue(s)
- https://github.com/Altinn/altinn-platform/issues/355

## Verification
- [x] All tests run green

## Documentation


tjololo (approved):
:godmode:

PR #238, opened 2023-05-10
Add support for "source" parameter on option based components

### LINKED ISSUE #234: PageComponentConverter: `source` is also valid as options
### Description of the bug

If you create a dropdown (or other component requiring options), you can pick between `options` (static list of options) `optionsId` (fetch a list of options from backend, may be a dynamic list, and may use `mapping`) or `source` (using a repeating group as the source of options).

The last one does not seem to be implemented in `PageComponentConverter`, See code permalink here: https://github.com/Altinn/app-lib-dotnet/blob/e93206c2adffcb466a255a4e54731eed1f1fefca/src/Altinn.App.Core/Models/Layout/PageComponentComverter.cs#L293-L307

### Steps To Reproduce

- https://altinn.slack.com/archives/C02EENC8KBR/p1682593423181709

### Additional Information

_No response_

ivarne:
Not sure if these validations of the json content to ensure that nullability annotations in C# are correct really makes sense to have, now that https://github.com/Altinn/app-lib-dotnet/pull/98 was abandoned, but I kind of like them.

## Related Issue(s)
- Fixes https://github.com/Altinn/app-lib-dotnet/issues/234

## Verification
- [x] All tests run green

## Documentation
- [ ] ~~User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)~~

### DIFF
--- a/src/Altinn.App.Core/Models/Layout/Components/OptionsComponent.cs
+++ b/src/Altinn.App.Core/Models/Layout/Components/OptionsComponent.cs
@@ -22,6 +22,11 @@ public class OptionsComponent : BaseComponent
     public List<AppOption>? Options { get; }
 
     /// <summary>
+    /// Alternaltive to <see cref="OptionId" /> where the options are sourced from a repeating group in the datamodel
+    /// </summary>
+    public OptionsSource? OptionsSource { get; }
+
+    /// <summary>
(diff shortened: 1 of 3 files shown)


tjololo (approved):
Thanks! Looks good to me :)

PR #239, opened 2023-05-12
Draft for server side validation of option values

ivarne:
It would be good to have server side validation of options the same way we do server side validation of required.

Unfortunately it was harder than expected, so I'll just leave it here if anyone want to have a starting point.


ivarne:
Were not keeping PR open as a starting point

Issue #241, opened 2023-05-23
Validate endpoint returns 500 if call to platform results in 403


tjololo:
### Description of the bug

The ValidationController does not check for PlatformException from InstanceController, if a user is not authorized to write/confirm platform returns 403, but as no catch of PlatformException is done in the controller it results in a 500 respons from the app.

The controller method should catch platformexceptions and return 403 if the error from Platform was 403. All other cases should be 500 at the moment to avoid exposing internal state unintended.

### Steps To Reproduce

See isse: https://github.com/Altinn/app-frontend-react/issues/426 for steps to reproduce

### Additional Information

Frontend part of the issue: https://github.com/Altinn/app-frontend-react/issues/426

PR #243, opened 2023-05-24
Fix unauthorized submit resulting in 500

Magnusrm:
Handle exception thrown when platform responds with 403 when trying to update process. 
## Description
There has been added a try-catch that catches a PlatformHTTPException and responds with 403 if the exceptions statuscode is 403. If the exception contains any other statuscode, we rethrow the exception.

## Related Issue(s)
- #241 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/ValidateController.cs
+++ b/src/Altinn.App.Api/Controllers/ValidateController.cs
@@ -61,7 +61,7 @@ namespace Altinn.App.Api.Controllers
                 throw new ValidationException("Unable to validate instance without a started process.");
             }
 
-            List<ValidationIssue> messages = new List<ValidationIssue>();
+            List<ValidationIssue> messages;
 
             try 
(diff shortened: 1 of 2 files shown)


src/Altinn.App.Api/Controllers/ValidateController.cs
tjololo, line 1: Instead of creating `List<ValidationIssue> messages;` before the try/catch block and `return Ok(messages);` after. We can just create and return inside the try block.
This does not change the actual logic, but I think it makes the code more readable

tjololo (approved):
Looks good, one comment about code style

src/Altinn.App.Api/Controllers/ValidateController.cs
Magnusrm: I agree with the suggestion! 👍

Issue #244, opened 2023-05-24
New PDF doesn't display sender, even if set to true in InstanceInformation element


violaguttorm:
### Description of the bug

When generating PDF with the new way, it doesn't display the sender. If I put `?pdf=preview` while editing the form, it displays correctly with sender when I press Ctrl+P in Chrome, but when I actually generate the PDF, it doesn't show up. The image below shows the difference (the lower one is how the preview looks, the higher one is what actually comes out when the PDF is generated. (The below one is a screenshot from before sending in).

```
      {
        "id": "pdf-instance",
        "type": "InstanceInformation",
        "elements": {
          "dateSent": true,
          "sender": true,
          "receiver": true,
          "referenceNumber": true
        }
      },
```

The app is https://altinn.studio/repos/brg/rrh-innrapportering 

### Steps To Reproduce

1. Fill in a form in Chrome. 
2.  Add ?pdf=preview in URL
3. Ctrl+P to observe the preview is correct
4. Send in form
5. View PDF to observe the generated PDF lacks sender

### Additional Information

![image](https://github.com/Altinn/altinn-studio/assets/57175682/3a582a95-fe4f-4d32-9238-04946b2a6285)

PR #245, opened 2023-05-25
Fix: Sender not displayed in PDF when representing an org

bjosttveit:
## Description

The frontend relies on the `AltinnPartyId` cookie to get the party information (org.nr., org name...), since this was not available to the PDF service, it could not display the "Sender" in the PDF when you submitted on behalf of an organization. It would only get its personal party information, which did not match with the instanceOwnerParty. This change sets this cookie in the PDF service request.

Fixing this in app-frontend would be more complicated and would likely have unintended consequences by changing the selected party.

## Related Issue(s)
- #244 

## Verification
- [x] All tests run green

## Documentation


ivarne:
It would be better to fix this by making frontend read `instance.instanceOwner.partyId`, because that would also fix another the issue when you have multiple apps open simultaneously representing different organizations.

RonnyB71 (approved):
🕹️

bjosttveit:
Waiting to merge this as I suspect my fix for https://github.com/Altinn/app-frontend-react/issues/268 will also fix this issue.

bjosttveit:
Fixed by https://github.com/Altinn/app-frontend-react/pull/1204

PR #264, opened 2023-06-22
Feature/262 custom action authorization

### LINKED ISSUE #262: Make it possible to add custom code to authorise user for actions in tasks
### Description

In some cases it is required to have rules beyond what is possible with altinn authorisation.
One example is for applications with multiple signing tasks it should be possible to configure that the signature is unique across signature steps ie. the same user cannot sign off on multiple steps even if he/she has the roles to do so.

This example should be supported by Altinn 3 by configuration only, but we should implement this in such a way that service owners can use the same injection points to define their own rules.

### Additional Information

Should be possible to inject custom code with dependency injection to control what actions are available for a user beyond authorisation/policy file.

#### Options
1. This interface/endpoint should take the current task id, user information and a list of actions and return a filtered list of actions.
2. This interface/endpoint should take the current task, user information and a action and return true/false

### Tasks

- [ ] add injection point
- [ ] make defining and reusing logic for different task/actions easy
- [ ] we use the same injection point as service owner

### Acceptance Criterias

_No response_

tjololo:
## Description
Add injection point for developers to create custom authorization logic for actions in tasks

## Related Issue(s)
Fixes: #262 

## Verification
- [ ] All tests run green

## Documentation


RonnyB71 (approved):
🕹️

PR #296, opened 2023-08-25
Rewrite of MimeTypeMap to support extensions with multiple mimetypes

### LINKED ISSUE #295: Support multiple mimetypes for fileextensions
### Description of the bug

Some fileextensions can be represented by multiple mimetypes for example `.zip` can be `application/zip`, `application/x-zip-compressed` or `application/zip-compressed`. The two later mimetypes are deprecated, but som browser and os combinations seems to still result in Content-Type to be set to `application/x-zip-compressed`.

When Content-Type is set to `application/x-zip-compressed` for .zip files the `DataRestrictionValidation.CompliesWithDataRestrictions(Request, dataTypeFromMetadata)` fails as it expects `application/zip` only.

### Steps To Reproduce

1. Perform a POST against data?dataType=<dataTypeId> with a zip file and Content-Type set to `application/x-zip-compressed`
2. The request fail with the respons code 400 with a message with missmatch in mimetype and fileextension

### Additional Information

_No response_

tjololo:
## Description
Some file extensions can be represented by multiple mimetypes
Example: zip files can be application/zip, application/zip-compressed or application/x-zip-compressed

There are still some missing mappings between extension and mimetypes, but I propose to fix them in a separate PR to make the changes more visible while reviewing

## Related Issue(s)
Fixes #295 

## Verification
- [ ] All tests run green

## Documentation


tjololo:
I removed the mimetype -> extension method as it was unused in our code and couldn't find any usage of it in altinn.studio. Either way it would potentially make the wrong assumption as .xsd and .xml both have mimetype text/xml for example

RonnyB71 (approved):
🕹️

PR #299, opened 2023-08-30
#294 Check if stream length is greater than 0 to avoid processing zero byte streams

### LINKED ISSUE #294: Create data in DataController should verify that stream is not empty
### Description of the bug

(Issue 253)[https://github.com/Altinn/altinn-storage/issues/253] was implemented in storage to reduce the chance of storing attachments with zero bytes. We should implement a similar check if possible in app to stop requests we know will fail as early as possible.

### Steps To Reproduce

1. Start a new instance of an app with attachments
2. Upload an attachment of 0 bytes
3. The request will fail with 400

If you view the trace for the request it will call all the way down to storage before erroring. We should stop the request in the application to reduce the load downstream.

### Additional Information

_No response_

tjololo:
## Description
Altinn-storage checks that files are not zero bytes. If the application receives a stream of zero bytes we know it will fail and should return an error before further processing is done

## Related Issue(s)
Fixes #294 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/DataController.cs
+++ b/src/Altinn.App.Api/Controllers/DataController.cs
@@ -165,13 +165,14 @@ namespace Altinn.App.Api.Controllers
                     await streamContent.CopyToAsync(fileStream);
                     if (fileStream.Length == 0)
                     {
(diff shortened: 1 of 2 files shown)


RonnyB71 (approved):
🕹️

PR #300, opened 2023-08-31
Fix `PageComponentConverter` and add unit tests

bjosttveit:
## Description

The PageComponentConverter which parses layout files to run expressions for `RemoveHiddenDataPreview` was broken in a few cases.
- It would crash when encountering a repeating group with multiPage (could not recognize IDs with `<index>:<component-id>` in its children)
- It could not handle group children unless the children where directly following the group in the correct order. This is not something we enforce in frontend, and it should work as long as the referenced group children are defined somewhere.
- It had not been updated to support `Grid` component children, and so could not remove hidden data if the parent Grid was hidden.

I added some unit tests to make sure it can parse layout files without crashing as well as creating the correct hierarchy structure.

Note: There is a test for button-group which does not test the hierarchy, this is just to make sure it does not crash. Since you cannot add components with formdata to its children it is not important to represent its children as actual child components.

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/issues/726

## Verification
- [x] All tests run green

## Documentation


olemartinorg:
> It could not handle group children unless the components following were in the correct order. This is not something we enforce in frontend, and **it should work as long as the group is defined before its child components**.

Technically, we don't enforce that either in `app-frontend-react`. We iterate over all components first, build a map over them, and then process the hierarchy - so it's entirely possible (we even have tests for it) to define the Group component last.

Over time I want to possible to also:
- Be possible for two Group-components to adopt instances of the same child components
- Be possible to adopt child components defined on other pages
- Be possible to define data model bindings as relative to the Group they're inside of

Implementing these things will be quite the chore if we need to maintain two compatible implementations in two different languages, so for the record my opinion remains that it would be better to pre-process all of this in a BFF (backend for frontend) derived from the same code in `app-frontend-react`.

Apart from the rant, I approve of fixing these issues mentioned in the PR! 🙌

bjosttveit:
> Technically, we don't enforce that either in `app-frontend-react`. We iterate over all components first, build a map over them, and then process the hierarchy - so it's entirely possible (we even have tests for it) to define the Group component last.

Ait, I must have missed that part then.

> Implementing these things will be quite the chore if we need to maintain two compatible implementations in two different languages, so for the record my opinion remains that it would be better to pre-process all of this in a BFF (backend for frontend) derived from the same code in app-frontend-react.

I very much support processing this in only one place!

bjosttveit:
I don't think it will require much more work to make it accept that a group is defined after its children, so I will try to make that work as well.

bjosttveit:
@olemartinorg defining the group after its children should be possible now, added tests for this case as well 🤝

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
```diff
@@ -0,0 +1,80 @@
...
+                continue;
+            }
+
+            foreach (var cell in row.Cells)
+            {
+                if (cell.ComponentId is not null)
+                {
+                    gridChildren.Add(cell.ComponentId);
+                }
+            }
```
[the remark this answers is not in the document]
tjololo: What's the reason for not using .Where in this case?

src/Altinn.App.Core/Models/Layout/PageComponentConverter.cs
```diff
@@ -129,38 +132,87 @@ private PageComponent ReadData(ref Utf8JsonReader reader, string pageName, JsonS
             }
         }
 
-        return new PageComponent(pageName, components, componentLookup, hidden, required, readOnly, additionalProperties);
+        if (componentListFlat is null || componentLookup is null || childToGroupMapping is null)
```
tjololo, line 135: I see sonar is marking this as a bug and failing the build, but I don't agree with sonar.
I'm going to mark it as a false positiv in sonar if you don't have any objection.

src/Altinn.App.Core/Models/Layout/Components/PageComponent.cs
```diff
@@ -25,4 +23,12 @@ public class PageComponent : GroupComponent
     /// Helper dictionary to find components without traversing childern.
     /// </summary>
     public Dictionary<string, BaseComponent> ComponentLookup { get; }
+
+    /// <summary>
+    /// AddChild is not needed for PageComponent
+    /// </summary>
+    public override void AddChild(BaseComponent child)
```
RonnyB71, line 30: Not changed in this PR, but is it correct for the PageComponent to inherit from the GroupComponent in the first place? Or should it inherit directly from BaseComponent?

bjosttveit: I suppose it makes the logic simpler when you can just check if it `is GroupComponent` whenever you need to access children

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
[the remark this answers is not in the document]
bjosttveit: Nothing special, what would that look like?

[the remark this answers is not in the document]
tjololo:
```suggestion
            foreach (var cell in row.Cells.Where(c => c.ComponentId is not null))
            {
                gridChildren.Add(cell.ComponentId!);
            }
```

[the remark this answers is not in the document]
bjosttveit: Changed to use .Where now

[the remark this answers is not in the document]
tjololo: All out LINQ-expression I see :)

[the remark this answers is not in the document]
bjosttveit: Oh that was what it meant by it 😅  I thought it just wanted to avoid for loops at all cost 😅

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <summary>
+    /// Constructor for RepeatingGroupComponent
+    /// </summary>
+    public GridComponent(string id, string type, IReadOnlyDictionary<string, string>? dataModelBindings, IEnumerable<BaseComponent> children, IEnumerable<string>? childIDs, Expression? hidden, Expression? required, Expression? readOnly, IReadOnlyDictionary<string, string>? additionalProperties) :
+        base(id, type, dataModelBindings, children, childIDs, hidden, required, readOnly, additionalProperties)
+    { }
+}
+
+
+/// <inheritdoc/>
```
tjololo: Fix documentation for this class. Class and method is using `/// <inheritdoc/>`, but the class does not extend any class so there is nothin to inherit.
Class is fairly simple so just short text are needed I think :)

src/Altinn.App.Core/Models/Layout/Components/PageComponent.cs
tjololo: Do we want a accidental call to AddChild method on a PageComponent to result in a exception or should the method just not do anything?

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+
+/// <summary>
+/// Class for parsing a Grid component's rows and cells and extracting the child component IDs
+/// </summary>
+public class GridConfig
+{
+
+    /// <inheritdoc/>
```
tjololo:
```suggestion
    /// <summary>
    /// Reads the Grid component's rows and returns the child component IDs
    /// </summary>
    /// <param name="reader"></param>
    /// <param name="options"></param>
    /// <returns></returns>
```

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
```diff
@@ -0,0 +1,66 @@
...
+    public static List<string> ReadGridChildren(ref Utf8JsonReader reader, JsonSerializerOptions options)
+    {
+        var rows = JsonSerializer.Deserialize<GridRow[]>(ref reader, options);
+        var gridConfig = new GridConfig { Rows = rows };
+        return gridConfig.Children();
+    }
+
+    /// <inheritdoc/>
```
tjololo:
```suggestion
    /// <summary>
    /// Rows in the grid
    /// </summary>
```

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
```diff
@@ -0,0 +1,66 @@
...
+    [JsonPropertyName("rows")]
+    public GridRow[]? Rows { get; set; }
+
+    /// <inheritdoc/>
```
tjololo:
```suggestion
    /// <summary>
    /// Defines a row in a grid
    /// </summary>
```

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
```diff
@@ -0,0 +1,66 @@
...
+    public class GridRow
+    {
+        /// <inheritdoc/>
```
tjololo:
```suggestion
        /// <summary>
        /// Cells in the row
        /// </summary>
```

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
```diff
@@ -0,0 +1,66 @@
...
+        [JsonPropertyName("cells")]
+        public GridCell[]? Cells { get; set; }
+
+        /// <inheritdoc/>
```
tjololo:
```suggestion
        /// <summary>
        /// Defines a cell in a grid
        /// </summary>
```

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
```diff
@@ -0,0 +1,66 @@
...
+        public class GridCell
+        {
+            /// <inheritdoc/>
```
tjololo:
```suggestion
            /// <summary>
            /// The component ID of the cell
            /// </summary>
```

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
```diff
@@ -0,0 +1,66 @@
...
+            [JsonPropertyName("component")]
+            public string? ComponentId { get; set; }
+        }
+    }
+
+    /// <inheritdoc/>
```
tjololo:
```suggestion
    /// <summary>
    /// Returns the child component IDs
    /// </summary>
    /// <returns></returns>
```

tjololo (approved):
Just to get rid of the warnings when working with the file. As a bonus this will improve the documentation presented in most developer tools

PR #305, opened 2023-09-11
Fix remove hidden data to support list binding and hiddenRow

bjosttveit:
## Description

This updates the removeHiddenData functionality to support `hiddenRow` in repeating groups as well as fixing the implementation of `AddIndicesRecursive` to not crash when encountering a `list`-binding which can be found in a file-upload component with multiple files allowed.

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/issues/726

## Verification
- [x] All tests run green

## Documentation


tjololo:
Looks good, this code has a lot of loops and conditions so I think we should stive to have good testcoverage. Should we take a pair programming session and add som tests for hiddenrows?

bjosttveit:
> Looks good, this code has a lot of loops and conditions so I think we should stive to have good testcoverage. Should we take a pair programming session and add som tests for hiddenrows?

Sure lets do it 🤝

tjololo (approved):
:godmode:

RonnyB71 (approved):
🕹️

olemartinorg:
Checking in here to ask some questions... I believe me and @ivarne discussed the `hiddenRows` functionality at one point when thinking about removing hidden data, and I've talked to SSB about it as well. There are multiple use-cases for `hiddenRows`, and for SSB (at least) they said they would not expect hidden rows to cause deletions from the data model upon submitting. That's because they would like to use `hiddenRows` to filter the (prefill) data in ways to show only the relevant parts for the user - but at the same time they had systems on the receiving end that required all the data to be present in the data model even if some of it was hidden for the user.

Another use-case I've heard of is having two repeating groups on two different pages, bound to the same structure in the data model, but showing different sets of rows (typically using hiddenRow with inverse logic on one of them). We should make sure _no data is deleted_ in those cases, as all data is shown to the user.

So, are we sure this is what users want? Maybe we should have flags for it?

tjololo:
Good queations, have had some of the same "conserns" not just about `hiddenRows`, but also `hidden` I can imagine we would have cases where some hidden fields should be removed and some not... Would it be possible to have a function in the expression language the can be used to determin if the execution is backend or frontend and by leveraging that differ the evaluation backend/frontend if the developer chooses to do so.

Thats not excluding the possibility to have separate flags for RemoveHidden and RemoveHiddenRows

ivarne:
If I remember correctly the removal is implemented with aggregation of two lists, one of the keys to remove and one of visible fields. Before removing anything I check that the relevant fields are not visible anywhere else. HiddenRows should obviously follow the same pattern.

Generally I think it makes most sense to remove all data not visible to the end user when submitting, but the current implementation preserves fields that isn’t mapped in any component (as a way to implement hidden fields that probably will be ignored or manipulated by machine integrations and cause bugs in validation or backend systems).

If there is a use case for not removing fields that are actively hidden by a component, I think the most sensible approach would be a `"hiddenNotRemove": true` just under the hidden expression. Modifying the expression just seems way too syntax heavy.

tjololo:
For now I suggest we just split the removehiddendata into to flags (so we don't make upgrade of apps relying on hiddenRows not being removed impossible). If we down the road see that we need hiddenNotRemoved functionality i like @ivarne approach

bjosttveit:
The changes here makes hiddenRow work in the same way as hidden, so the row gets deleted if it is hidden unless it is visible somewhere else. So if you have multiple groups bound to the same data where different rows are visible in different groups, any row which is visible in any group will be preserved so that use case is not affected.

However, if you add prefill rows which is never shown to the user that would indeed get deleted now, which I suppose is how this functionality is expected to work. I generally think it is kind of weird for the user to submit data they have no knowledge of. I agree that to support this we should rather have a separate configuration to explicitly say that we should not delete this data even if it is hidden, but maybe we should wait for v8 to add this since this right now is a breaking change if it is used like this today already.

I suggest that we change the behavior of this to simply ignore deleting rows on submit? (but still remove rows for validation, as the user cannot be expected to fix errors in hidden data) Would that be acceptable @olemartinorg ?

olemartinorg:
Hmm, I agree with the principle that no data should be hidden from the user while also requiring them to submit it. It goes against the whole point of having a receipt that shows the user what they submitted, and we've discussed before that from the users perspective it's the _receipt_ they confirm as being their valid submission and _not_ the data model itself. So I'm all for keeping the functionality as implemented here - but we should make sure it's visible in the v8 release notes and/or docs that the data removal functionality now removes more data than it used to. No service owner should find out by surprise, at least.

And when it comes to `"hiddenNotRemove": true` I would prefer it being named `"removeDataWhenHidden": false` as to avoid a double-negative. Also, configuring it on the component-level might not be optimal, as each component can have one or many bindings to the data model, and we may not want all of them to be kept/removed. The data modelling tools in Studio can probably add some custom configuration for this, but it might not be easy to read metadata like that from C#. In app-frontend we can now easily read metadata from the data model, at least in v4.

tjololo:
Trying a small summary:
We should revert the changes in this PR that removes hiddenRows from the datamodel if RemoveHiddenDataPreview is set to true to keep the functionality unchanged for v7.

For v8 we need to make a clear decision on what will happen if data is hidden for a user. 
When the developer should use shadowfields and where they should use hidden on the component. Introducing yet another hidden/remove param seems like a bad idea.

Issue #287, opened 2023-08-16
Add `Source` to `ValidationIssue`


bjosttveit:
### Description

Add sources for the different kind of validations so that frontend can filter out the types of validations it does on its own to avoid duplicate validations with different text.

tjololo:
Before upgrading to this version of app-lib:


![Image](https://github.com/Altinn/app-lib-dotnet/assets/1145298/898dc4a2-5610-4b3c-b12f-752de837cd25)

After upgrading to this version of app-lib:


![Image](https://github.com/Altinn/app-lib-dotnet/assets/1145298/c74bf3da-1db2-45da-81ec-b07495a1d907)

PR #314, opened 2023-09-25
Fix: Use different source for custom validation

bjosttveit:
It turns out that custom validation modifies the same ModelStateDictionary as the data model validation, meaning that custom validation would get mapped with the same `Source` as data model validation and would be filtered out in the frontend. This PR keeps the validations separate and gives them different `Source`-fields.

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/287
- https://altinndevops.slack.com/archives/C045EB3JA9X/p1695629729345499

## Verification
- [ ] All tests run green

## Documentation


tjololo (approved):
:godmode:

PR #318, opened 2023-09-28
Remove static feature flags

bjosttveit:
## Description

The `footer` and `processActions` feature flags are static and only exist to make app-frontend backwards compatible with older versions of app-lib. app-frontend v4 will assume that these features are available and will no longer check for these flags since: https://github.com/Altinn/app-frontend-react/pull/1450 and https://github.com/Altinn/app-frontend-react/pull/1516.

Assuming that v8 app-lib will require v4 app-frontend these flags can be removed.

### DIFF
--- a/src/Altinn.App.Core/Internal/App/FrontendFeatures.cs
+++ b/src/Altinn.App.Core/Internal/App/FrontendFeatures.cs
@@ -15,9 +15,6 @@ namespace Altinn.App.Core.Internal.App
         /// </summary>
         public FrontendFeatures(IFeatureManager featureManager)
         {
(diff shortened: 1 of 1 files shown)


tjololo:
We should hold this PR back until we decide if it is necessary to make v4 of frontend mandatory for using v8 of app-lib

bjosttveit:
> We should hold this PR back until we decide if it is necessary to make v4 of frontend mandatory for using v8 of app-lib

Yes

tjololo:
As we plan to support v3 of frontend with v8 I close this PR

PR #323, opened 2023-10-13
#322 Deserializing applicationmetadata keeps unmapped properties

### LINKED ISSUE #322: ApplicationMetadata in app-lib should allow unmapped properties on root level
### Description

ApplicationMetadata holds configuration for frontend, backend and storage. To add frontend only configuration we need to update app-lib and force developers to update the nuget to expose the new configurations to frontend.
As a first step in making this json config easier to extend with configurations that only is necessary for frontend backend should not remove fields in the root object that is not explicitly mapped in the strongly typed C# model

### Additional Information

_No response_

tjololo:
## Description
Fixes #322 

## Related Issue(s)
- #322

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/ApplicationMetadataController.cs
+++ b/src/Altinn.App.Api/Controllers/ApplicationMetadataController.cs
@@ -30,7 +30,7 @@ namespace Altinn.App.Api.Controllers
         }
 
         /// <summary>
-        /// Get the application metadata
+        /// Get the application metadata https://altinncdn.no/schemas/json/application/application-metadata.schema.v1.json
         ///
         /// If org and app does not match, this returns a 409 Conflict response
         /// </summary>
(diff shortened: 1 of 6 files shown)


ivarne:
If we need to support newtonsoft json, there should probably be a test for it.

Also, the important thing to test here is the round trip deserialization - serialization. I don’t think we actually care much about accessing unknown properties through the dictionary.

tjololo:
> If we need to support newtonsoft json, there should probably be a test for it.
> 
> Also, the important thing to test here is the round trip deserialization - serialization. I don’t think we actually care much about accessing unknown properties through the dictionary.

While trying to create tests for newtonsoft I realised that supporting newtonsoft would result in a different json returned if serializing with system.text and deserializing with newtonsoft (JsonElement vs JToken) as we still plan to remove Newtonsoft in favor of System.Text I don't think it is worth the effort to support both when we are using System.Text.Json to deserialize the file in AppMetadata.cs

ivarne (approved):

PR #328, opened 2023-10-18
Ivarne/warning fixes

ivarne:
A collection of commits that fixes warnings and add tests where behaviour is changed. For review it might be helpful to look at one commit at the time.

## Verification
- [ ] All tests run green

### DIFF
--- a/src/Altinn.App.Api/Controllers/DataController.cs
+++ b/src/Altinn.App.Api/Controllers/DataController.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 using System.Net;
 using System.Security.Claims;
 using Altinn.App.Api.Helpers.RequestHandling;
@@ -113,18 +115,13 @@ namespace Altinn.App.Api.Controllers
             [FromRoute] Guid instanceGuid,
             [FromQuery] string dataType)
         {
-            if (string.IsNullOrWhiteSpace(dataType))
-            {
-                return BadRequest("Element type must be provided.");
-            }
-
             /* The Body of the request is read much later when it has been made sure it is worth it. */
 
             try
             {
                 Application application = await _appMetadata.GetApplicationMetadata();
-                
-                DataType dataTypeFromMetadata = application.DataTypes.FirstOrDefault(e => e.Id.Equals(dataType, StringComparison.InvariantCultureIgnoreCase));
+
+                DataType? dataTypeFromMetadata = application.DataTypes.FirstOrDefault(e => e.Id.Equals(dataType, StringComparison.InvariantCultureIgnoreCase));
 
                 if (dataTypeFromMetadata == null)
                 {
@@ -158,7 +155,7 @@ namespace Altinn.App.Api.Controllers
                     (bool validationRestrictionSuccess, List<ValidationIssue> errors) = DataRestrictionValidation.CompliesWithDataRestrictions(Request, dataTypeFromMetadata);
                     if (!validationRestrictionSuccess)
                     {
-                        return new BadRequestObjectResult(await GetErrorDetails(errors));
+                        return BadRequest(await GetErrorDetails(errors));
                     }
 
                     StreamContent streamContent = Request.CreateContentStream();
@@ -175,11 +172,11 @@ namespace Altinn.App.Api.Controllers
                             Description = errorMessage
                         };
                         _logger.LogError(errorMessage);
-                        return new BadRequestObjectResult(await GetErrorDetails(new List<ValidationIssue> { error }));
+                        return BadRequest(await GetErrorDetails(new List<ValidationIssue> { error }));
                     }
                     
                     bool parseSuccess = Request.Headers.TryGetValue("Content-Disposition", out StringValues headerValues);
-                    string filename = parseSuccess ? DataRestrictionValidation.GetFileNameFromHeader(headerValues) : string.Empty;
+                    string? filename = parseSuccess ? DataRestrictionValidation.GetFileNameFromHeader(headerValues) : null;
 
                     IEnumerable<FileAnalysisResult> fileAnalysisResults = new List<FileAnalysisResult>();
                     if (FileAnalysisEnabledForDataType(dataTypeFromMetadata))
@@ -196,7 +193,7 @@ namespace Altinn.App.Api.Controllers
 
                     if (!fileValidationSuccess)
                     {
-                        return new BadRequestObjectResult(await GetErrorDetails(validationIssues));
+                        return BadRequest(await GetErrorDetails(validationIssues));
                     }
 
                     fileStream.Seek(0, SeekOrigin.Begin);
@@ -258,7 +255,7 @@ namespace Altinn.App.Api.Controllers
                     return NotFound($"Did not find instance {instance}");
                 }
 
-                DataElement dataElement = instance.Data.FirstOrDefault(m => m.Id.Equals(dataGuid.ToString()));
+                DataElement? dataElement = instance.Data.FirstOrDefault(m => m.Id.Equals(dataGuid.ToString()));
 
                 if (dataElement == null)
                 {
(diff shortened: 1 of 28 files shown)


src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -113,18 +115,13 @@ public class DataController : ControllerBase
             [FromRoute] Guid instanceGuid,
             [FromQuery] string dataType)
         {
-            if (string.IsNullOrWhiteSpace(dataType))
-            {
-                return BadRequest("Element type must be provided.");
-            }
-
```
tjololo, line 120: This will change the error response to the user as not providing dataType param will lead to BadReuqest from line 128 `Element type {dataType} not allowed for instance {instanceOwnerPartyId}/{instanceGuid}.`

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -421,21 +418,19 @@ private ActionResult ExceptionResponse(Exception exception, string message)
...
             {
-                ServiceException se = exception as ServiceException;
                 return StatusCode((int)se.StatusCode, se.Message);
             }
 
             return StatusCode(500, $"{message}");
         }
 
-        private async Task<ActionResult> CreateBinaryData(Instance instanceBefore, string dataType, string contentType, string filename, Stream fileStream)
+        private async Task<ActionResult> CreateBinaryData(Instance instanceBefore, string dataType, string contentType, string? filename, Stream fileStream)
```
tjololo, line 434: Checked the call chain for this method in regards to filename beeing null. Seems to me that DataRestrictionValidation.CompliesWithDataRestrictions that is called from the Post Data method should ensure that the filename is not null, but it's not quite obvious so the compiler might not catch up on it

src/Altinn.App.Api/Controllers/OptionsController.cs
```diff
@@ -38,10 +40,10 @@ public OptionsController(IAppOptionsService appOptionsService)
         [HttpGet("{optionsId}")]
         public async Task<IActionResult> Get(
             [FromRoute] string optionsId,
-            [FromQuery] string language,
+            [FromQuery] string? language,
```
tjololo, line 43: Could we set "nb" as default param value instead. Makes the fact that nb is the default more obvious both in the code and swagger

src/Altinn.App.Api/Controllers/OptionsController.cs
```diff
@@ -74,12 +76,12 @@ public OptionsController(IAppOptionsService appOptionsService)
             [FromRoute] int instanceOwnerPartyId,
             [FromRoute] Guid instanceGuid,
             [FromRoute] string optionsId,
-            [FromQuery] string language,
+            [FromQuery] string? language,
```
tjololo, line 79: Could we set "nb" as default param value instead. Makes the fact that nb is the default more obvious both in the code and swagger

src/Altinn.App.Api/Controllers/ProcessController.cs
```diff
@@ -257,7 +257,7 @@
                 
                 Instance instance = await _instanceClient.GetInstance(app, org, instanceOwnerPartyId, instanceGuid);
 
-                if (instance.Process == null)
+                if (instance?.Process == null)
```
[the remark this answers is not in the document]
tjololo: Not sure if I agree with this, yes what instance is fetched is controlled by the user, but we control if the process is set or not. Do you disagree @ivarne ?

src/Altinn.App.Api/Controllers/ProcessController.cs
```diff
@@ -413,11 +418,11 @@
 
                 counter++;
             }
-            while (instance?.Process?.EndEvent == null || counter > MaxIterationsAllowed);
+            while (instance.Process.EndEvent == null || counter > MaxIterationsAllowed);
```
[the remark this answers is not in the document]
tjololo: Seems like a false positive as well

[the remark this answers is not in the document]
tjololo: Seems like a false positive as well

tjololo (approved):
Nice work! Some comments to evaluate

src/Altinn.App.Api/Controllers/DataController.cs
ivarne: No, the addition `#nullable enable` on line 1 will trigger asp net to return `BadRequest` with a `Problem` json structure that explains that `dataType` is a required part of the query string, because it isn't nullable.

src/Altinn.App.Api/Controllers/DataController.cs
ivarne: Nice, I only traced it to being read from a header I think is optional, so I figured it might be null. We should probably store the file with a filename anyway, so a better solution to the nullability problem might be to throw or use `Guid.NewGuid()`

src/Altinn.App.Api/Controllers/OptionsController.cs
ivarne: That was my first attempt, but in C# you can't put a default value in the middle of a parameter list. `[FromQuery] Dictionary` seems a little bit odd to not have last. I'll try to see if `queryParams = null!` works.

src/Altinn.App.Api/Controllers/ProcessController.cs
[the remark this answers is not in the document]
ivarne: It looks very much like false positives to me. Maybe an indication that our code isn't as clear as it should be.

src/Altinn.App.Api/Controllers/DataController.cs
tjololo: Have you tested this? Might be som error in my test setup, but when I tested it locally it didn't return bad request with Problem json when I omitted the dataType param

src/Altinn.App.Api/Controllers/ProcessController.cs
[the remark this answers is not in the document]
tjololo: For now lets just mark the codeQL check as a false positiv, but I agree that this is an indication that our code isn't as clear as ut should be

src/Altinn.App.Api/Controllers/OptionsController.cs
tjololo: I see, didn't think of that during my review. Don't but to much effort into fixing my comment!

src/Altinn.App.Api/Controllers/DataController.cs
ivarne: Hmm... Not here specifically. I got that behaviour (breaking a test) when running tests under net8.0 with `<nullable>enabled</nullable>` in project.json. Therefore I was very careful with [FromQuery] parameters that was not nullable. Maybe it's only when nullability is enabled globally, maybe only net8.0?

ivarne: You were right. The difference was that I tested this with `OptionsController` and it was annotated with `[ApiController]` which causes automatic BadRequest for non nullable query strings when nullable is enabled. I wrote a test for this "BadRequest" response and added the annotation to make the test work in ab43649f.

There is no difference between dovnet 6-8 or file scoped/project scoped nullable

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -332,19 +330,19 @@
 
                 if (appLogic == null)
                 {
-                    _logger.LogError($"Could not determine if {dataType} requires app logic for application {org}/{app}");
+                    _logger.LogError("Could not determine if {dataType} requires app logic for application {org}/{app}", dataType, org, app);
```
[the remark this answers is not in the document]
ivarne: log4j: Ooof, we use application insights that shouldn't have any of these issues, but application owners might use other tools. I don't think there is anything we can do preventively here?

ivarne:
@tjololo Ser ut som du approvet før jeg gjorde de siste endringene. Vil du godkjenne igjen, eller skal jeg bare merge?

src/Altinn.App.Api/Controllers/DataController.cs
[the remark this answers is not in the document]
tjololo: Agreed, this change does not make it worse/better and we have alot of log4shell warnings/errors

tjololo (approved):

Issue #316, opened 2023-09-25
Return the full data model to app-frontend after processing a write operation


olemartinorg:
### Description

Instead of diffing the data model for changes and returning those changes to app-frontend, the backend should just return the full data model as-is.

This could also include:
- Removing data not present in the strict model
- Any changes from ProcessDataWrite

And has the advantage of:
- Removing JsonHelper altogether
- Simplifying logic when saving data on the backend, and shifting this responsibility to the frontend instead

### Additional Information

- The frontend part of this issue: https://github.com/Altinn/app-frontend-react/issues/1501
- The origin: https://github.com/Altinn/app-lib-dotnet/pull/116#issuecomment-1361112973
- #315

olemartinorg:
Fixed as of #393

PR #329, opened 2023-10-24
Multipart form data

### LINKED ISSUE #128: Ignore return value from ProcessDataWrite
## Problem
Writing all the if statements to correctly keep track of the changed status of data after calculations is hard, error prone, and time consuming

## Solution
Take a copy of the data before calling ProcessDataWrite, and compare afterwards to get the changed status. This also have the added benefit that we might send just the changes to the frontend, eliminating the risk of user changes being reverted when the calculation is ready.

## Describe alternatives you've considered
Document clearly that users can always return true, and everything will work. Give help on evaluating the trade off and consequences.

### LINKED ISSUE #187: Support multipart saving
### Description

After Altinn/app-frontend-react/pull/899 is merged, multipart saving will be implemented in `app-frontend-react`. This means that the PUT request for saving the data model _can_ start sending multipart bodies containing not just the datamodel we want to save, but also the paths inside that datamodel that changed (along with their previous values).

As an example, here is a request we might send. I've added whitespace to the JSON for readability, although we won't send that in practice.

```http
PUT /org/app/instances/512345/uuid/data/uuid
Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryyWYh3C7EIYlu8pJj

------WebKitFormBoundaryyWYh3C7EIYlu8pJj
Content-Disposition: form-data; name="dataModel"

{
  "skjemanummer": "1603",
  "spesifikasjonsnummer": "12392",
  "blankettnummer": "RF-1366",
  "tittel": "Endringsmelding",
  "gruppeid": "9785",
  "Endringsmelding-grp-9786": {
    "Avgiver-grp-9787": {
      "KontaktpersonEPost-datadef-27688": {
        "value": "Ja"
      }
    },
    "OversiktOverEndringene-grp-9788": [
      {
        "SkattemeldingEndringEtterFristOpprinneligBelop-datadef-37131": {
          "value": "1"
        }
      }
    ]
  }
}
------WebKitFormBoundaryyWYh3C7EIYlu8pJj
Content-Disposition: form-data; name="previousValues"

{
  "Endringsmelding-grp-9786.OversiktOverEndringene-grp-9788[0].SkattemeldingEndringEtterFristOpprinneligBelop-datadef-37131.value": null
}
------WebKitFormBoundaryyWYh3C7EIYlu8pJj--
```

You now get two objects, not just one:
1. The `dataModel` (always named as such, regardless of the data model name - which we don't really know or care about in app-frontend). This is the same object we used to send.
2. The `previousValues` map. This uses dot-notation to indicate a path inside the data model that changed. For every item that changed, the previous value is listed as the map value. If the path did not exist before, the value will be `null`.

The above example contains a path inside a repeating group that changed - specifically on row 0.

Another interesting example is seen here. I added 3 rows, with the following values:
1. 1 and 2
2. 3 and 4
3. 5 and 6

Then I deleted the second row (containing the values 3 and 4), which leads to this request:

```http
PUT /org/app/instances/512345/uuid/data/uuid
Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryBkzRPylfSvVRa1ym

------WebKitFormBoundaryBkzRPylfSvVRa1ym
Content-Disposition: form-data; name="dataModel"

{
  "skjemanummer": "1603",
  "spesifikasjonsnummer": "12392",
  "blankettnummer": "RF-1366",
  "tittel": "Endringsmelding",
  "gruppeid": "9785",
  "Endringsmelding-grp-9786": {
    "Avgiver-grp-9787": {
      "KontaktpersonEPost-datadef-27688": {
        "value": "Ja"
      }
    },
    "OversiktOverEndringene-grp-9788": [
      {
        "SkattemeldingEndringEtterFristOpprinneligBelop-datadef-37131": {
          "value": "1"
        },
        "SkattemeldingEndringEtterFristNyttBelop-datadef-37132": {
          "value": "2"
        }
      },
      {
        "SkattemeldingEndringEtterFristOpprinneligBelop-datadef-37131": {
          "value": "5"
        },
        "SkattemeldingEndringEtterFristNyttBelop-datadef-37132": {
          "value": "6"
        }
      }
    ]
  }
}
------WebKitFormBoundaryBkzRPylfSvVRa1ym
Content-Disposition: form-data; name="previousValues"

{
  "Endringsmelding-grp-9786.OversiktOverEndringene-grp-9788[1].SkattemeldingEndringEtterFristOpprinneligBelop-datadef-37131.value": "3",
  "Endringsmelding-grp-9786.OversiktOverEndringene-grp-9788[1].SkattemeldingEndringEtterFristNyttBelop-datadef-37132.value": "4",
  "Endringsmelding-grp-9786.OversiktOverEndringene-grp-9788[2].SkattemeldingEndringEtterFristOpprinneligBelop-datadef-37131.value": "5",
  "Endringsmelding-grp-9786.OversiktOverEndringene-grp-9788[2].SkattemeldingEndringEtterFristNyttBelop-datadef-37132.value": "6"
}
------WebKitFormBoundaryBkzRPylfSvVRa1ym--
```

Notice that the row indexes 1 and 2 had changed (row 0 did not change). Index 1 used to have the values 3 and 4 (but from the data model we can tell they now have the values 5 and 6), while index 2 used to have the values 5 and 6 (but index 2 no longer exists in the data model).

Other things to be aware of:
- As an app developer, I would expect to have a somewhat rich interface to interact with this information (the raw data is a bit cumbersome to work with). I.e., a way to get a list of "changes" with previous/new values.
- For this feature to work with app-frontend, the backend needs to inform that the `multiPartSave` feature is enabled (see #188).
- The headers we used to send (`X-DataField` and `X-ComponentId`) are no longer passed to the backend when multipart saving is enabled. It is advised that app developers can choose to enable multipart saving when they are ready for it, although we should encourage them to do so (possibly also updating the template to enable the feature by default for new apps).

### Additional Information

- Altinn/app-frontend-react/issues/810
- Altinn/app-frontend-react/pull/899 (client-side implementation)

ivarne:
Work in process

Builds on https://github.com/Altinn/app-lib-dotnet/pull/328, so temporary the target branch is different

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/187
- closes #128

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Api/Controllers/StatelessDataController.cs
```diff
@@ -115,8 +115,12 @@ public class StatelessDataController : ControllerBase
...
             await _prefillService.PrefillDataModel(owner.PartyId, dataType, appModel);
 
-            Instance virutalInstance = new Instance() { InstanceOwner = owner };
-            await _dataProcessor.ProcessDataRead(virutalInstance, null, appModel);
+            Instance virtualInstance = new Instance() { InstanceOwner = owner };
+            foreach (var dataProcessor in _dataProcessors)
+            {
+                _logger.LogInformation("ProcessDataRead for {modelType} using {dataProcesor}", appModel.GetType().Name, dataProcessor.GetType().Name);
+                await dataProcessor.ProcessDataRead(virtualInstance, null, appModel);
+            }
```
tjololo: Could be extracted to a private method as the same lines of code are repeated multiple times.

src/Altinn.App.Core/Features/DataProcessing/NullDataProcessor.cs
tjololo, line 1: 👍

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
```diff
@@ -61,50 +58,74 @@ public ModelDeserializer(ILogger logger, Type modelType)
...
+                return modelResult;
+            }
 
+            MultipartSection? secondSection = await reader.ReadNextSectionAsync();
+            Dictionary<string, string?>? reportedChanges = null;
+            if (secondSection is not null)
+            {
+                if (secondSection?.ContentDisposition?.Contains("name=\"previousValues\"") != true)
+                {
+                    return ModelDeserializerResult.FromError("First entry in mulipart serialization must have name=\"dataModel\"");
```
tjololo: Missleading error returned
```suggestion
                    return ModelDeserializerResult.FromError("Second entry in mulipart serialization must have name=\"previousValues\"");
```

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
```diff
@@ -61,50 +58,74 @@ public ModelDeserializer(ILogger logger, Type modelType)
...
+                }
+                reportedChanges = await System.Text.Json.JsonSerializer.DeserializeAsync<Dictionary<string, string?>>(secondSection.Body);
+                if (await reader.ReadNextSectionAsync() != null)
+                {
+                    return ModelDeserializerResult.FromError("Multipart request had more than 2 elements");
```
tjololo: Should the error be even more specific and say that we expect "dataModel" and optionally a "previousValues" as we know that we only support these?

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
tjololo, line 1: Should we take the effort to remove Newtonsoft from this part of the code now that we have to test this part of the code anyways?
Our goal is still to migrate all code from newtonsoft to system.text

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializerResult.cs
```diff
@@ -0,0 +1,64 @@
...
+    ///     //result.Model is not null here
+    /// }
+    /// else
+    /// {
+    ///     //result.Error is not null here
+    /// }
+    /// </code>
+    /// </example>
+    [MemberNotNullWhen(true, nameof(Error))]
+    [MemberNotNullWhen(false, nameof(Model))]
```
tjololo, line 47: Nice! Did not know about this attribute, this is why I love reviews :)

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
```diff
@@ -171,15 +171,15 @@ public async Task<object> GetFormData(Guid instanceGuid, Type type, string org,
             {
                 using Stream stream = await response.Content.ReadAsStreamAsync();
                 ModelDeserializer deserializer = new ModelDeserializer(_logger, type);
-                object? model = await deserializer.DeserializeAsync(stream, "application/xml");
+                ModelDeserializerResult deserializerResult = await deserializer.DeserializeAsync(stream, "application/xml");
 
-                if (deserializer.Error != null || model is null)
+                if (deserializerResult.Error is not null || deserializerResult.Model is null )
```
tjololo: Could HasError be used instead? That should cover the same checks if I read the code correctly

tjololo (commented):
Nice work! :godmode:

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializerResult.cs
ivarne: It's quite new but really great for readability of the resulting code.

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
ivarne: It probably should. I'll test and update.

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
ivarne: Not sure if that is required. To get here we need the first element to be `dataModel` and the second to be `previousValues`, otherwise the earlier errors will trigger.

src/Altinn.App.Api/Controllers/StatelessDataController.cs
ivarne: Probably not a bad idea. I rejected it because to do it properly I'd need a new service to extract the common parts between stateless and normal.

```c#
internal class DataProcessorWrapper(IEnumerable<IDataProcessor> dataProcessors, ILogger<DataProcessorWrapper> logger)
{
    internal async Task DataProcessRead(...)
    {
        foreach(...)
        {
            // log
            // processRead(...)
        }
    }
    // ProcessDataWrite
}
```

But I did not really like the pattern.

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
ivarne: I think that should be its own issue and PR. It has consequences and we might need it to be a setting in case someone has issues.

ivarne: Should probably look into it with v8

ivarne (commented):
Fikset noen av kommentarene

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
tjololo: thats true!

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
tjololo: :+1: created a issue for this specific case: https://github.com/Altinn/app-lib-dotnet/issues/345

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
```diff
@@ -61,50 +58,74 @@ public ModelDeserializer(ILogger logger, Type modelType)
...
             catch (Exception ex)
             {
-                string message = $"Unexpected exception when attempting to deserialize JSON into '{_modelType}'";
-                _logger.LogError(ex, message);
-                Error = message;
-                return null;
-            }
+                _logger.LogError(ex, "Unexpected exception when attempting to deserialize JSON into '{modelType}'", _modelType);
+                return ModelDeserializerResult.FromError($"Unexpected exception when attempting to deserialize JSON into '{_modelType}'");
+            };
```
tjololo:
```suggestion
            }
```

tjololo (approved):
One last thing I missed, but looks good. Lets get this from review to test 🙁

ivarne:
@tjololo 😡 The tests works on OSX, but not in linux? Any idea how I might debug that issue?

It seems likely to be an issue with unreadable files due to case-sensitivity causing authentication failure, but I did not find anything.

tjololo:
I discovered the same thing yesterday on a branch I have. It's the same issue we have in app-template. I have a linux laptop so plan to spend the day investigating what the issue is

tjololo:
Can we add some test for the bad requests. Like that the first part not having name=dataModel and so on?

olemartinorg:
I tried this against frontend v4/main now, and it's not quite there. Should be the same with v3, as I haven't tested with the form data rewrite in frontend yet.

1. A feature flag, `multiPartSave` is expected to be returned in applicationmetadata for frontend to start sending multipart changes. I added this manually in `FrontendFeatues.cs` when testing.
2. Frontend sends `name="dataModel"`, not `name=dataModel` (note the quotes). I manually added some quotes in `ModelDeserializer` when testing.
3. The whole data model seems to be returned, not the metadata combined with `changedFields`, as is expected in v3. Be sure not to conflate this task with #316, as that's an entirely different beast. Also, returning the entire data model on the top level might not be what we want, as it would be nice to include validation messages etc as well in the future.

olemartinorg:
Tested again now, and it's all working against v4/main as far as I can tell! :raised_hands:

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -58,7 +58,7 @@ public class DataController : ControllerBase
         /// <param name="instanceClient">instance service to store instances</param>
         /// <param name="instantiationProcessor">Instantiation processor</param>
         /// <param name="dataClient">A service with access to data storage.</param>
-        /// <param name="dataProcessor">Serive implemnting logic during data read/write</param>
+        /// <param name="dataProcessors">Serive implemnting logic during data read/write</param>
```
tjololo:
```suggestion
        /// <param name="dataProcessors">Services implementing logic during data read/write</param>
```

tjololo (approved):
Have a look at the code smells/codeql warnings, and try to add the last few percentages of testscoverage and we should be good to go as far as I can see

src/Altinn.App.Api/Controllers/StatelessDataController.cs
```diff
@@ -115,12 +115,24 @@ public class StatelessDataController : ControllerBase
             // runs prefill from repo configuration if config exists
             await _prefillService.PrefillDataModel(owner.PartyId, dataType, appModel);
 
-            Instance virutalInstance = new Instance() { InstanceOwner = owner };
-            await _dataProcessor.ProcessDataRead(virutalInstance, null, appModel);
+            Instance virtualInstance = new Instance() { InstanceOwner = owner };
+            await ProcessAllDataWrite(virtualInstance, appModel);
```
ivarne, line 119: @tjololo This was not good!!!

Happens to be OK, as `ProcessAllDataWrite` actually calls `processDataRead` 😊 ☺️ I'll fix in the next rewrite that uses json patch

PR #331, opened 2023-10-26
Fix more warnings

ivarne:
* ImplicitUsings is enabled with "enabled" not "true"
* Fix nullability warnings
* use FluentAssertions in a test to avoid nullability warning.

## Verification
- [ ] All tests run green

### DIFF
--- a/src/Altinn.App.Api/Helpers/StartupHelper.cs
+++ b/src/Altinn.App.Api/Helpers/StartupHelper.cs
@@ -43,6 +43,6 @@ public static class StartupHelper
     {
         string appMetaDataString = File.ReadAllText("config/applicationmetadata.json");
         JObject appMetadataJObject = JObject.Parse(appMetaDataString);
-        return appMetadataJObject.SelectToken("id").Value<string>();
+        return appMetadataJObject.SelectToken("id")?.Value<string>() ?? throw new Exception("config/applicationmetadata.json does not contain an \"id\" property");
     }
 }
--- a/src/Altinn.App.Api/Altinn.App.Api.csproj
+++ b/src/Altinn.App.Api/Altinn.App.Api.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
     <TargetFrameworks>net6.0</TargetFrameworks>
@@ -13,7 +13,7 @@
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>https://github.com/Altinn/app-lib-dotnet</RepositoryUrl>
     <IsPackable>true</IsPackable>
-    <ImplicitUsings>true</ImplicitUsings>
+    <ImplicitUsings>enable</ImplicitUsings>
 
     <!-- SonarCloud requires a ProjectGuid to separate projects. -->
     <ProjectGuid>{E8F29FE8-6B62-41F1-A08C-2A318DD08BB4}</ProjectGuid>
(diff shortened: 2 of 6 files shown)


src/Altinn.App.Api/Helpers/StartupHelper.cs
```diff
@@ -43,6 +43,6 @@ public static string GetApplicationId()
     {
         string appMetaDataString = File.ReadAllText("config/applicationmetadata.json");
         JObject appMetadataJObject = JObject.Parse(appMetaDataString);
-        return appMetadataJObject.SelectToken("id").Value<string>();
+        return appMetadataJObject.SelectToken("id")?.Value<string>() ?? throw new Exception("config/applicationmetadata.json does not contain an \"id\" property");
```
tjololo, line 46: Instead of throwing a generic exception should we throw [Altinn.App.Core.Internal.App.ApplicationConfigException](https://github.com/Altinn/app-lib-dotnet/blob/main/src/Altinn.App.Core/Internal/App/ApplicationConfigException.cs)?

ivarne: My current opinion is that exceptions should be reserved for unexpected error conditions (to be handled by a catch all exception handler) or to unwind a stack in a recursive structure (eg Json parsing that fails deep in a stack of parsing functions).

Exception types doesn't really add much value in readability/debuggability and trying to implement a hierarchy of exceptions that isn't used for any purpose is a sure way to design a hierarchy that can't be used for any purpose.

The current behaviour is to thow a NullReferenceException without any message about what is wrong, and the key part of the change is to add the explanation in the message.

tjololo: In this case I can agree that throwing a more specific exception would add much in terms of read-/debug-ability, but would throwing a non generic exception have any negativ impact? This is as far as I can see a AppConfigException, by throwing a generic exception we are forcing consumers of the code to catch generic exceptions to catch this.
[As dotnet guidelines states you should not intentionally throw System.Exception from your code](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/exceptions/creating-and-throwing-exceptions#things-to-avoid-when-throwing-exceptions) I think that we should avoid it when there is a obvious alternativ

ivarne: I and others think that guideline is wrong, but if you think it is correct you have a great argument bringing it up. Dotnet has in later versions provided tuple returns as an alternative for error handling to having an `out int errno` or throwing exceptions.

Another point is that it says "**intentionally** from your own source code", which in my mind means when a not totally unexpected condition occurs. This is a totally unexpected condition, as you can't deploy an app without id.

tjololo: I think this is something we should discuss and agree upon as a team. I happy with both ways, but we should be consistent across our codebase IMO

ivarne: I'm also fine both ways, but I think that throwing a normal `Exception` with a helpful message rather than a `NullReferenceException` hits the proper balance here.

Throwing custom exceptions just in case someone wants to catch and continue from a missing app id in applicationmetadata.json seems like optimization for something we're certain that won't happen.

tjololo: Suggestion on a "team standard": if the code throwing the exception is not expected to be used and handled by any developer we can throw generic exception, if there is a natural usecase where the exception could/should be handled by code from other developers we should throw more specific exceptions so that they can be properly handled?

ivarne: I'd write "handled by *catch expression*", because the developer is the human that will read the stack traces from generic exceptions and fix the reason instead of automatically handling the exception. But I think we mean exactly the same.

tjololo (approved):

Issue #315, opened 2023-09-25
Expose the app-lib-dotnet version to app-frontend in applicationmetadata


olemartinorg:
### Description

We have a need for requiring a minimum backend version in app-frontend, so the backend should expose its current version to us in some way.

### Additional Information

- https://github.com/Altinn/app-frontend-react/issues/1496

ivarne:
Fixed by https://github.com/Altinn/app-lib-dotnet/pull/333

PR #333, opened 2023-11-02
New field AltinnNugetVersion in applicationMetadata.cs

ivarne:
Rather simple implementation that makes it possible to override, but if the property isn't set, the version info will be read by from the assembly that contains `ApplicationMetadata`.

Should probably consider adding some error handling, but I don't know when it might trigger, so it's hard to know what to say. 

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/315

## Verification
- [ ] All tests run green

## Documentation


tjololo (approved):

ivarne:
@tjololo This was updated to use a different version that doesn't include prerelease version info.

Previous: `8.0.0-preview.10.3+9c4ed1a127458a23d522f4398a591f5434fae0fd`

Now: `8.0.0.0`

PR #340, opened 2023-11-06
Update app-lib-dotnet to dotnet8 with c# 12

ivarne:
We should ensure that v8 works with the newest dotnet version, which at the time of release of app-lib v8 will be `net8.0`.

If someone finds a compelling idea for why we would need multi-targeting `net6.0` (or `net7.0`), I'm not against it, but forcing users to have similar setup and not have to ask (Oh, you run `net6.0`, we didn't properly test that) is a huge time saver in my opinion.

This PR does not include the fix of the update tool to change dotnet version, as it would conflict with changes in https://github.com/Altinn/app-lib-dotnet/pull/329

## Related Issue(s)
- No separate issue

## Remarks
Something funny happened with FeatureManagment when updating, and I had to use the newest version to make the tests compile. Seems like a very sensible change.

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -26,6 +26,16 @@ jobs:
         # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
 
     steps:
+    # The following step is required in order to use .NET 8 pre-release.
+    # We can remove if using an officially supported .NET version.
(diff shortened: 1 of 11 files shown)


tjololo (approved):
I can't see any reason for supporting multi targeting for v8 either

PR #343, opened 2023-11-15
Mark IPageOrder [Obsolete] as it will be unused in frontend V4

ivarne:
Frontend v4 will remove support for tracks / calculatePageOrder / IPageOrder, and thus it should be removed in backend as well. Currently the thinking is that backend v8 should support frontend v3, but as IPageOrder isn't required for hiding pages anymore, I think it would be good to mark it `[Obsolete]`.

Thinking more about this. As we seem to agree that calculating page order in backend is wrong anyway, and we have an alternative, this should be accepted either way.

## Related Issue(s)
- One part of https://github.com/Altinn/app-frontend-react/issues/1491

## Verification
- [ ] All tests run green

## Documentation


tjololo (approved):
Agreed, marking it as obsolete seems like the correct thing to do

PR #344, opened 2023-11-15
Adding [Serializable] on custom exceptions in dotnet 8 causes warning

ivarne:
This code causes a warning in dotnet 8 and the explanation is that you should just remove it. I know I added this code because of a warning somewhere, but the docs are pretty clear that it is not used any more.

See: https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0051

If you created a custom type derived from System.Exception, consider whether you really need it to be serializable. It's likely that you don't need it to be serializable, as exception serialization is primarily intended to support remoting, and support for remoting was dropped in .NET Core 1.0.

## Verification
- [ ] All tests run green

## Documentation


tjololo (approved):
:godmode:

PR #346, opened 2023-11-20
V8 feature/304 support custom button

### LINKED ISSUE #304: Support for custom button for executing app specific logic in a task
### Description

API part of issue https://github.com/Altinn/app-frontend-react/issues/133

App developers have expressed the need for triggering custom logic when a user clicks a button in the app.
This can be for example fetching some data from a external api based on fields in the datamodel og generating other parts of the data based on already provided data.

# Implementation principles
Custom buttons should call a api backend that in turn delegates the actual work to the app developers code.
Following our current design the custom logic will be customisable by implementing a interface and registering it with the DI system in the application and thereby making a good separation of our code and the app developers code.

## In scope
- [ ] Access control - Custom buttons should be authorized the same ways as ActionButtons (buttons that control process flow)
- Make it possible to inform frontend what fields have changed in the datamodel, if any
- Make it possible to inform frontend if it should move to next/previous page
- Custom buttons and connected action should work for stateless apps
## Out of scope
- Custom button in stateless apps

### Questions needing answers
 - [ ] Do backend need additional information from frontend, besides the action name of the button
 - [ ] Endpoint url backend. Suggestion `{org}/{appname}/instances/{instanceOwnerID}/{instanceID}/action` how should we provide instanceOwnerPartyId and instanceId
 - [ ] Should actions be connected to a process step or is that something the implementer needs to take into account? (authorization)

_No response_

tjololo:
## Description
Introduce actions endpoint for use by custom buttons

## Related Issue(s)
- #304 

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Core/Features/Action/UserActionFactory.cs
```diff
@@ -21,14 +23,30 @@ public UserActionFactory(IEnumerable<IUserAction> actionHandlers)
...
-    public IUserAction GetActionHandler(string? actionId)
+    /// <returns>The first implementation of <see cref="IUserAction"/> that matches the actionId. If no match null is returned</returns>
+    public IUserAction? GetActionHandlerOrDefault(string? actionId)
+    {
+        if (actionId != null)
+        {
+            return _actionHandlers.FirstOrDefault(ah => ah.Id.Equals(actionId, StringComparison.OrdinalIgnoreCase));
+        }
+
+        return null;
```
RonnyB71, line 34: Do we want to return null here or the NullUserAction?

RonnyB71: Noticed the next method where you can specify the defaultAction so I guess that should be used in cases where you would like to have the NullUserAction

src/Altinn.App.Core/Helpers/MultiDecisionHelper.cs
```diff
@@ -0,0 +1,186 @@
...
+using Altinn.Common.PEP.Constants;
+using Altinn.Common.PEP.Helpers;
+using Altinn.Platform.Storage.Interface.Models;
+
+namespace Altinn.App.Core.Helpers;
+
+/// <summary>
+/// Helper class for multi decision requests.
+/// </summary>
+public static class MultiDecisionHelper
```
RonnyB71, line 13: Nice job 👏

src/Altinn.App.Core/Helpers/MultiDecisionHelper.cs
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private static (string? InstanceId, string InstanceGuid, string? Task, string InstanceOwnerPartyId, string Org, string App) GetInstanceProperties(Instance instance)
+    {
+        string? instanceId = instance.Id.Contains('/') ? instance.Id : null;
+        string instanceGuid = instance.Id.Contains('/') ? instance.Id.Split("/")[1] : instance.Id;
+        string? task = instance.Process?.CurrentTask?.ElementId;
+        string instanceOwnerPartyId = instance.InstanceOwner.PartyId;
+        string org = instance.Org;
+        string app = instance.AppId.Split("/")[1];
```
RonnyB71: Use AppIdentifier and avoid leaking the split logic and possibly return the AppIdentifier instead of Org and App?

src/Altinn.App.Core/Models/UserAction/FrontendAction.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Creates a gotoPage action
+    /// </summary>
+    /// <param name="frontendPage"></param>
+    /// <returns></returns>
+    public static FrontendAction GotoPage(string frontendPage)
+    {
+        var frontendAction = new FrontendAction()
+        {
+            Name = "gotoPage",
```
RonnyB71: Is this the one we need now, or is it navigateNextPage?

RonnyB71 (approved):
🕹️

src/Altinn.App.Core/Features/Action/UserActionFactory.cs
tjololo: Opted to return null here so that the aciton controller could return 404 if the user is authenticated to perform an action with a given name, but there is no registered service with that id. Think that would make debugging a typo easier.
The process engine defaults to the NullUserAction as there might not be a need for custom code just an action to make a decision in a following gateway

tjololo: Noticed that the controller didn't return 404. Have added the code for that with a test

src/Altinn.App.Core/Models/UserAction/FrontendAction.cs
tjololo: Correct we need the navigateToNextPage, will replace gotoPage with navigateToNextPage.

src/Altinn.App.Core/Helpers/MultiDecisionHelper.cs
tjololo: Thanks! :+1:

src/Altinn.App.Core/Models/UserAction/FrontendAction.cs
tjololo: After having a quick look at the frontend issue the correct name is `nextPage`

src/Altinn.App.Core/Infrastructure/Clients/Authorization/AuthorizationClient.cs
```diff
@@ -3,24 +3,23 @@
...
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
-
 using Newtonsoft.Json;
 
 namespace Altinn.App.Core.Infrastructure.Clients.Authorization
 {
-    /// <summary>
+        /// <summary>
```
ivarne:
```suggestion
    /// <summary>
```

src/Altinn.App.Core/Infrastructure/Clients/Storage/ProcessClient.cs
```diff
@@ -46,7 +46,7 @@ public class ProcessClient : IProcessClient
         /// <inheritdoc/>
         public Stream GetProcessDefinition()
         {
-            string bpmnFilePath = _appSettings.AppBasePath + _appSettings.ConfigurationFolder + _appSettings.ProcessFolder + _appSettings.ProcessFileName;
+            string bpmnFilePath = Path.Combine(_appSettings.AppBasePath , _appSettings.ConfigurationFolder , _appSettings.ProcessFolder , _appSettings.ProcessFileName);
```
ivarne:
```suggestion
            string bpmnFilePath = Path.Join(_appSettings.AppBasePath , _appSettings.ConfigurationFolder , _appSettings.ProcessFolder , _appSettings.ProcessFileName);
```
https://learn.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=net-8.0#remarks

> This method is intended to concatenate individual strings into a single string that represents a file path. However, if an argument other than the first contains a rooted path, any previous path components are ignored, and the returned string begins with that rooted path component. As an alternative to the Combine method, consider using the [Join](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.join?view=net-8.0) or [TryJoin](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.tryjoin?view=net-8.0) methods.

src/Altinn.App.Api/Controllers/ActionsController.cs
```diff
@@ -0,0 +1,130 @@
...
+                });
+            }
+
+            return new OkObjectResult(new UserActionResponse()
+            {
+                FrontendActions = result.FrontendActions,
+                UpdatedDataModels = result.UpdatedDataModels
+            });
+        }
+        catch (NotFoundException)
```
ivarne: Do we need a custom exception here? Can't `_userActionService` just return `null` if action is not found, or just add some magic in `UserActionResponse` so that it can indicate the HTTP status code (either through a `statusCode` property in `ActionError`, or by looking at the `Code` property and change the status if it is `ActionNotFound`)

src/Altinn.App.Api/Models/UserActionRequest.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    /// Action performed
+    /// </summary>
+    [JsonPropertyName("action")]
+    public string? Action { get; set; }
+    
+    /// <summary>
+    /// The id of the button that was clicked
+    /// </summary>
+    public string? ButtonId { get; set; }
```
ivarne, line 21:
```suggestion
    [JsonPropertyName("buttonId")]
    public string? ButtonId { get; set; }
```

test/Altinn.App.Api.Tests/Utils/PrincipalUtil.cs
```diff
@@ -7,21 +7,25 @@ namespace Altinn.App.Api.Tests.Utils
...
             string issuer = "www.altinn.no";
-            claims.Add(new Claim(ClaimTypes.NameIdentifier, userId.ToString(), ClaimValueTypes.String, issuer));
-            claims.Add(new Claim(AltinnCoreClaimTypes.UserId, userId.ToString(), ClaimValueTypes.String, issuer));
+            if (userId != null)
+            {
+                claims.Add(new Claim(ClaimTypes.NameIdentifier, userId.ToString()!, ClaimValueTypes.String, issuer));
+                claims.Add(new Claim(AltinnCoreClaimTypes.UserId, userId.ToString()!, ClaimValueTypes.String, issuer));
+                claims.Add(new Claim(AltinnCoreClaimTypes.PartyID, userId.ToString()!, ClaimValueTypes.Integer32, issuer));
+            }
+
```
ivarne, line 27: What problem does this solve?

ivarne (commented):
Tror muligens det er en del filer som mangler `[JsonPropertyName("")]`

src/Altinn.App.Api/Controllers/ActionsController.cs
tjololo: Actually the whole `UserActionService` could be removed now. The main purpose of that service was to add the, now removed, validationIssues to a validation group. As this concept is removed we can remove the service and handle the whole action not found in the controller without using exceptions, and remove a service.
I will rewrite the code

ivarne: If you rewrite, I'd rather you merge `UserActionFactory` into `UserActionService` instead of merging the service into the controller.

test/Altinn.App.Api.Tests/Utils/PrincipalUtil.cs
tjololo: `ClaimsPrincipalExtensions.GetUserIdAsInt` can return null for userid if the `urn:altinn:userid` is not set in the token. To test that path in ActionsController I marked userId as nullable and a if statement to not add the userClaims if userId is null

src/Altinn.App.Core/Internal/Process/ProcessEngine.cs
```diff
@@ -131,7 +131,7 @@ public async Task<ProcessChangeResult> Next(ProcessNextRequest request)
                 ErrorType = ProcessErrorType.Conflict
             };
         }
-        var actionHandler = await _userActionFactory.GetActionHandlerOrDefault(request.Action, new NullUserAction()).HandleAction(new UserActionContext(request.Instance, userId.Value));
+        var actionHandler = await _userActionService.GetActionHandlerOrDefault(request.Action, new NullUserAction()).HandleAction(new UserActionContext(request.Instance, userId.Value));
```
ivarne: Is it actually correct to have an action with no implementation, so that the `NullUserAction` is required?

test/Altinn.App.Core.Tests/Features/Action/UserActionServiceTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        IUserAction userAction = factory.GetActionHandlerOrDefault(null, new NullUserAction());
+
+        userAction.Should().BeOfType<NullUserAction>();
+        userAction.Id.Should().Be("null");
+    }
+    
+    internal class DummyUserAction : IUserAction
+    {
+        public string Id { get; set; } = "dummy";
```
ivarne: This means the same, but is prettier
```suggestion
        public string Id => "dummy";
```

test/Altinn.App.Core.Tests/Features/Action/UserActionServiceTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        public Task<UserActionResult> HandleAction(UserActionContext context)
+        {
+            return Task.FromResult(UserActionResult.SuccessResult());
+        }
+    }
+    
+    internal class DummyUserAction2 : IUserAction
+    {
+        public string Id { get; set; } = "dummy";
```
ivarne:
```suggestion
        public string Id => "dummy";
```

src/Altinn.App.Core/Internal/Process/Elements/UserAction.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <summary>
+    /// Defines an altinn action for a task
+    /// </summary>
+    public class UserAction
+    {
+        /// <summary>
+        /// Gets or sets the ID of the action
+        /// </summary>
+        [JsonPropertyName("id")]
+        public string Id { get; set; }
```
ivarne: I'm itching in my fingers to add this everywhere. Just curious when it will cause bugs.
```suggestion
        public required string Id { get; set; }
```

src/Altinn.App.Core/Models/UserAction/ActionError.cs
```diff
@@ -0,0 +1,27 @@
...
+/// <summary>
+/// Defines an error object that should be returned if the action fails
+/// </summary>
+public class ActionError
+{
+    /// <summary>
+    /// Machine readable error code
+    /// </summary>
+    [JsonPropertyName("code")]
+    public string Code { get; set; } = string.Empty;
```
ivarne: Could this also be `required` instead of setting a default?

ivarne (approved):
Just fix the code smell/warning, and I think this code is ready to run and find bugs runtime instead.

src/Altinn.App.Core/Internal/Process/ProcessEngine.cs
tjololo: You can use the action Id to navigate the process without having any custom logic connected to it. In those cases I think we shouldn't force the developer to implement a empty IUserAction and just default to the `NullUserAction`

src/Altinn.App.Core/Models/UserAction/ActionError.cs
tjololo: I'm not sure, our frontend code would probably not read this field, but it is meant for those applications with api consumers. Should we force apps without API consumers to set a field that is never used?

ivarne: That is why I'm asking. If it is optional, it might make more sense to have it nullable?

src/Altinn.App.Core/Internal/Process/ProcessEngine.cs
ivarne: In that case I'd rather check the null return value from `UserActionServcie.GetActionHandler`, so the code explicitly states how to handle a case without custom logic.

src/Altinn.App.Core/Models/UserAction/ActionError.cs
tjololo: good point, lets set it as nullable 👍

test/Altinn.App.Api.Tests/Utils/PrincipalUtil.cs
ivarne: Will have to look at this when https://github.com/Altinn/app-lib-dotnet/pull/353 and this is done.

ivarne:
🕹️

Issue #369, opened 2023-12-04
New validation extension


tjololo:
### Description

Backend issue for EPIC https://github.com/Altinn/app-template-dotnet/issues/180
Today custom validations performed in the backend is done by implementing the `IInstanceValidator` interface.
The code is triggered by frontend either just before the process/next call is performed or if singelfield validation is configured.
The same code is executed for both singel and full validation and is a place for mistakes and unclear logic.
Another problem is that if a previously error is fixed the code needs to generate a `"FIXED"` validation issue so that frontend can remove it.

The new implementation should support that one group of validations should be triggered if a field is changed, but all validation logic should be triggered when the process is tried to be moved to the next step.

The response from a group of validations should be the same either if it's ran as a singelfield validation or a full validation and the data is the same.

List of wanted functionality, this is a work in progress:
* A set of validations should be possible to group together as a set of fields as a whole could trigger/fix a issue.
* A validation group cannot clear or create issues for another validation group
* All validation should be executed and checks on process/next
* ValidationIssues are always group under their validation groups 
* It should be clear to the developer when an implementation of the new validation logic is triggered (what fields, if any, triggers the validation. Remember all validation are always checked on process/next)
* One field in the datamodel can be checked in different validation groups
* When a multipart save is triggered the validators with explicit mentions of one or more of the fields should be triggered.
* Validators without any fields defined should only be executed on process/next
* `Altinn.App.Core.Models.Validation.ValidationIssue` should be extended with `ValidationUrgency` to give the developer a way to define when the issue should be shown.

### Additional Information

Example of a repsone from a controller where one validation implementation is triggered:
```json
{
  "validationGroupName": [
    { //obj of type Altinn.App.Core.Models.Validation.ValidationIssue },
    { //obj of type Altinn.App.Core.Models.Validation.ValidationIssue }
  ]
}
```

Full validation with issues from multiple validation groups:
```json
{
  "validationGroupName1": [
    { //obj of type Altinn.App.Core.Models.Validation.ValidationIssue },
    { //obj of type Altinn.App.Core.Models.Validation.ValidationIssue }
  ],
  "validationGroupName2": [
    { //obj of type Altinn.App.Core.Models.Validation.ValidationIssue },
    { //obj of type Altinn.App.Core.Models.Validation.ValidationIssue }
  ]
}
```

olemartinorg:
> Validators without any fields defined should only be executed on process/next

[...]

PR #378, opened 2023-12-12
Simplify File analysis/validation

ivarne:
While working on rewriting the interfaces for validation to remove `ModelStateDictionary` and related infrastructure in #369 / #374, I took a look at File validation and think I found something to simplify there as well.

This PR has no functional changes, but introduces a new way to registrer file analysers and validators through keyed services, to match the new way of registrering validations.

```c#
services.AddKeyedTransient<IFileValidator, MyFileValidator>("dataTypeId")
```
instead of the more convoluted config where `myValidator.Id => "MY_VALIDATOR"` and `datatype.EnabledValidators[]` contains `"MY_VALIDATOR"`.

I think we should strive to use only one concept to register any type of validator, and the double approach used here is probably not a very good idea.

Note: This PR includes #369, so for review only look at [91bfcd1](https://github.com/Altinn/app-lib-dotnet/commit/91bfcd1a50a84dfcf6d5df4e354676cd64892ede)

## Verification
- [X] All tests run green (there were no tests 👿 )

## Documentation


ivarne:
We do not have any issue

PR #381, opened 2023-12-18
Add JsonEquivilentTo extension for FluentAssertions

ivarne:
See https://github.com/fluentassertions/fluentassertions/pull/2493

This greatly simplifies writing tests that ensure that a json response is the same as another without carefully ensuring that spaces are the same.

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/AppLibDotnet.sln
+++ b/AppLibDotnet.sln
@@ -18,6 +18,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{7AD5FADE-607
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Altinn.App.Core", "src\Altinn.App.Core\Altinn.App.Core.csproj", "{1745B251-BD5C-43B7-BA7D-9C4BFAB37535}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Altinn.App.TestUtils", "test\Altinn.App.TestUtils\Altinn.App.TestUtils.csproj", "{88930354-9AE0-4038-A9D5-70A3D3FF6404}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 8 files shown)


ivarne:
Hurra for tidssoner. Testene funker lokalt i GMT+1, men feiler i CI fordi de bruker GMT+0. Funket tidligere fordi vi deserialiserte til C# klasser og Equivalent sammenligner tidssone konvertering. Synes egentlig koden vår burde være stabil for tidssone, men det er vel litt utenfor scope nå.

tjololo:
I see the nice ness of the `BeJsonEquivalentTo` comparator, but not sure if this is the correct way. I would much rather leave our test as they are focus on providing an solution upstream to fluentassertion and wait for it to come as a part of the library.

If we choose to create a new nuget with altinn test utils I would rather have it in a separate repository. That way we can leverage github releases and have a cleaner way of releasing and sharing the testutils with the whole organisation and others.

At the same time as I'm writing this the change of these test I think exposed a mistake in our codebase, doing a quick search in the code I see that we in most cases use DateTime.UtcNow, but two places (excluding logging/console.write) we use DateTime.Now. Have created an isse #382 for further discussions on the timezone issue (as that seems to be out of scope for this PR as you say)

ivarne:
This was not intended to only fix this test, but something I'd like to be able to do for other tests as well. It is obviously a separate issue, so I just used it in an existing test for illustration. The fact that it caught an issue is in my opinion a validation of the testing approach.

>  I would much rather leave our test as they are focus on providing an solution upstream to fluentassertion and wait for it to come as a part of the library.

If I'm to try to contribute this functionality to fluentassertions, I probably need some exprience catching issues like this to help advocate for the change.

> If we choose to create a new nuget with altinn test utils I would rather have it in a separate repository. That way we can leverage github releases and have a cleaner way of releasing and sharing the testutils with the whole organisation and others.

Yes, If we find this useful to share and fluentassertions does not merge it in properly soon, we should probably maintain it as a separate git repo and nugget package. I don't see why we can't develop and catch issues here first. Versioning cross repo is not fun while developing this as a side project.

tjololo:
Not sure if the timezone difference  an actual bug, more of a codesmell, we should look into catching this with a linter/static code analysis tool. 
We should definitely fix it, but if the api consumer does not handle timezones thats a bug on their end IMO. We should strive to make it easier for the consumer, but timezones and the timeformat defining the timezone is there for a reason.
Unless it is necessary to have this functionality for delivering the functionality we have commited to this and the upcoming sprints I think we should hold back on this until those are delivered and see how the PR against fluentassertions progresses while we work on delivering the functionality to our users

ivarne:
> Not sure if the timezone difference an actual bug, more of a codesmell, we should look into catching this with a linter/static code analysis tool.

I'm not sure it is a bug in the code either. The main issue is that our test did not catch that the output isn't stable between systems.

PR #389, opened 2023-12-21
Chore/382 consistent timezone

### LINKED ISSUE #382: Remove use of DateTime.Now in ValidationAppSI and DefaultEformidlingService
### Description of the bug

In all other places where we need to get the current time we use DateTime.UtcNow. In `ValidationAppSI` and `DefaultEFormidlingService` we use DateTIme.Now. I the correct way is to use UtcNow as that will remove the connection to the timesone where the code is executed.
If UtcNow is wrong we at least need to be consistent across our codebase

### Steps To Reproduce

1. Create an app submit a for
2. See Timestampe in the instance `validated` block. This has timezone matching the computers timezone
3. Other timestamps are in utc

### Additional Information

_No response_

tjololo:
## Description
Fixes #382 
Some other changes to `DefaultEFormidlingService` to make testing easier.

## Related Issue(s)
- #382 
## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/EFormidling/Implementation/DefaultEFormidlingService.cs
+++ b/src/Altinn.App.Core/EFormidling/Implementation/DefaultEFormidlingService.cs
@@ -2,6 +2,7 @@ using Altinn.App.Core.Configuration;
 using Altinn.App.Core.Constants;
 using Altinn.App.Core.EFormidling.Interface;
 using Altinn.App.Core.Internal.App;
+using Altinn.App.Core.Internal.Auth;
 using Altinn.App.Core.Internal.Data;
 using Altinn.App.Core.Internal.Events;
 using Altinn.App.Core.Models;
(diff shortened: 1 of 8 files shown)


src/Altinn.App.Core/EFormidling/Implementation/DefaultEFormidlingService.cs
```diff
@@ -66,7 +65,7 @@ public class DefaultEFormidlingService : IEFormidlingService
     public async Task SendEFormidlingShipment(Instance instance)
     {
         if (_eFormidlingClient == null || _tokenGenerator == null || _eFormidlingMetadata == null ||
-            _appSettings == null || _platformSettings == null)
+            _appSettings == null || _platformSettings == null || _userTokenProvider == null)
```
bjorntore: I think this null check of _userTokenProvider can be removed?

The depdendency container in asp net will throw an error if it fails to construct and inject this dependency, since it's not optional. If you still want to null check it in case the class is instantiated manually, I would throw in the constructor instead.

bjorntore (approved):

ivarne (approved):

src/Altinn.App.Core/EFormidling/Implementation/DefaultEFormidlingService.cs
tjololo: You are right, I will remove the null check for _userTokenProvider

PR #393, opened 2024-01-08
New validation and partial validation via PATCH endpoint

ivarne:
Combination of #374 and #384 

## Patch
New endpoint on the DataController that allows frontends to update single (or multiple) fields and get updated validation on those fields.

Implement a new PATCH endpoint using https://docs.json-everything.net/path/basics/

### New service
**IValidationService**:
A ServiceLocator for validation that does the different kinds of validation. Some logic in data element validation to download the actual data if the element specifies AppLogic and can do that kind of deep validation.

### New interfaces:
* **`IValidationService`**: The rest of the systems does all its validation needs through this service. It does not mutate any state in instance, process or data elements. Old extension points are accessed through this. Gets services from a `SerivceProvider` with a service locator pattern.
  * `ValidateTask`: Validates that a task is ready for process next. Finds all relevant data elements and other validations to run.
  * `ValidateDataElement`: Used for partial or full validation of a single data element.
* **`ITaskValidator`**: Interface for App developers to add functionality for whole task validation.
  * Registered as a keyed service that specifies `TaskId` so that we know what tasks to trigger the validation for.
  * Does not support incremental validation.
* **`IDataElementValidator`**: Interface for validation of data elements without loading the actual content of the data element.
* **`IFormDataValidator`**: Interface for App developers to add incremental validation to form data elements.
  * Keyed service `DataType` to detect what Data elements to apply the validation to.
  * Methods `ShouldRun` and a `Code` that is automatically added to all validation issues allows many classes to be registered, but only run when relevant fields are changed. This means that the whole validation implementation needs to either run, or not run to ensure that we can tell front-end what validations has been updated and possibly fixed.

### New functionality
* **`GenericFormDataValidator`**: abstract class that app developers can use to simplify the implementation of `IFormDataValidator` adding a few convenience methods
  * `ValidateFormData` that gets the typed model, instead of `object data`
  * `CreateValidationIssue`: Very convenient way of adding errors.
    * Allows inputing an expression `m=>m.field.children[1].name` instead of a json path to bind the error to a field.
    * Uses `AsyncLocal` to not have to pass or return a list of issues. Just causes less clutter, and works even if the service is registered as a singleton.
* **`LinqExpressionHelpers`**: static methods to allow parsing of expressions instead of strings to create json paths.

### Validator registration
Validators can be registered in 3 ways
* `DataType` or `TaskId` is set as `"*"` and it is registered as a service in dependency injection
* `DataType`/`TaskId`, set as a specific DataType or task to run (Typically custom validator in a specific app)
* (As a keyed service, so that users override the `DataType`/`TaskId` and run on a general validator on a specific type)

### Compatibility
* **`IValidation`**, together with `ValidationAppSI` has been removed and all standard shared validation has been moved to default validators
  * **`DataAnnotationValidation`**: Runs attribute validation from `System.ComponentModel.DataAnnotations` (eg: `[RegEx()]`)
  * **`DefaultDataElementValidation`**: Validates maxSize, contentType, and fileScan
  * **`DefaultTaskValidator`**: Validates min and max count for data elements
  * **`ExpressionValidator`**: Validates data elements with app logic and ensures that expression validation has run.
  * **`LegacyIInstanceValidatorFormDataValidator`**: Runs existing `IInstanceValidator.ValidateData` class registered in dependency injection for backwards compatibility
  * **`LegacyIInstanceValidatorTaskValidator`**: Runs existing `IInstanceValidator.ValidateTask` class registered in dependency injection for backwards compatibility
  * **`RequiredLayoutValidator`**: Runs expressions in layout to ensure that every visible component with `"required"` actually has data.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Api/Controllers/ValidateController.cs
```diff
@@ -130,19 +127,21 @@ public class ValidateController : ControllerBase
                 throw new ValidationException("Unknown element type.");
             }
 
-            messages.AddRange(await _validationService.ValidateDataElement(instance, dataType, element));
+            messages.AddRange(await _validationService.ValidateDataElement(instance, element, dataType));
 
             string taskId = instance.Process.CurrentTask.ElementId;
+
+            // Should this be a BadRequest instead?
```
tjololo, line 133: Kind of wondering the same thing myself. What would be the usecase for validation a datatype that potentially could be locked.

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -353,6 +366,167 @@ private static bool FileValidationEnabledForDataType(DataType dataTypeFromMetada
...
+        /// <summary>
+        /// Part of <see cref="PatchFormData" /> that is separated out for testing purposes.
+        /// </summary>
+        /// <param name="dataType">The type of the data element</param>
+        /// <param name="dataElement">The data element</param>
+        /// <param name="dataPatchRequest">Container object for the <see cref="JsonPatch" /> and list of ignored validators</param>
+        /// <param name="oldModel">The old state of the form data</param>
+        /// <param name="instance">The instance</param>
+        /// <returns>DataPatchResponse after this patch operation</returns>
+        internal async Task<(DataPatchResponse, ProblemDetails?)> PatchFormDataImplementation(DataType dataType, DataElement dataElement, DataPatchRequest dataPatchRequest, object oldModel, Instance instance)
```
tjololo: Could this be extracted into a service class? The controller is "starting" to get rather big

src/Altinn.App.Core/Features/IDataElementValidator.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// The data type that this validator should run for. This is the id of the data type from applicationmetadata.json
+    /// </summary>
+    /// <remarks>
+    ///
+    /// </remarks>
+    string DataType { get; }
+
+    /// <summary>
+    /// Override this if you want a different name for the validation source
```
tjololo: Document the default value?

src/Altinn.App.Core/Features/IFormDataValidator.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <param name="changedFields">List of the json path to all changed fields for incremental validation</param>
+    bool ShouldRun(List<string> changedFields);
+
+    /// <summary>
+    /// Returns the group id of the validator. This is used to run partial validations on the backend.
+    /// </summary>
+    /// <remarks>
+    /// The default implementation should work for most cases.
+    /// </remarks>
+    public string ValidationSource => $"{this.GetType().FullName}-{DataType}";
```
tjololo, line 32: Should the summary specify that this is optional to define and what the default is?

src/Altinn.App.Core/Features/ITaskValidator.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// public MyTaskValidator([ServiceKey] string taskId)
+    /// {
+    ///     TaskId = taskId;
+    /// }
+    /// </code>
+    /// </example>
+    string TaskId { get; }
+
+    /// <summary>
+    /// Override this if you want a different name for the validation source
```
tjololo: Document default?

src/Altinn.App.Core/Features/Validation/IValidationService.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Altinn.App.Core.Features.Validation;
+
+/// <summary>
+/// Core interface for validation of instances. Only a single implementation of this interface should exist in the app.
+/// </summary>
+public interface IValidationService
+{
+    /// <summary>
+    /// Validates the instance with all data elements on the current task and ensures that the instance is read for process next.
```
tjololo:
```suggestion
    /// Validates the instance with all data elements on the current task and ensures that the instance is ready for process next.
```

src/Altinn.App.Core/Features/Validation/ValidationService.cs
```diff
@@ -0,0 +1,156 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Altinn.App.Core.Features.Validation;
+
+/// <summary>
+/// Main validation service that encapsulates all validation logic
+/// </summary>
+public class ValidationService : IValidationService
+{
+    private readonly IServiceProvider _serviceProvider;
```
tjololo, line 16: Can we be more specific here and inject IEnumerable of the interfaces we actually use?
This hides all the information about what services this class uses

src/Altinn.App.Core/Features/Validation/ValidationService.cs
```diff
@@ -0,0 +1,156 @@
...
+        // Run extra validation on form data elements with app logic
+        if(dataType.AppLogic?.ClassRef is not null)
+        {
+            Type modelType = _appModel.GetModelType(dataType.AppLogic.ClassRef);
+
+            Guid instanceGuid = Guid.Parse(instance.Id.Split("/")[1]);
+            string app = instance.AppId.Split("/")[1];
+            int instanceOwnerPartyId = int.Parse(instance.InstanceOwner.PartyId);
+            var data = await _dataClient.GetFormData(instanceGuid, modelType, instance.Org, app, instanceOwnerPartyId, Guid.Parse(dataElement.Id)); // TODO: Add method that accepts instance and dataElement
+            var formDataIssuesDictionary = await ValidateFormData(instance, dataElement, dataType, data, null);
```
tjololo: No need to define the default value
```suggestion
            var formDataIssuesDictionary = await ValidateFormData(instance, dataElement, dataType, data);
```

src/Altinn.App.Core/Features/Validation/ValidationService.cs
```diff
@@ -0,0 +1,156 @@
...
+        ArgumentNullException.ThrowIfNull(dataElement);
+        ArgumentNullException.ThrowIfNull(dataElement.DataType);
+        ArgumentNullException.ThrowIfNull(data);
+
+        // Locate the relevant data validator services from normal and keyed services
+        var dataValidators = _serviceProvider.GetServices<IFormDataValidator>()
+            .Where(dv => dv.DataType == "*" || dv.DataType == dataType.Id)
+            // .Concat(_serviceProvider.GetKeyedServices<IFormDataValidator>(dataElement.DataType))
+            .Where(dv => ignoredValidators?.Contains(dv.ValidationSource) != true)
+            .Where(dv => changedFields is null ? true : dv.ShouldRun(changedFields))
```
tjololo: As suggested by codeQL
```suggestion
            .Where(dv => changedFields is null || dv.ShouldRun(changedFields))
```

src/Altinn.App.Core/Helpers/ObjectUtils.cs
```diff
@@ -0,0 +1,53 @@
...
+/// <summary>
+/// Utilities for working with model instances
+/// </summary>
+public static class ObjectUtils
+{
+    /// <summary>
+    /// Recursively initialize all <see cref="List{T}"/> properties on the object that are currently null
+    /// </summary>
+    /// <param name="model"></param>
+    public static void InitializeListsRecursively(object model)
```
tjololo: Think I need a walkthrough of the why. I believe you, but would be great to have an actual live demo and see the actual why :)
Also seems to me like this method does more than initialize List now

src/Altinn.App.Core/Internal/Expressions/LayoutEvaluatorState.cs
```diff
@@ -179,6 +179,14 @@ public ComponentContext GetComponentContext(string pageName, string componentId,
         return _dataModel.GetModelData(key, context?.RowIndices);
     }
 
+    /// <summary>
+    /// Get all of the resoved keys (including all possible indexes) from a data model key
```
tjololo:
```suggestion
    /// Get all of the resolved keys (including all possible indexes) from a data model key
```

tjololo (commented):
Some initial comments. Haven't looked at the test code yet, but wanted to get the first src comments out the door

src/Altinn.App.Core/Features/Validation/ValidationService.cs
ivarne: Using the ServiceProvider directly is required if we want to have support for Keyed services. Currently those lines adding that support are commented out, so currently it doesn't matter.

My opinion is that there isn't much practical difference between an ienumerable or a service provider, and we're still having a servicelocator pattern because we filter the ienumerable

src/Altinn.App.Core/Helpers/ObjectUtils.cs
ivarne: Serializing a C# class to xml and back to C# again changes null lists to empty lists and some empty strings to null strings.

We do the same changes so that the full model we return for generating patches, is the same as the model we fetch from storage when applying the patch.

src/Altinn.App.Api/Controllers/ProcessController.cs
```diff
@@ -202,15 +203,15 @@ public class ProcessController : ControllerBase
...
+        private async Task<bool> CanTaskBeEnded(Instance instance, string currentTaskId)
         {
             List<ValidationIssue> validationIssues = new List<ValidationIssue>();
 
             bool canEndTask;
 
             if (instance.Process?.CurrentTask?.Validated == null || !instance.Process.CurrentTask.Validated.CanCompleteTask)
             {
-                validationIssues = await _validationService.ValidateAndUpdateProcess(instance, currentElementId);
+                validationIssues = await _validationService.ValidateInstanceAtTask(instance, currentTaskId);
```
ivarne: I've stopped updating `instance.Process.CurrentTask.Validated.CanCompleteTask` in the `_validationSerivce` and not added it back in anywhere, so we should either add it back or remove this check. It really does not make any sense to cache `CanCompleteTask` without locking all data elements that the validation is dependent on.

src/Altinn.App.Core/Features/Validation/Default/DataAnnotationValidator.cs
```diff
@@ -0,0 +1,69 @@
...
+            ValidationStateDictionary validationState = new ValidationStateDictionary();
+            _objectModelValidator.Validate(actionContext, validationState, null!, data);
+
+            return Task.FromResult(ModelStateHelpers.ModelStateToIssueList(modelState, instance, dataElement, _generalSettings, data.GetType(), ValidationIssueSources.ModelState));
+        }
+        catch (Exception e)
+        {
+            return Task.FromException<List<ValidationIssue>>(e);
+        }
+    }
```
ivarne, line 68: Typisk. Her prøver jeg å være flink og unngå async og heller bruke `Task.FromResult` og `Task.FromException`, men så får jeg bare en ny warning. Heller sterkt mot å bare skru av CS1998, så man kan skrive vanlig async kode og få exception med riktig stacktrace på rett sted.

src/Altinn.App.Api/Controllers/DataController.cs
ivarne: Yes, it could. I made an attempt, but it was hard to write a good summary for what that service class would do. and draw reasonable separation lines.

```c#
/// <summary>
/// Service class for the <see cref="DataController.PatchFormData"> because the controller was getting "rather big"
/// </summary>
```

Doesn't really seem to justify why an extra file is required.

tjololo: A service for handling patch of a datamodel doesn't seem like a totally unreasonable service

src/Altinn.App.Core/Features/Validation/ValidationService.cs
tjololo: By injecting the ServiceProvider any service could be used in this service and as we aren't currently using keyedservices I think filtering IEnumerable of a given interface is a much more explicit way of doing it.

src/Altinn.App.Api/Controllers/ProcessController.cs
tjololo: So now we run the actual validations when we try to move the process to the next step instead of relying on a possible outdated status?

src/Altinn.App.Core/Features/Validation/Default/DataAnnotationValidator.cs
tjololo: Hvis det gir mening å catche Exception her så dismisser vi bare notisen fra CodeQL :)

src/Altinn.App.Api/Controllers/ProcessController.cs
ivarne: Yes, when I've stopped updating the status, we need to run validation again before processing next. We either need to remove the check here, or add back the caching when we validate.

tjololo: If i remeber correctly we agreed that we should run validation implicitly when process next is called. That decision still holds true, I belive, so I think we just should remove the check on the cached value.

ivarne: Yes, I'll remove this check. Question is what to do with instance.Process.CurrentTask.Validated.CanCompleteTask, if we don't use it anymore.

src/Altinn.App.Api/Controllers/DataController.cs
ivarne: I tried, but I don't think the code organisation is improved. 6d51c4db125bd6ec1b35a24b4dcbd572e127f2be

src/Altinn.App.Api/Controllers/ProcessController.cs
tjololo: I think we need to leave it, as the object/field is "owned" by storage and storage needs to support older versions of app-lib that rely on it.

src/Altinn.App.Api/Controllers/DataController.cs
tjololo: I think it makes sense as we don't need the InternalsVisibleFor in the project file. Making the code easier to test is IMO good enough reason to split the code into a separate class

RonnyB71 (approved):
Approving in order to merge, leaving the open conversations to be resolved later.

PR #394, opened 2024-01-09
Update expression tests after splitting up Group component.

Magnusrm:
Som test config had to be updated in the frontend version of the expression tests after splitting out `RepeatingGroup` from the `Group` component. 

Mirroring the changes in the test configs in this PR.


ivarne:
https://github.com/Altinn/app-lib-dotnet/commit/adac4934d36464575cc2df33af34e35450ced101 is a crude and backwards compatible way to fix the tests. If it isn't correct we need more tests.

ivarne (approved):

PR #402, opened 2024-01-21
v7: Fix issue with non-ascii characters in code list parameters not being correctly encoded in parameter header.

ivarne:
I tried adding some tests, but the ascii validation from the stack trace in https://github.com/Altinn/codelists-lib-dotnet/issues/23, does not seem to trigger with WebApplicationFactory. Thus this requires a manual test with an app.

## Related Issue(s)
- Likely fix https://github.com/Altinn/codelists-lib-dotnet/issues/23

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #404, opened 2024-01-24
merge changes from v7 and main into v8 branch.

ivarne:
Mostly dependencies, but also https://github.com/Altinn/app-lib-dotnet/pull/402


tjololo:
As PRs in this repo only allow "squash and merge" would this result in merge conflicts if this is merged?

ivarne:
I don't think so in this case, but it is often a problem so we should allow normal merges in some cases.

tjololo (approved):
If we belive this wouldn't result in merge conflicts I see no problems. In addition all these changes have already been review :+1:

PR #405, opened 2024-01-24
Add language to IDataProcessor and as an optional query param to apis

ivarne:
In data processing it is sometimes valuable to know the language the user currently is showing in the schema. 

At least two cases comes to mind:
1. Saving the language to the data model for knowing the language to use for potential follow up with the reportee
2. Updating presentation values with correct language (`Uke 2 2023` vs `Week 2 2023`).

This PR also adds a Roslyn rewrite to fix the interface.

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/16
- (Alternative solution for) https://github.com/Altinn/app-frontend-react/issues/1661

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #406, opened 2024-01-24
Make language nullable in option lists and data lists

ivarne:
Make language parameter nullable to match dataProcessWrite in #405 
Now that we can rewrite interfaces to nullable in all apps on upgrade, it makes much more sense to make language nullable than to have `"nb"` as a default language.

Note that functionally this is a revert of some changes in #328, so should be good.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/OptionsController.cs
+++ b/src/Altinn.App.Api/Controllers/OptionsController.cs
@@ -34,16 +34,16 @@ namespace Altinn.App.Api.Controllers
         /// Api that exposes app related options
         /// </summary>
         /// <param name="optionsId">The optionsId</param>
+        /// <param name="queryParams">Query parameters supplied</param>
         /// <param name="language">The language selected by the user.</param>
-        /// <param name="queryParams">Query parameteres supplied</param>
         /// <returns>The options list</returns>
         [HttpGet("{optionsId}")]
         public async Task<IActionResult> Get(
             [FromRoute] string optionsId,
-            [FromQuery] string? language,
-            [FromQuery] Dictionary<string, string> queryParams)
+            [FromQuery] Dictionary<string, string> queryParams,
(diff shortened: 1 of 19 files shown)


src/Altinn.App.Api/Controllers/OptionsController.cs
```diff
@@ -34,16 +34,16 @@ public OptionsController(IAppOptionsService appOptionsService)
...
         /// <returns>The options list</returns>
         [HttpGet("{optionsId}")]
         public async Task<IActionResult> Get(
             [FromRoute] string optionsId,
-            [FromQuery] string? language,
-            [FromQuery] Dictionary<string, string> queryParams)
+            [FromQuery] Dictionary<string, string> queryParams,
+            [FromQuery] string? language = null)
         {
-            AppOptions appOptions = await _appOptionsService.GetOptionsAsync(optionsId, language ?? "nb", queryParams);
```
tjololo, line 46: Agreeing that changing this from default "nb" to null makes sense, but in doing so we need to make a new version of https://github.com/Altinn/codelists-lib-dotnet. Some external codelists might expect "nb" as default, but that should be an implementational detail if each `IAppOptionsProvider` implementation.
If we actually decide to remove the nb-default we should create an issue in that repo

tjololo (commented):
Looks good to me, but would like some input from @RonnyB71 as he has developed the codelists-lib-dotnet project that is the code that is most heavily affected by the removal of `nb` as default language

ivarne:
`"nb"` as default was introduced in https://github.com/Altinn/app-lib-dotnet/pull/328, and no one can depend on it working. This is just reverting to the v7 behavior and making it explicit.

tjololo (approved):

PR #418, opened 2024-01-30
Always return 200 for success in PUT formdata api

### LINKED ISSUE #52: Form data save should return HTTP 200/201
## Describe the bug
When an app contains calculation in the `CalculationHandler.cs`, a PUT call to update the formdate of the instance returns a `303 See other` as response code without a Location in the response headers. And k6 tests fail expecting a location response header on a 303 response. 

Refer this [link](https://datatracker.ietf.org/doc/html/rfc7231#section-6.4.4) where it says location is must on 303 responses.

## To Reproduce
1. Start an app instance that has logic defined in the `calculationhandler.cs`
2. Send a PUT call to update the form data of the instance
3. The response code is 303 and the response header misses the parameter location.

## Expected behavior
303 responses should be have location in the response headers

## Additional info
env: at21
app: ttd/security-level1-app

## Acceptance criteria
- [ ] Should work with V3 of frontend.

ivarne:
As requested in https://github.com/Altinn/app-lib-dotnet/issues/52

Not sure this change is worth it, but we should not keep simple issues open for so long and v8 is a great time to change this.

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/52

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/DataController.cs
+++ b/src/Altinn.App.Api/Controllers/DataController.cs
@@ -317,7 +317,7 @@ namespace Altinn.App.Api.Controllers
         [DisableFormValueModelBinding]
         [RequestSizeLimit(REQUEST_SIZE_LIMIT)]
         [ProducesResponseType(typeof(DataElement), 201)]
-        [ProducesResponseType(typeof(CalculationResult), 303)]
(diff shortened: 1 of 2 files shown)


tjololo (approved):
:godmode:

PR #419, opened 2024-01-30
Remove old pdf generation (it probably doesn't work for any newish app)

### LINKED ISSUE #415: Remove support for old pdf in v8 and use the new one by default
### Description

The old pdf is buggy, and isn't compatible with the layout files that v4 use anyway.

### Additional Information

_No response_

### Tasks

_No response_

### Acceptance Criterias

_No response_

ivarne:
Not sure what to do about `IPdfFormatter` there is an api that frontend uses in PDF generation, but only if we don't have a pdf specific layout. I see that it is used in 33 production apps, so keeping it for now and marking it `[Obsolete]` might be the best right now.

## Related Issue(s)
- #415 

## Verification
- [ ] All tests run green

## Documentation


tjololo (approved):
:godmode:

Issue #420, opened 2024-01-31
Validation should include the current language


ivarne:
### Description

Some validations might want to know the language the user is currently using to provide a better experience without using the translation system on frontend.

Issue #421, opened 2024-01-31
Remove Service Locator pattern from IValidationService


ivarne:
### Description

ServiceLocator was a feared pattern.

PR #423, opened 2024-02-01
Add string? language to all validation functions

ivarne:
Validations might sometimes want to know the current user language to customise validation text when it is hard to use the translation system.

This PR also includes a commit that does the following
* Extract a `IValidatorFactory` interface from `ValidatorService` to make a clear distinction between finding the correct validator and executing the validator
* Move lots of validation classes/interfaces to `Altinn.App.Core.Internal.Validation`, as they are not valid extension points for app developers.

## Related Issue(s)
- fixes https://github.com/Altinn/app-lib-dotnet/issues/420
- similar pattern to https://github.com/Altinn/app-lib-dotnet/pull/405
- https://github.com/Altinn/app-lib-dotnet/issues/421

## Verification
- [ ] All tests run green

## Documentation


ivarne:
@olemartinorg This adds the expectation that frontend adds the `?language=${language}` query parameter in
* `{org}/{app}/instances/{instanceOwnerPartyId:int}/{instanceGuid:guid}/process/completeProcess` (not sure if frontend uses this)
* `validation` (still used for that one initial call)
* `instance/../data` (previously added in https://github.com/Altinn/app-lib-dotnet/pull/405)

Are they OK, or should I open an issue for frontend?

olemartinorg:
Fantastic! Yep, would be nice with an issue for frontend here, so that it's easily visible in the board, etc. Could just be a copy of your comment here.. :wink:

src/Altinn.App.Core/Features/IFormDataValidator.cs
```diff
@@ -37,6 +37,7 @@ public interface IFormDataValidator
     /// <param name="instance"></param>
     /// <param name="dataElement"></param>
     /// <param name="data"></param>
+    /// <param name="language">Current language used by the user</param>
```
tjololo:
```suggestion
    /// <param name="language">The currently used language by the user (or null if not available)</param>
```

Sync it with the rest of the language descriptions

src/Altinn.App.Core/Internal/Validation/IValidatorFactory.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// <summary>
+    /// Gets all form data validators for a given data element.
+    /// </summary>
+    public IEnumerable<IFormDataValidator> GetFormDataValidators(string dataTypeId);
+}
+
+/// <summary>
+/// Implementation of <see cref="IValidatorFactory"/> that uses the <see cref="IServiceProvider"/> to locate validators.
+/// </summary>
+public class ValidatorFactoryWithLocator : IValidatorFactory
```
tjololo: Do we need this class? I like the `ValidatorFactory` better and I can't see that this code is used anywhere

test/Altinn.App.Api.Tests/Controllers/DataController_PatchFormDataImplementation.cs
```diff
@@ -62,7 +62,7 @@ public DataController_PatchFormDataImplementation()
         _serviceCollection.AddSingleton(_dataElementValidator);
         _serviceProvider = _serviceCollection.BuildServiceProvider();
         var validationService = new ValidationService(
-            _serviceProvider,
+            new ValidatorFactory(_serviceProvider.GetServices<ITaskValidator>(), _serviceProvider.GetServices<IDataElementValidator>(), _serviceProvider.GetServices<IFormDataValidator>()),
```
tjololo: Any reason for registering services and building a serviceProvider? Mocks can, as far as I can see, be injected directly into the ValidationFactory. I think this in turn could remove the need for implementing IAsyncDisposable.

ivarne: The reason is that the previous implementation used a ServiceProivder and I did not see any reason to rewrite this testing logic.

src/Altinn.App.Core/Internal/Validation/IValidatorFactory.cs
ivarne: No, it is not used anywhere. Initially I thought I'd use it for some of the tests, but both implementations worked with no issues, so there was no reason.

tjololo: Then I think we should remove this code as its not used

test/Altinn.App.Api.Tests/Controllers/DataController_PatchFormDataImplementation.cs
tjololo: I think we should strive to keep the test as simple as possible for the next one visiting this part of the code

src/Altinn.App.Core/Internal/Validation/IValidatorFactory.cs
ivarne: Done

test/Altinn.App.Api.Tests/Controllers/DataController_PatchFormDataImplementation.cs
ivarne: Done

tjololo (approved):

PR #417, opened 2024-01-30
Support number, boolean and null as AppOption values

ivarne:
Add support for multiple types as AppOptionValues. The internal representation is still string for simplicity for app developers and backwards compatibility, but you can specify a type for json serialisation.

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/413

## Verification
- [x] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #426, opened 2024-02-04
Ensure app options gets parsed with System.Text.Json (and let API return 404)

ivarne:
In https://github.com/Altinn/app-lib-dotnet/pull/417 I assumed that System.Text.Json would be used for parsing as well as serialization. While removing newtonsoft I discovered that it was not the case for app options in files, and while writing tests, I discovered that the endpoint never returns 404 and fixed that issue aswell.

## Related Issue(s)
- #417

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Features/Options/AppOptionsFileHandler.cs
+++ b/src/Altinn.App.Core/Features/Options/AppOptionsFileHandler.cs
@@ -1,9 +1,9 @@
 using System.Text;
+using System.Text.Json;
 using Altinn.App.Core.Configuration;
 using Altinn.App.Core.Helpers;
 using Altinn.App.Core.Models;
 using Microsoft.Extensions.Options;
-using Newtonsoft.Json;
 
 namespace Altinn.App.Core.Features.Options
 {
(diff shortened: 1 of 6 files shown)


src/Altinn.App.Core/Features/Options/AppOptionsFileHandler.cs
```diff
@@ -21,20 +26,20 @@ public AppOptionsFileHandler(IOptions<AppSettings> settings)
...
             if (File.Exists(filename))
             {
                 string fileData = await File.ReadAllTextAsync(filename, Encoding.UTF8);
-                List<AppOption> options = JsonConvert.DeserializeObject<List<AppOption>>(fileData)!;
+                List<AppOption> options = JsonSerializer.Deserialize<List<AppOption>>(fileData, JSON_SERIALIZER_SETTINGS)!;
                 return options;
             }
 
-            return new List<AppOption>();
+            return null;
```
RonnyB71, line 42: I think we previously have said that for enumerable types we generally return an empty list instead of null. Any reason to change this?

ivarne: The most obvious reason is that the controller checks for NULL in order to return a 404 status, and that functionality is currently broken.

The other, more general is that there is a difference between `MY-OPTIONS` has no values, and that `MY-OPTIONS` does not exist. In other cases where there is no difference between "this instance has no data elements" and "the list of data elements is empty", the policy makes sense.

With nullable reference types the policy does not make much sense anymore in my opinion, but that is a different discussion.

RonnyB71 (approved):
🕹️

src/Altinn.App.Core/Features/Options/AppOptionsFileHandler.cs
RonnyB71: 👍

PR #429, opened 2024-02-07
Don't run ProcessDataRead for locked data elements

ivarne:
I'm working on writing tests, but this change is mostly about changing expected behaviour, not really testable logic.

The location I added the check for locked has a few behaviour changes I think is fine
* I don't run `ProcessDataRead` for locked elements (it seems strange to fake change data)
* I don't set `ReadStatus` to `read` for locked elements. (it is set when fetching an instance, so I'm puzzled how anyone is supposed to get a data element with a unknown guid without first fetching the instance).

## Related Issue(s)
- I don't think an issue was created.

## Verification
- [ ] All tests run green

## Documentation


RonnyB71 (approved):
🕹️

PR #431, opened 2024-02-07
fix: rename Name to Id in ClientActions

mikaelrss:
## Description
This fixes a bug where users of the CustomButton that returned ClientActions were seeing an error message, and no ClientAction being performed.

## Related Issue(s)
- #430 

## Verification
- [ ] All tests run green

## Documentation
- [x] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable) Documentation is already up to date.

### DIFF
--- a/src/Altinn.App.Core/Models/UserAction/ClientAction.cs
+++ b/src/Altinn.App.Core/Models/UserAction/ClientAction.cs
@@ -8,10 +8,10 @@ namespace Altinn.App.Core.Models.UserAction;
 public class ClientAction
 {
     /// <summary>
-    /// Name of the action
+    /// Id of the action
     /// </summary>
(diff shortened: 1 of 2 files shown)


ivarne (approved):
Just to be certain I understand

This is because frontend used `id` while backend used `name`. Can we add a reference (in a comment or something) so that it is easier to find the usages?

mikaelrss:
> Just to be certain I understand
> 
> This is because frontend used `id` while backend used `name`. Can we add a reference (in a comment or something) so that it is easier to find the usages?

Yes. The frontend uses the Id property of a `ClientAction` to identify what to run in the frontend. I'll update the comment on the property!

PR #433, opened 2024-02-08
Rename the ?lang= parameter in process/next to ?language

ivarne:
This is to match other endpoints and make a more uniform api

Obviously while keeping backwards compatibility

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/issues/1774

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/ProcessController.cs
+++ b/src/Altinn.App.Api/Controllers/ProcessController.cs
@@ -219,10 +219,7 @@ namespace Altinn.App.Api.Controllers
         /// <param name="instanceOwnerPartyId">unique id of the party that is the owner of the instance</param>
         /// <param name="instanceGuid">unique id to identify the instance</param>
         /// <param name="elementId">obsolete: alias for action</param>
-        /// <param name="lang">Optional parameter to pass on the language used in the form if this differs from the profile language,
-        /// which otherwise is used automatically. The language is picked up when generating the PDF when leaving a step, 
-        /// and is not used for anything else.
-        /// </param>
+        /// <param name="language">Signal the language to use for pdf generation, error messages...</param>
         [HttpPut("next")]
         [ProducesResponseType(StatusCodes.Status200OK)]
         [ProducesResponseType(StatusCodes.Status404NotFound)]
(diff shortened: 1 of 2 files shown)


tjololo:
How should we label this so it's clearly marked in the release notes? It's not a breaking change since we support the old param, but we should notify other api consumers that are potentially using this that the param has changed from `lang` to `language` and that we are deprecating the `lang` param, if that is what we are doing?

ivarne:
> How should we label this so it's clearly marked in the release notes? It's not a breaking change since we support the old param, but we should notify other api consumers that are potentially using this that the param has changed from lang to language and that we are deprecating the lang param, if that is what we are doing?

Yes, that is what I was doing. I initially thought this was only for our code, but obviously someone might have used `IHttpContextFactory` to get this value in one of the interface hooks. It seems far fetched, but might be possible. Maybe just search all repos to see if someone has used it?

tjololo:
> Yes, that is what I was doing. I initially thought this was only for our code, but obviously someone might have used `IHttpContextFactory` to get this value in one of the interface hooks. It seems far fetched, but might be possible. Maybe just search all repos to see if someone has used it?

I did a quick search in studio through giteas explore mode for "lang" can't see any C# code that are accessing the `IHttpContextFactory` to extract `lang` any where. Another factor is the consumers of the apps api, some applications are not tied to our frontend but used directly through the API. When a ServiceOwner upgrades his application to v8 we need to tell that consumers of the API should use `language` instead of `lang` when they call process/next, if they are using it.
This will then give them a "grace"-period where both work and the can migrate their solutions using the apps API

ivarne:
> This will then give them a "grace"-period where both work and the can migrate their solutions using the apps API

Hmm... This PR only makes both versions work for pdf language, not potential future usages like other process tasks. Do you think I should add compatibility on the controller as well? It seems kind of strange as it isn't used right now, but would be a sign for the future that clients might set `lang` instead of `language`

tjololo:
No didn't mean we should add it to the controller, just that we need to inform the app developers that they need to inform their api consumers about this change

tjololo (approved):
We should document this in the changelog and release notes. I agree that we should use the same param name in all api endpoint

RonnyB71 (approved):
🕹️

PR #441, opened 2024-02-13
AltinnRowId support

ivarne:
This PR assumes models will be extended with the following property

```C#
    [XmlAttribute("altinnRowId")]
    [JsonPropertyName("altinnRowId")]
    [System.Text.Json.Serialization.JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
    public Guid AltinnRowId { get; set; };

    public bool AltinnRowIdSpecified()
    {
        return AltinnRowId != default;
    }

```
## Description
PS: There should be more automated tests for this, but I don't think that prevents review.

Most significant changes:
* Added code for initialising and cleaning AltinnRowIds
* `IUserAction` related endpoints now runs validation
* fixed an error in `UserActionResult` that stated that dataType should be used as key for the updated models instead of dataElementId.
* Declare the result dictionary in `IUserAction` to be not nullable for the updated elements. It makes no sense to specify an element without the updated data for that element.

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/issues/1812

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Api/Controllers/ActionsController.cs
```diff
@@ -156,7 +156,7 @@ public class ActionsController : ControllerBase
     private async Task SaveChangedModels(Instance instance, Dictionary<string, object?> resultUpdatedDataModels)
     {
         var instanceIdentifier = new InstanceIdentifier(instance);
-        foreach (var (dataType, newModel) in resultUpdatedDataModels)
+        foreach (var (elementId, newModel) in resultUpdatedDataModels)
```
ivarne, line 160: The comment about what the IUserAction should return states 

https://github.com/Altinn/app-lib-dotnet/blob/3f2dbd03700122cf9f59d3350686010f73513900/src/Altinn.App.Core/Models/UserAction/UserActionResult.cs#L17-L20

But I'm really happy if frontend used the more sensible elementId from the start. It removes one paragraph from https://github.com/Altinn/app-lib-dotnet/issues/446

olemartinorg: Huh, I didn't know - that comment must be wrong. The key is, and must be, the elementId. I changed this here because existing user actions broke. :thinking:

ivarne: If it broke existing actions that is really great 🎉🎉🎉🎉

Much easier to fix the comment than to make kompatibility with both options.

olemartinorg:
One more path that creates models but don't currently inject row IDs: `IInstantiationProcessor.DataCreation()`. We have a test for this in `ttd/frontend-test`, where we create Likert rows this way.

ivarne:
> One more path that creates models but don't currently inject row IDs: IInstantiationProcessor.DataCreation(). We have a test for this in ttd/frontend-test, where we create Likert rows this way.

What API gets called in that case? As far as I'm able to read all code paths leads to an api that returns a url or a `DataElement`, which as far as I understand means that frontend needs to use the GET endpoint to fetch the data element, witch initialises row ids.

src/Altinn.App.Api/Controllers/ActionsController.cs
tjololo: Yeah, sorry my bad! Comment and code is really misleading here, the key is elementId and always was so this is just a renaming luckily!

olemartinorg:
> What API gets called in that case? As far as I'm able to read all code paths leads to an api that returns a url or a `DataElement`, which as far as I understand means that frontend needs to use the GET endpoint to fetch the data element, witch initialises row ids.

Strange, it's just called via the normal GET endpoint. No row IDs are added. You can test this yourself in the `ttd/frontend-test` app, with the `unique-row-ids` branch (probably best to run app-frontend locally with the `feat/row-uuids` branch as well). When starting the app, you can click the custom button to skip to a task, and then skip to the Likert task. It gives me this data model on the GET endpoint:

```json
{
  "Questions": [
    {
      "Id": "question-1",
      "Answer": ""
    },
    {
      "Id": "question-2",
      "Answer": ""
    },
    {
      "Id": "question-3",
      "Answer": ""
    },
    {
      "Id": "question-4",
      "Answer": ""
    },
    {
      "Id": "question-5",
      "Answer": ""
    },
    {
      "Id": "question-6",
      "Answer": ""
    }
  ]
}
```

ivarne:
Did you add the `AltinnRowId` guid, manually on the row element C# class, or did you run the update tool I have not written yet?

olemartinorg:
Yup, I added it manually (did not use the tool that doesn't exist! :sweat_smile:) It all works if I manually add row IDs in the `IInstantiationProcessor` where this data is first created, but I expect that not to be needed.

olemartinorg:
Btw, it also seems that we get row IDs back when fetching the data models after the instance has been submitted. We have automatic tests that compare those models to a ground truth, and they currently get the data models with row IDs. 

We discussed using a special header or query parameter when fetching the model from app-frontend so that the row IDs do not show up for other users. Let me know what you think, so I can add such a header/query parameter.

ivarne:
I think I prefer an optional GET parameter, but need some help on naming. `includeRowId`, `includeMetadata`, `includeInternalFields`, `frontendMetadata`?

olemartinorg:
I think I prefer `includeRowId` if we're going for the explicit `altinnRowId` field, otherwise probably `frontendMetadata` if we're settling on a metadata object.

ivarne:
~~Æsj, parameteren ble hetende initializeAltinnRowId, for jeg klarte ikke finne ut hvor du hadde kommet med forslaget (trodde det var slack).~~ **Fixet nå**

Uansett har jeg nå lagt til initialisering av AltinnRowId en masse nye steder, men det var bare GET data der det ga mening å ha en parameter for å velge.

Håper å få fikset mer tester etter hvert.

olemartinorg:
I ran our full e2e test-suite for `ttd/frontend-test` before the last commit (along with the changes in app-frontend I'm working on), and everything passed! :partying_face:

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -289,7 +291,7 @@ private static bool FileValidationEnabledForDataType(DataType dataTypeFromMetada
                 }
                 else if (dataType.AppLogic?.ClassRef is not null)
                 {
-                    return await GetFormData(org, app, instanceOwnerPartyId, instanceGuid, dataGuid, dataType, language, instance);
+                    return await GetFormData(org, app, instanceOwnerPartyId, instanceGuid, instance, dataGuid, dataElement, dataType, includeRowId, language);
```
RonnyB71, line 294: Minor, but do we need both instance and instanceGuid? Since instance is introduced, perhaps just remove instanceGuid and get the id from instance object.

ivarne: It is just that `Guid.Parse(instance.Id.Split("/")[1])` gets a little repetitive.

src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -844,6 +844,8 @@ private async Task CopyDataFromSourceInstance(ApplicationMetadata application, I
 
                     await _instantiationProcessor.DataCreation(targetInstance, data, null);
 
+                    ObjectUtils.InitializeAltinnRowId(data);
```
RonnyB71, line 847: Candidate method for moving to a future DataService implementation instead of Utils class, but that should also include refactoring  this controller to not use the IDataClient directly but rather use the DataService and perhaps refactor even more logic out from the controller. But not critical now I think.

src/Altinn.App.Core/Helpers/ObjectUtils.cs
```diff
@@ -10,13 +10,25 @@ public static class ObjectUtils
...
     /// <param name="model">The object to mutate</param>
-    public static void InitializeListsAndNullEmptyStrings(object model)
+    public static void InitializeAltinnRowId(object model)
     {
+        ArgumentNullException.ThrowIfNull(model);
+
         foreach (var prop in model.GetType().GetProperties())
         {
-            if (prop.PropertyType.IsGenericType && prop.PropertyType.GetGenericTypeDefinition() == typeof(List<>))
+            if (prop.PropertyType == typeof(Guid) && prop.Name == "AltinnRowId")
```
RonnyB71: Perhaps a simple refactor of this to a `PropertyIsAltinnRowGuid(PropertyInfo prop)` just remove the duplicated logic in the `RemoveAltinnRowIds `method and keep the string name in one place.

```
    private static bool PropertyIsAltinnRowGuid(System.Reflection.PropertyInfo prop)
    {
        return prop.PropertyType == typeof(Guid) && prop.Name == "AltinnRowId";
    }
```

src/Altinn.App.Core/Implementation/DefaultTaskEvents.cs
```diff
@@ -160,20 +160,77 @@ public async Task OnEndProcessTask(string endEvent, Instance instance)
...
+        // Download the data
+        Type modelType = _appModel.GetModelType(dataType.AppLogic.ClassRef);
+        var instanceGuid = Guid.Parse(instance.Id.Split("/")[1]);
+        var dataGuid = Guid.Parse(dataElement.Id);
+        string app = instance.AppId.Split("/")[1];
+        int instanceOwnerPartyId = int.Parse(instance.InstanceOwner.PartyId);
+        object data = await _dataClient.GetFormData(
+            instanceGuid, modelType, instance.Org, app, instanceOwnerPartyId, dataGuid);
+
+        // Remove hidden data before validation, ignore hidden rows. TODO: Determine how hidden rows should be handled going forward.
```
RonnyB71: Is this TODO still relevant?

src/Altinn.App.Core/Implementation/DefaultTaskEvents.cs
```diff
@@ -160,20 +160,77 @@ public async Task OnEndProcessTask(string endEvent, Instance instance)
...
+                await _dataClient.InsertFormData(updatedData, instanceGuid, saveToModelType ?? modelType, instance.Org, app, instanceOwnerPartyId, saveToDataType.Id);
+            }
+            else
+            {
+                // Remove the shadow fields from the data
+                data = JsonSerializer.Deserialize(serializedData, modelType)!;
+            }
         }
+        // remove AltinnRowIds
+        ObjectUtils.RemoveAltinnRowId(data);
```
RonnyB71, line 230: Should we always remove the row id upon completion?

RonnyB71 (approved):
🕹️ approved we a few minor comments/questions.

src/Altinn.App.Core/Implementation/DefaultTaskEvents.cs
olemartinorg: Yup, I believe so. They're strictly only needed for us, internally when filling out a form. If app developers have other needs - such as requiring that rows are unique in some way, where no rows may contain the same data as any other row - that is never handled by our row ID anyway, and would have to be implemented using validation messages.

EDIT: Also, importantly, as long as the row ID is not part of the XSD or JsonSchema, the data would not be valid when containing row IDs.

src/Altinn.App.Api/Controllers/InstancesController.cs
ivarne: Yes, out of scope now, but would be great if we had test coverage to ensure that the future rewrite doesn't break the behavior.

src/Altinn.App.Core/Implementation/DefaultTaskEvents.cs
ivarne: No, this one has been fixed.

ivarne:
```suggestion
        // Remove hidden data before validation, ignore hidden rows.
```

src/Altinn.App.Core/Helpers/ObjectUtils.cs
ivarne: Done

PR #454, opened 2024-02-23
Fix expression validation crash

bjosttveit:
## Description

There was a small error where the dataElementId was passed to the layoutEvaluatorStateInitializer instead of the layoutSetId, which would cause a 500 internal server error on validation if expression validation was enabled.

I also updated the frontend error severity to the new values used in v4 (which is required for expression validation to work both places simultaneously).

I also moved some shared tests into separate folders. Since frontend no longer cares about `hidden` when evaluating validations, the tests for that have been removed in frontend, but these are still needed for backend.

I also removed the original `ExpressionValidationTests` and only kept the new `ExpressionValidatorTests`, as they both test the exact same things in different ways.

## Verification
- [x] All tests run green


ivarne (approved):
Looks good!

> I also removed the original ExpressionValidationTests and only kept the new ExpressionValidatorTests, as they both test the exact same things in different ways.

Probably for the best. I just tried to leave as many tests intact when I reorganised the server side validations.

The test coverage is somewhat low, and in ExpressionValidator.cs there are a few error branches that looks to be reachable in a valid app, so it would be great to have test cases that cover them.

PR #455, opened 2024-02-23
Add pipeline job to verify clean dotnet format

ivarne:
I think it would be good if we can run dotnet format on the whole project without getting lots of unrelated changes. There might be some requirement for tweaks to editorconfig to make this reasonable.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


tjololo:
Nice! Ensuring formar is a really good thing.
Not sure if the codeQL is the right workflow though?
Would rather place this in the test and analyze or create a new workflow if this needs to run on a schedule

ivarne:
@tjololo Updated and created a separate job. Seems like GitHub actions is much quicker to clone repos, so having multiple jobs isn't the big issue I thought it might be.

Would be nice to have this before we start getting even more devs working in parallel.

bjorntore (approved):

PR #462, opened 2024-02-27
Fix expression validation (hidden & rowIndices)

bjosttveit:
- Attempt number 2 at hidden data for expresion validation 😅 This time it does not remove the hidden data, but only gets the fields that would be removed and ignores those. It was necessary to treat `hiddenRow` slightly differently, just by not ignoring the children if the row was already hidden so it gets full paths.

- This should also fix the problem I mentioned in slack where it did not have row indices for context. Instead of trying to connect expression validations to layout nodes, it uses a more simple context than the component context, that only contains row indices. This means it cannot refer directly to components, but that should not really limit what you can do anyway. I'll make sure to document that.

Added some more tests to cover the cases with direct datamodel references, updated the unit tests to not remove hidden data before validating. Also tested manually in `expression-validation-test` app, so the cypress tests in app-frontend should pass.

## Related Issue(s)
- #{issue number}

## Verification
- [x] All tests run green

## Documentation


ivarne:
I'm not sure about the inheritance for context. Isn't it much easier to read and write the code when the properties are nullable instead?

src/Altinn.App.Core/Features/Validation/Default/ExpressionValidator.cs
```diff
@@ -58,11 +60,12 @@ public async Task<List<ValidationIssue>> ValidateFormData(Instance instance, Dat
         var validationConfig = JsonDocument.Parse(rawValidationConfig).RootElement;
         var layoutSet = _appResourceService.GetLayoutSetForTask(instance.Process.CurrentTask.ElementId);
```
ivarne: I'm not sure this is correct (though it might work fine)? This makes validation dependent on the current task, not the task the data type is assigned to.

In Requried validator I do it differently, but it should at least be the same?
https://github.com/Altinn/app-lib-dotnet/blob/bf2765abc35f5129709980ccc75fea44823851b4/src/Altinn.App.Core/Features/Validation/Default/RequiredValidator.cs#L48

src/Altinn.App.Core/Features/Validation/Default/ExpressionValidator.cs
```diff
@@ -58,11 +60,12 @@ public async Task<List<ValidationIssue>> ValidateFormData(Instance instance, Dat
...
         var evaluatorState = await _layoutEvaluatorStateInitializer.Init(instance, data, layoutSet?.Id);
-        return Validate(validationConfig, evaluatorState, _logger);
+        var hiddenFields = LayoutEvaluator.GetHiddenFieldsForRemoval(evaluatorState, true);
+        return Validate(validationConfig, evaluatorState, hiddenFields, _logger);
     }
 
 
-    internal static List<ValidationIssue> Validate(JsonElement validationConfig, LayoutEvaluatorState evaluatorState, ILogger logger)
+    internal static List<ValidationIssue> Validate(JsonElement validationConfig, LayoutEvaluatorState evaluatorState, List<string> hiddenFields, ILogger logger)
```
ivarne: I think it makes more sense to calculate `hiddenFields` inside the internal method, but after you deleted the tests that used the internal method the proper rewrite is probably to inline the internal method.

src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
```diff
@@ -125,7 +125,7 @@ public static bool EvaluateBooleanExpression(LayoutEvaluatorState state, Compone
             throw new NotImplementedException("Component lookup for components in groups not implemented");
         }
 
-        if (!targetContext.Component.DataModelBindings.TryGetValue("simpleBinding", out var binding))
+        if (targetContext.Component is not BaseComponent || !targetContext.Component.DataModelBindings.TryGetValue("simpleBinding", out var binding))
```
ivarne: Jeg ville gått for `is not null`, siden jeg leser dette som et hint om at den kan være noe annet en BaseComponent.

Ellers er det vel mulig å skrive
```suggestion
        if (targetContext.Component?.DataModelBindings.TryGetValue("simpleBinding", out var binding) != true)
```
siden `..?.DataModelBindings.TryGetValue` kan være `null`, `true` eller `false` og men alt annet enn true skal trigge feilmelding.

ivarne (approved):
Ser veldig bra ut. Noen små kommentarer du kan fikse hvis du har tid, være uenig, eller bare ignorere.

ivarne:
Ser forresten at sonar cloud har funnet to steder der vi får warning fordi `context.Component.DataModelBindings` kan trigge null pointer exception om Component er null. Det bør vi vel også fikse?

src/Altinn.App.Core/Features/Validation/Default/ExpressionValidator.cs
bjosttveit: Agree

src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
bjosttveit: Jepp det var faktisk det jeg gjorde først, kan ta det tilbake. Funker like bra det

ivarne (approved):

PR #463, opened 2024-02-28
Let validation service remove all hidden data before validation.

ivarne:
Implements the same solution as in https://github.com/Altinn/app-lib-dotnet/issues/1157, but in the backend.

This makes a few extra in-memory copies of the data, but at least it is centralised once in the validation service, instead of doing it in each validator.

* Tests fails
* Currently not opt-in

## Related Issue(s)
- Altinn/app-frontend-react#461
- https://github.com/Altinn/app-lib-dotnet/pull/462

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Validation/ValidationService.cs
+++ b/src/Altinn.App.Core/Internal/Validation/ValidationService.cs
@@ -1,7 +1,9 @@
-using Altinn.App.Core.Features.Validation;
+using System.Text.Json;
+using Altinn.App.Core.Helpers;
 using Altinn.App.Core.Internal.App;
 using Altinn.App.Core.Internal.AppModel;
 using Altinn.App.Core.Internal.Data;
+using Altinn.App.Core.Internal.Expressions;
 using Altinn.App.Core.Models.Validation;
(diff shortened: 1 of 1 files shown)


olemartinorg:
> Implements the same solution as in Altinn/app-lib-dotnet#1157, but in the backend.
> Currently not opt-in

It's unclear to me, the mentioned issue (1709) is about removing hidden data, not only for validation purposes, but everywhere. If this PR enables the feature when actually saving data (removing hidden data from the data moden in Storage) that's not the way to go, I think.

If it's about removing hidden data in the data model copy only when running validation, so that `["component", "..."]` lookups return `null` for hidden components, the way it used to be, that should be fine.

ivarne:
> It's unclear to me...

Functionality the same, but obviously not including the cache of removed data in the browser. It removes from a copy used only for validation.

Importantly that also includes `["dataModel",...]`

ivarne:
Closing this, as it was decided to wait for https://github.com/Altinn/app-lib-dotnet/issues/1157 to introduce this form of automatic filtration of data.

PR #465, opened 2024-02-28
Move userId check into the signing action

tjololo:
## Description
Checking if userId is null in processengine seems to be wrong. Made the field nullable and moved the check into the SigningAction.
I'm not happy with the magic string for code, so suggestions are welcome

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


ivarne:
I'd just add a new enum field to `UserActionResult` to store the system readable information. Why not just use the `ProcessErrorType` we already have?

ivarne:
PS: I also noticed that `ProcessChangeResult.ErrorType` is declared nullable, instead of just adding `NoError` as the 0th value in the enum.

I don't know what is the most recognised pattern, but it kind of makes sense to not carry the nullable wrapper everywhere.

ivarne:
Another idea!
Can we remove the `userId` from the UserActionRequest altogether and just fetch it from the cookie in the `SigningAction` instead? It seems like something that would only be required there, and the signing is very dependent on the cookie (and in the future possible other cryptographic data).

tjololo:
> I'd just add a new enum field to `UserActionResult` to store the system readable information. Why not just use the `ProcessErrorType` we already have?

Sounds like a good solution!

> PS: I also noticed that `ProcessChangeResult.ErrorType` is declared nullable, instead of just adding `NoError` as the 0th value in the enum.
> 
> I don't know what is the most recognised pattern, but it kind of makes sense to not carry the nullable wrapper everywhere.

I'm seeing that the whole ProcessChangeResult class might be replaced by `ServiceResult<ProcessStateChange, ProcessError>` where `ProcessError` is a new object holding `ErrorMessage` and `ProcessErrorType`. Then we don't need to have it nullable as that is the use case for `ServiceResult`. So we can make a change after that is merged.



> Another idea! Can we remove the `userId` from the UserActionRequest altogether and just fetch it from the cookie in the `SigningAction` instead? It seems like something that would only be required there, and the signing is very dependent on the cookie (and in the future possible other cryptographic data).

I think we should leave it to the Api project to work on the httpContext/cookies and just require that core receives the objects it needs from those from the Api project. IMO the core project ideally shouldn't need to be changed if we remove the cookie or change protocol down the road.

ivarne:
> I think we should leave it to the Api project to work on the httpContext/cookies and just require that core receives the objects it needs from those from the Api project. IMO the core project ideally shouldn't need to be changed if we remove the cookie or change protocol down the road.

Yes, that change would include moving the signature functionality to Api.

tjololo:
I was think that I should as mentioned rewrite the ProcessEnginge to use the new ServiceResult. But decided that since we have a rewrite of the process engine in another commit that would just create conflicts.
I suggest we just create an issue for it and the refactor that after those changes are merged.

ivarne (approved):

PR #477, opened 2024-03-05
Add unit test that stores a copy of swagger.json|yaml

ivarne:
This makes it more obvious when a change affects the swagger document

I need some input on how to actually write the test. Currently it does not have assertions, but depends on a user checking the git diff.
* Should it fail if the previous run was different, but automatically store the change?
* Should it fail and only update the file if a specific Environment variable is set?


bjorntore (approved):

bjorntore:
> This makes it more obvious when a change affects the swagger document
> 
> I need some input on how to actually write the test. Currently it does not have assertions, but depends on a user checking the git diff.
> 
>     * Should it fail if the previous run was different, but automatically store the change?
> 
>     * Should it fail and only update the file if a specific Environment variable is set?

I think I'm OK with starting with this version, where the user will see the change in git and determine if the change was intentional.

PR #478, opened 2024-03-06
Make prefill nullable in IProcessTaskStart interface so we don't have…

bjorntore:
Quick house keeping change for V8 release, if allowed.


tjololo (approved):
:godmode:

PR #497, opened 2024-03-11
Fix workflows

ivarne:
I noticed that we run lots of pull request actions for renovate PRs when we merge to main, so this will disable a few of the less likely actions to fail for renovate.

* Don't run code analyzis on renovate pull requests (they trigger on every merge to master)
* Use list syntax for branches
* Don't refernece the v8 branch

## Verification
- [ ] All tests run green

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -19,8 +19,6 @@ jobs:
       actions: read
       contents: read
       security-events: write
-    if: |
-      github.event.pull_request.user.login != 'renovate[bot]'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -2,10 +2,12 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "main", "v8" ]
+    branches: 
+      - "main"
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [ "main", "v8" ]
+    branches: 
+      - "main"
   schedule:
     - cron: '37 20 * * 3'
 
@@ -17,7 +19,8 @@ jobs:
       actions: read
       contents: read
       security-events: write
-
+    if: |
+      github.event.pull_request.user.login != 'renovate'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -20,7 +20,7 @@ jobs:
       contents: read
       security-events: write
     if: |
-      github.event.pull_request.user.login != 'renovate'
+      github.event.pull_request.user.login != 'renovate[bot]'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/dotnet-format.yml
+++ b/.github/workflows/dotnet-format.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   verify-no-changes:
     if: |
-      github.event.pull_request.user.login != 'renovate'
+      github.event.pull_request.user.login != 'renovate[bot]'
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/dotnet-format.yml
+++ b/.github/workflows/dotnet-format.yml
@@ -2,11 +2,13 @@ name: Verify dotnet format
 
 on:
   pull_request:
-    branches: [ "main", "v8" ]
+    branches: 
+      - "main"
 
 jobs:
-  build:
-
+  verify-no-changes:
+    if: |
+      github.event.pull_request.user.login != 'renovate'
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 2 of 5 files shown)


.github/workflows/codeql.yml
```diff
@@ -2,10 +2,12 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "main", "v8" ]
+    branches: 
+      - "main"
```
tjololo, line 6:
```suggestion
      - "main"
      - "v7"
```

.github/workflows/codeql.yml
```diff
@@ -2,10 +2,12 @@ name: "CodeQL"
...
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [ "main", "v8" ]
+    branches: 
+      - "main"
```
tjololo, line 10:
```suggestion
      - "main"
      - "v7"
```

.github/workflows/dotnet-format.yml
```diff
@@ -2,11 +2,13 @@ name: Verify dotnet format
 
 on:
   pull_request:
-    branches: [ "main", "v8" ]
+    branches: 
+      - "main"
```
tjololo, line 6:
```suggestion
      - "main"
      - "v7"
```

.github/workflows/dotnet-test.yml
```diff
@@ -1,9 +1,11 @@
 name: Build and Test on windows, macos and ubuntu
 on:
   push:
-    branches: [ main, v8 ]
+    branches:
+      - "main"
```
tjololo, line 5:
```suggestion
      - "main"
      - "v7"
```

.github/workflows/dotnet-test.yml
```diff
@@ -1,9 +1,11 @@
...
   pull_request:
-    branches: [ main, v8 ]
+    branches:
+      - "main"
```
tjololo, line 8:
```suggestion
      - "main"
      - "v7"
```

.github/workflows/test-and-analyze-fork.yml
```diff
@@ -1,7 +1,8 @@
 name: Code test and analysis (fork)
 on:
   pull_request:
-    branches: [ main, v8 ]
+    branches:
+      - "main"
```
tjololo, line 5:
```suggestion
      - "main"
      - "v7"
```

.github/workflows/test-and-analyze.yml
```diff
@@ -1,14 +1,17 @@
 name: Code test and analysis
 on:
   push:
-    branches: [ main, v8 ]
+    branches:
+      - "main"
```
tjololo, line 5:
```suggestion
      - "main"
      - "v7"
```

.github/workflows/test-and-analyze.yml
```diff
@@ -1,14 +1,17 @@
...
   pull_request:
-    branches: [ main, v8 ]
+    branches:
+      - "main"
```
tjololo, line 8:
```suggestion
      - "main"
      - "v7"
```

.github/workflows/codeql.yml
```diff
@@ -17,7 +19,8 @@ jobs:
       actions: read
       contents: read
       security-events: write
-
+    if: |
+      github.event.pull_request.user.login != 'renovate[bot]'
```
tjololo: Could we keep codeQL? there is no reason to why it should fail, but it doesn't hurt to run either IMO

tjololo (changes requested):
Lets add v7 instead of v8.

ivarne:
@tjololo Adding `"v7"` in the `"main"` branch does nothing (it needs to be on the `v7` branch to take effect), so I'm not sure it makes sense to even add them here. 

With regards to CodeQL I have no opinion, but the main goal was to reduce the build storm so I just took everything that wasn't the simple `dotnet test`. As we don't have a warning free main branch (so we can run warning as error) it might make sense to include one of the analysis tools that might detect that an upgrade caused new warnings.

tjololo:
> @tjololo Adding `"v7"` in the `"main"` branch does nothing (it needs to be on the `v7` branch to take effect), so I'm not sure it makes sense to even add them here.
> 
> With regards to CodeQL I have no opinion, but the main goal was to reduce the build storm so I just took everything that wasn't the simple `dotnet test`. As we don't have a warning free main branch (so we can run warning as error) it might make sense to include one of the analysis tools that might detect that an upgrade caused new warnings.

Sorry, you are correct, adding v7 here solves nothing. I will remove the suggestions

.github/workflows/codeql.yml
tjololo:
```suggestion
```

tjololo (approved):
I suggest keeping one analysis workflow. As codeQL generates no comments if everything is OK I prefer that over sonar

ivarne:
I'll have another look at this tomorrow. I just realized that branch restrictions for pull requests should be removed altogether. It's a shitty heuristic to think that a PR to the payment branch is less likely to benefit from analysis than one to main. Branch PR are so rare anyway.

PR #503, opened 2024-03-13
Rewrite of maskinporten usage from eFormidling

RonnyB71:
## Description
This change includes using JWK as the authentication mechanism against Maskinporten instead of a X509 certificate when using eFormidling integration.

## Related Issue(s)
- #{issue number}

## Verification
- [x] All tests run green

## Documentation


RonnyB71:
The last remaining % of test coverage doesn't seem possible, or worth it, to achieve. I ran into issues with X509 certificate being impossible to mock.

test/Altinn.App.Core.Tests/Altinn.App.Core.Tests.csproj
```diff
@@ -97,6 +97,8 @@
 
   <ItemGroup>
     <Watch Remove="Extensions\DictionaryExtensionsTests.cs" />
+    <Watch Remove="Internal\Maskinporten\MaskinportenExtensionsTests.cs" />
+    <Watch Remove="Internal\Maskinporten\MaskinportenJwkTokenProviderTests.cs" />
```
ivarne: Why exclude from dotnet watch?

src/Altinn.App.Core/EFormidling/Implementation/EformidlingStatusCheckEventHandler.cs
```diff
@@ -29,15 +30,45 @@ public class EformidlingStatusCheckEventHandler : IEventHandler
...
+            _httpClientFactory = httpClientFactory;
+            _maskinportenTokenProvider = maskinportenTokenProvider;
+            _platformSettings = platformSettings.Value;
+            _generalSettings = generalSettings.Value;
+        }
+
         /// <summary>
         /// Initializes a new instance of the <see cref="EformidlingStatusCheckEventHandler"/> class.
         /// </summary>
+        [Obsolete("Use the constructor using IMaskinportenTokenProvider instead. This constructor will be removed when V9 is released.")]
```
ivarne: I don't think it is required to have this compatibility mode. I can't find any usages of this class on gitea prod (only a few cases where it has been replaced with a comment that the standard one is useless, and nobody would call the constructor directly anyway, but rely on DI)

Another issue is that C# DI has to pick a constructor to call, [and it picks the one with the most parameters](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#multiple-constructor-discovery-rules), witch happens to be the old one. Your DI test likely works because you don't have everything required registered for the old constructor

Here are the hits I found on the main branch of every repo.

```
apps/main/digdir/tilskudd-dig-delt-komp/App/Program.cs:
  29      services.AddEFormidlingServices<EFormidlingMetadata, EFormidlingReceivers>(config);
  30:     services.AddTransient<IEventHandler, EFormidlingStatusCheckEventHandler>();
  31  

apps/main/digdir/tilskudd-dig-delt-komp/App/logic/EFormidling/EFormidlingStatusCheckEventHandler.cs:
  26  
  27: public class EFormidlingStatusCheckEventHandler : IEventHandler
  28  {
  29      private readonly IEFormidlingClient _eFormidlingClient;
  30:     private readonly ILogger<EFormidlingStatusCheckEventHandler> _logger;
  31      private readonly IHttpClientFactory _httpClientFactory;

  39      /// <summary>
  40:     /// Initializes a new instance of the <see cref="EFormidlingStatusCheckEventHandler"/> class.
  41      /// </summary>
  42:     public EFormidlingStatusCheckEventHandler(
  43          IEFormidlingClient eFormidlingClient,
  44          IHttpClientFactory httpClientFactory,
  45:         ILogger<EFormidlingStatusCheckEventHandler> logger,
  46          IMaskinportenService maskinportenService,

apps/main/digdir/tilskudd-dig-delt-komp/App/logic/EFormidling/ServiceCollectionExtensions.cs:
  21          // This is the local implementation. Note the uppercase F in EFormidling as opposed to the lowercase in the default implementation
  22:         services.AddTransient<IEventHandler, EFormidlingStatusCheckEventHandler>();
  23          services.AddHostedService<EformidlingStartup>();

apps/main/dihe/redusert-foreldrebetaling-bhg/App/logic/Eformidling/CustomEformidlingStatusCheckEventHandler.cs:
  39      //     OBS:
  40:     //          Due to a flaw in the default EformidlingStatusCheckEventHandler from App.Core, it only supports authentication through certificates. As this app do not use certificates, this handler is added as a re-implementation of the default version.
  41      //          When Altinn adds support for default eFormidling with jwk, this file can be removed

  63          // Summary:
  64:         //     Initializes a new instance of the Altinn.App.Core.EFormidling.Implementation.EformidlingStatusCheckEventHandler
  65          //     class.

apps/main/dihe/sjenkebevilling-2023/App/Program.cs:
  116  
  117:     // Use of custom EformidlingStatusCheckEventHandler requires the different services of eFormidling to be added individually 		
  118      services.AddTransient(typeof(IEFormidlingReceivers), typeof(EFormidlingReceivers));		

apps/main/dihe/sjenkebevilling-2023/App/logic/Eformidling/CustomEformidlingStatusCheckEventHandler.cs:
  36      //     OBS:
  37:     //          Due to a flaw in the default EformidlingStatusCheckEventHandler from App.Core, it only supports authentication through certificates. As this app do not use certificates, this handler is added as a re-implementation of the default version.
  38      //          When Altinn adds support for default eFormidling with jwk, this file can be removed

  60          // Summary:
  61:         //     Initializes a new instance of the Altinn.App.Core.EFormidling.Implementation.EformidlingStatusCheckEventHandler
  62          //     class.

apps/main/dihe/soknad-elevpermisjon/App/Program.cs:
  118  
  119:     // Use of custom EformidlingStatusCheckEventHandler requires the different services of eFormidling to be added individually 
  120      services.AddTransient(typeof(IEFormidlingReceivers), typeof(EFormidlingReceivers));

apps/main/dihe/soknad-elevpermisjon/App/logic/Eformidling/CustomEformidlingStatusCheckEventHandler.cs:
  39      //     OBS:
  40:     //          Due to a flaw in the default EformidlingStatusCheckEventHandler from App.Core, it only supports authentication through certificates. As this app do not use certificates, this handler is added as a re-implementation of the default version.
  41      //          When Altinn adds support for default eFormidling with jwk, this file can be removed

  63          // Summary:
  64:         //     Initializes a new instance of the Altinn.App.Core.EFormidling.Implementation.EformidlingStatusCheckEventHandler
  65          //     class.

```

ivarne (commented):
I'm not sure I understand your compatibility plan here. The `[Obsolete]` annotations won't be seen by app developers as far as I can tell.

Shouldn't there be two implementations of `IMaskinportenTokenProvider`, one that depends on the `IX509CertificateProvider` and a totally separate one that uses the JWK approach? With different versions of `services.AddEFormidlingServices` to ensure that the correct one is registered. The previous one can be marked as obsolete and to be removed in `v9`

bjorntore (approved):
Looks good to me in general, but let's conclude Ivar's comments before moving on.

test/Altinn.App.Core.Tests/Altinn.App.Core.Tests.csproj
RonnyB71: Not sure why those where added, I'll just remove them.

src/Altinn.App.Core/EFormidling/Implementation/EformidlingStatusCheckEventHandler.cs
RonnyB71: You are right that this compatibility relies on either the Jwk or X509 provider being registered, but not both. If the both are it will use the X509 as this has the most parameters. This was what I mentioned in the description of this PR.

But I agree, I think I overcomplicated this by trying to keep the same EformidlingStatusCheckEventHandler class and could only have created a new version of it using the IMaskinportenTokenProvider.

PR #511, opened 2024-03-14
Feat 494/email notification

### LINKED ISSUE #494: Implement e-mail support for app developers
### Description

As an application developer I would like to have support for sending e-mails from an Altinn 3 app in order to support e-mail notifications as part of the process. 

### Additional Information

Team Core have implemented a [Notification service](https://github.com/Altinn/altinn-notifications) with a endpoint for sending e-mails.

See [related issue for SMS](https://github.com/Altinn/app-lib-dotnet/issues/493)

### Tasks

- [x] Implement a client to talk to the e-mail api

### Acceptance Criterias

- [x] How to use from within a application is documented
- [x] The logic is made available to the application developers as a injectable service

HauklandJ:
## Description
Add client for email, using Altinn.Notifications API ("notifications/api/v1/orders/email"). 

## Related Issue(s)
- #494 

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Core/Infrastructure/Clients/Email/EmailNotificationClient.cs
```diff
@@ -0,0 +1,126 @@
...
+                Content = new StringContent(body, new MediaTypeHeaderValue("application/json")),
+            };
+            httpRequestMessage.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+            httpRequestMessage.Headers.Add(
+                "PlatformAccessToken",
+                _accessTokenGenerator.GenerateAccessToken(application.Org, application.AppIdentifier.App)
+            );
+
+            httpResponseMessage = await httpClient.SendAsync(httpRequestMessage, ct);
+            httpContent = await httpResponseMessage.Content.ReadAsStringAsync(ct);
```
RonnyB71, line 80: Reading the content twice? Same as line 66.

src/Altinn.App.Core/Infrastructure/Clients/Email/EmailNotificationClient.cs
```diff
@@ -0,0 +1,126 @@
...
+
+                orderResponse = JsonSerializer.Deserialize<EmailOrderResponse>(httpContent);
+                if (orderResponse is null)
+                    throw new InvalidOperationException("Couldn't deserialize email notification order response.");
+
+                _orderCount.WithLabels("success").Inc();
+            }
+            else
+            {
+                throw new HttpRequestException("Got error status code for email notification order");
```
RonnyB71, line 91: Should this also increment the _orderCount.WithLabels("error")?

HauklandJ: The error thrown here is caught by the generic catch clause - which then increments the label

src/Altinn.App.Core/Infrastructure/Clients/Email/EmailNotificationClient.cs
```diff
@@ -0,0 +1,124 @@
...
+/// </summary>
+public sealed class EmailNotificationClient : IEmailNotificationClient
+{
+    private readonly IHttpClientFactory _httpClientFactory;
+    private readonly IAppMetadata _appMetadata;
+    private readonly PlatformSettings _platformSettings;
+    private readonly IAccessTokenGenerator _accessTokenGenerator;
+    private readonly TelemetryClient _telemetryClient;
+    private static readonly Counter _orderCount = Metrics
+        .CreateCounter("altinn_app_notification_order_request_count", "Number of notification order requests.", labelNames: ["email"]);
```
martinothamar:
```suggestion
        .CreateCounter("altinn_app_notification_order_request_count", "Number of notification order requests.", labelNames: ["type", "result"]);
```
Since we want the same label name for both types of notifications

src/Altinn.App.Core/Infrastructure/Clients/Email/EmailNotificationClient.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            httpResponseMessage = await httpClient.SendAsync(httpRequestMessage, ct);
+            httpContent = await httpResponseMessage.Content.ReadAsStringAsync(ct);
+            if (httpResponseMessage.IsSuccessStatusCode)
+            {
+                orderResponse = JsonSerializer.Deserialize<EmailOrderResponse>(httpContent);
+                if (orderResponse is null)
+                    throw new InvalidOperationException("Couldn't deserialize email notification order response.");
+
+                _orderCount.WithLabels("success").Inc();
```
martinothamar:
```suggestion
                _orderCount.WithLabels("email", "success").Inc();
```

martinothamar (changes requested):
Small adjustment to metric labels

martinothamar:
Something we can also consider, is to actually handle and return an error object for bad requests (400)
image

martinothamar (approved):
LGTM

test/Altinn.App.Core.Tests/Infrastructure/Clients/Email/EmailNotificationClientTests.cs
```diff
@@ -0,0 +1,171 @@
...
+        httpClientFactoryMock.Setup(h => h.CreateClient(It.IsAny<string>())).Returns(httpClient);
+
+        var emailNotificationClient = CreateEmailNotificationClient(httpClientFactoryMock);
+        var recipients = new List<EmailRecipient>()
+        {
+            new("test.testesen@testdirektoratet.no")
+        };
+
+        var emailNotification = new EmailNotification("subject", "body", recipients, "testref");
+        var ct = new CancellationTokenSource().Token;
```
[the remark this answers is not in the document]
ivarne: I tend to use `CancellationToken.None` instead of creating a new token that will never trigger. Any reason that won't work here?

src/Altinn.App.Core/Models/Email/EmailNotification.cs
```diff
@@ -0,0 +1,60 @@
...
+        string sendersReference,
+        string contentType = "Plain",
+        DateTime? requestedSendTime = null)
+    {
+        Subject = subject;
+        Body = body;
+        Recipients = emailRecipients;
+        SendersReference = sendersReference;
+        ContentType = contentType;
+        RequestedSendTime = requestedSendTime is null ? DateTime.Now : (DateTime)requestedSendTime;
```
ivarne: Do we need to think about timezones here? I would think it makes more sense to send `UTC` times to the API?

Also, we don't have many constructors for data objects, and now that dotnet 8 supports `required` I don't see any reason why we should start adding of them. 

I also tend to favour explicitly annotating with `[JsonPropertyName()]`, instead of requiring `PropertyNamingPolicy = JsonNamingPolicy.CamelCase` every place it is serialised, in case someone didn't use `PropertyNameCaseInsensitive = true` when deserialising.

src/Altinn.App.Core/Infrastructure/Clients/Email/EmailNotificationClient.cs
```diff
@@ -0,0 +1,125 @@
...
+        IAppMetadata appMetadata,
+        IAccessTokenGenerator accessTokenGenerator,
+        TelemetryClient telemetryClient)
+    {
+        _platformSettings = platformSettings.Value;
+        _httpClientFactory = httpClientFactory;
+        _appMetadata = appMetadata;
+        _accessTokenGenerator = accessTokenGenerator;
+        _telemetryClient = telemetryClient;
+    }
```
ivarne: Constructor at the bottom of the class is OK I guess, but not what I usually see.

src/Altinn.App.Core/Infrastructure/Clients/Email/EmailNotificationClient.cs
```diff
@@ -0,0 +1,125 @@
...
+            return orderResponse;
+        }
+        catch (Exception e)
+        {
+            _orderCount.WithLabels("email", "error").Inc();
+            var ex = new EmailNotificationException("Something went wrong when processing the email order, see inner exception for details.", e);
+            ex.Data.Add("responseContent", httpContent);
+            ex.Data.Add("responseStatusCode", httpResponseMessage?.StatusCode.ToString());
+            ex.Data.Add("responseReasonPhrase", httpResponseMessage?.ReasonPhrase);
+            _telemetryClient.TrackException(ex);
```
ivarne: Is this manual work with `TelemetryClient` really required? In other parts of the code we just inject an `ILogger<ClassName>` and do `_logger.LogError(ex, "Comment about were the exception happened")`, and it seems to be picked up by application insights just fine. For dependencies we rely on the automatic instrumentation of httpClient, that also (to my knowledge) does most of what you seem to attempt (apart from hiding the actual server name with a more pretty `Altinn.Notifications.OrderEmailNotification`). Won't the current implementation show up as two dependencies that are really the same?

test/Altinn.App.Core.Tests/Altinn.App.Core.Tests.csproj
```diff
@@ -97,6 +97,7 @@
 
   <ItemGroup>
     <Watch Remove="Extensions\DictionaryExtensionsTests.cs" />
+    <Watch Remove="Infrastructure\Clients\Email\EmailNotificationClientTests.cs" />
```
ivarne: What are these for?

ivarne (commented):
A few comments/discussion points.

test/Altinn.App.Core.Tests/Infrastructure/Clients/Email/EmailNotificationClientTests.cs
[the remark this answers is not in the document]
HauklandJ: No reason that would not work.

test/Altinn.App.Core.Tests/Altinn.App.Core.Tests.csproj
HauklandJ: 'dotnet watch' will ignore the files listed here

src/Altinn.App.Core/Models/Email/EmailNotification.cs
martinothamar: FWIW I checked the notifications API source, it does ensure whatever is provided is UTC when persisted. Timezone is included when serialized to JSON 
image

I agree though that UTC makes more sense. I also prefer  `DateTimeOffset` but it doesn't seem like that is used much

src/Altinn.App.Core/Infrastructure/Clients/Email/EmailNotificationClient.cs
HauklandJ: I would prefer primary constructor, but I'll move it to under fields for now

test/Altinn.App.Core.Tests/Altinn.App.Core.Tests.csproj
ivarne: Why should it ignore changes to those files?

ivarne: Is there anything that would cause this file to change that should not cause a `dotnet watch test` to run tests again?

src/Altinn.App.Core/Infrastructure/Clients/Email/EmailNotificationClient.cs
ivarne: I liked primary constructors for DI, until I learned that they were not "readonly". Then I discovered that our style guides forbids using the _underscore prefix, which is really confusing inside the methods where I can't tell the difference between DI services and local variables.

test/Altinn.App.Core.Tests/Altinn.App.Core.Tests.csproj
HauklandJ: The only reason I can think of to have it here is to avoid hot reload when running with "dotnet watch" instead of "dotnet run". It was not a change I made manually or with intent so I will remove it.

HauklandJ: I also just realized which csproj this is in... Yeah it makes no sense

src/Altinn.App.Core/Infrastructure/Clients/Email/EmailNotificationClient.cs
HauklandJ: TrackException not necessary as this is done by default, tested to verify. TrackDependency is nice as the class does not only call an http endpoint, it gets applicationdata and fetches a plattform token as well.

PR #525, opened 2024-03-20
SMS notifications client, some refactoring that affects email client

### LINKED ISSUE #493: Implement SMS support for app developers
### Description

As an application developer I would like to have support for sending SMS from an Altinn 3 app in order to support notifications as part of the process.

### Additional Information

Team Core have implemented a [Notification service](https://github.com/Altinn/altinn-notifications) with a endpoint for sending SMS.

See [related issue for e-mail](https://github.com/Altinn/app-lib-dotnet/issues/494)

### Tasks

- [x] Implement a client to talk to the SMS api
- [x] Implement localtest setup to use locally

### Acceptance Criterias

- [x] How to use from within a application is documented
- [x] The logic is made available to the application developers as a injectable service

martinothamar:
Implementing the SMS notification client, and some refactoring

## Description
* Implemented SMS client
* Moved related code into `Notifications/` folder and namespace
* Shared telemetry class to centralize the config
* Made implementation classes internal
* Made injection of `TelemetryClient` optional as it may not exist in the container if no instrumentation key or connection string is configured (which might be the case when running locally)
* Made exception constructors internal and to string building there based on HTTP info
* Additional test and edge case for `RequestedSendTime`. Removed `AddMinutes(1)` from the email notification timestamp as the [Notifications API accounts for that](https://github.com/Altinn/altinn-notifications/blob/7a1b3960ad1c3f9fb2bfe2aded220b644f2dfa20/src/Altinn.Notifications/Validators/SmsNotificationOrderRequestValidator.cs#L29)
* Added `ILogger`-logging for errors
* Move interfaces and implementations to Altinn.App.Core.(Features/Models)

- [x] Implement SMS support in localtest - https://github.com/Altinn/app-localtest/pull/94

## Related Issue(s)
- #493

## Verification
- [x] All tests run green

## Documentation
- [x] https://github.com/Altinn/altinn-studio-docs/pull/1498


src/Altinn.App.Core/Infrastructure/Clients/Notifications/Email/EmailNotificationClient.cs
```diff
@@ -1,58 +1,45 @@
...
-    /// <param name="accessTokenGenerator">An access token generator to create an access token.</param>
-    /// <param name="telemetryClient">Client used to track dependencies.</param>
     public EmailNotificationClient(
+        ILogger<EmailNotificationClient> logger,
         IHttpClientFactory httpClientFactory,
         IOptions<PlatformSettings> platformSettings,
         IAppMetadata appMetadata,
         IAccessTokenGenerator accessTokenGenerator,
-        TelemetryClient telemetryClient)
+        IServiceProvider serviceProvider)
```
tjololo: Would `TelemetryClient? telemetryClient = null` yield the same result? This would in my opinion make the constructor easier to read next time around, and a bit easier to use in tests.

martinothamar: Yup! That works 👍

HauklandJ (approved):
lgtm

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Sms/SmsNotificationClient.cs
```diff
@@ -0,0 +1,97 @@
...
+            );
+
+            httpResponseMessage = await httpClient.SendAsync(httpRequestMessage, ct);
+            httpContent = await httpResponseMessage.Content.ReadAsStringAsync(ct);
+
+            if (httpResponseMessage.IsSuccessStatusCode)
+            {
+                orderResponse = JsonSerializer.Deserialize<SmsNotificationOrderResponse>(httpContent);
+                if (orderResponse is null)
+                    throw new Exception("Couldn't deserialize SMS notification order response");
```
HauklandJ: Generic exception thrown, this was changed in email client after sonarcloud scan complaint

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Telemetry.cs
```diff
@@ -0,0 +1,53 @@
...
+        public readonly void Dispose()
+        {
+            var stopTimestamp = Stopwatch.GetTimestamp();
+            var elapsed = Stopwatch.GetElapsedTime(_startTimestamp, stopTimestamp);
+
+            _telemetryClient?.TrackDependency(
+                "Altinn.Notifications",
+                "OrderNotification",
+                null,
+                new DateTime(_startTimestamp),
```
HauklandJ: new DateTime(_startTimestamp, DateTimeKind.Utc),

martinothamar: What I was doing was apparantly all wrong, pushed a fix...
image

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Sms/SmsNotificationClient.cs
```diff
@@ -0,0 +1,97 @@
...
+using Microsoft.ApplicationInsights;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Altinn.App.Core.Infrastructure.Clients.Notifications.Sms;
+
+internal sealed class SmsNotificationClient : ISmsNotificationClient
+{
+    private readonly ILogger<SmsNotificationClient> _logger;
+    private readonly IHttpClientFactory _httpClientFactory;
```
ivarne: Currently we have one class that uses `_httpClientFactory.CreateClient()` (EformidlingStatusCheckEventHandler) and 21 classes that are registered using `AddHttpClient` and injects a `HttpClient` directly. Any reason you chose the other approach?

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Email/EmailNotificationClient.cs
```diff
@@ -1,58 +1,44 @@
...
         _telemetryClient = telemetryClient;
     }
 
-    /// <inheritdoc/>
-    /// <exception cref="EmailNotificationException"/>
     public async Task<EmailOrderResponse> Order(EmailNotification emailNotification, CancellationToken ct)
     {
-        var startTime = DateTime.UtcNow;
-        var timer = Stopwatch.StartNew();
+        using var dependency = new Telemetry.Dependency(_telemetryClient);
```
ivarne: If the `TelemetryClient` interface from ApplicationInsights isn't good enough for our purposes (I kind of agree and like the automatic stopwatch), we should create our own telemetry service interface and inject an implementation that uses `TelemetryClient` instead.

Implementing `IDisposable` to be able to use the `using` statement is a little too much magic in my opinion to be a helper class that isn't even possible to share with other dependency tracking.

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Sms/SmsNotificationClient.cs
```diff
@@ -0,0 +1,97 @@
...
+            );
+
+            httpResponseMessage = await httpClient.SendAsync(httpRequestMessage, ct);
+            httpContent = await httpResponseMessage.Content.ReadAsStringAsync(ct);
+
+            if (httpResponseMessage.IsSuccessStatusCode)
+            {
+                orderResponse = JsonSerializer.Deserialize<SmsNotificationOrderResponse>(httpContent);
+                if (orderResponse is null)
+                    throw new InvalidOperationException("Couldn't deserialize SMS notification order response");
```
ivarne: I tend to just add `!` to ignore the fact that `Deserialize` might return null, but otherwise I throw the same exception as `Deserialize` would have thrown if they knew the calling context has nullable enabled when specifying a non nullable type to deserialise to.
```suggestion
                orderResponse = JsonSerializer.Deserialize<SmsNotificationOrderResponse>(httpContent)
                    ?? throw new JsonException("sms endpoint returned the string `null`");
```

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Sms/SmsNotificationClient.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            using var httpRequestMessage = new HttpRequestMessage(HttpMethod.Post, uri)
+            {
+                Content = new StringContent(body, new MediaTypeHeaderValue("application/json")),
+            };
+            httpRequestMessage.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+            httpRequestMessage.Headers.Add(
+                "PlatformAccessToken",
+                _accessTokenGenerator.GenerateAccessToken(application.Org, application.AppIdentifier.App)
+            );
```
ivarne: Most other clients use [Altinn.App.Core/Extensions/HttpClientExtension.cs](https://github.com/Altinn/app-lib-dotnet/blob/8bdb34be672b1af2959d6174d0ec80ca05ae1ae7/src/Altinn.App.Core/Extensions/HttpClientExtension.cs)

Longer term we should probably consider providing something that looks like [MIcrosoft.AspNetCore.HeaderPropagation](https://github.com/dotnet/aspnetcore/tree/main/src/Middleware/HeaderPropagation) so we can just do `AddHttpClient<IClient,Client>().AddPlatformToken()`

src/Altinn.App.Core/Internal/Notifications/Email/EmailNotificationException.cs
```diff
@@ -0,0 +1,13 @@
+namespace Altinn.App.Core.Internal.Notifications.Email;
+
+/// <summary>
+/// Class representing an exception throw when an email notification could not be sent.
+/// </summary>
+public sealed class EmailNotificationException : Exception
+{
+    internal EmailNotificationException(string? message, HttpResponseMessage? response, string? content, Exception? innerException)
+    : base($"{message}: StatuCode={response?.StatusCode}\nReason={response?.ReasonPhrase}\nBody={content}\n", innerException)
```
HauklandJ: typo "StatuCode" -> "StatusCode"

src/Altinn.App.Core/Internal/Notifications/Sms/SmsNotificationClientException.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Altinn.App.Core.Internal.Notifications.Sms;
+
+/// <summary>
+/// Class representing an exception thrown when a SMS notificcation order could not be created
+/// </summary>
+public sealed class SmsNotificationException : Exception
+{
+    internal SmsNotificationException(string? message, HttpResponseMessage? response, string? content, Exception? innerException)
+    : base($"{message}: StatuCode={response?.StatusCode}\nReason={response?.ReasonPhrase}\nBody={content}\n", innerException)
```
HauklandJ: same typo

src/Altinn.App.Core/Internal/Notifications/Email/EmailNotificationException.cs
martinothamar: oo good catch, thanks

ivarne (commented):
A few comments. I note a few changes from the approach used so far. Do not interpret as a command that we should not change.

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Sms/SmsNotificationClient.cs
martinothamar: I saw that it was used once, so that's why I dared do it myself too 😄 
My rationale: object lifecycle and allocation is easier to reason about - I know `SmsNotificationClient` and `IHttpClientFactory` are singletons, whenever I create and dipose the clients the caching mechanism for underlying handlers in `IHttpClientFactory` have the chance to do their thing. With typed clients, the lifetime is an implementation detail (transient atm), so the lifetime of the client (and when the factory can manage the lifetime of the handlers) depends on the surrouding scope (e.g. if the client was actived from a scoped or root `IServiceProvider`)

I usually try to make code do as little work as possible aside from performing the desired function (i.e. simple code). Transient lifetime for instance is not required and generates noise (in the form of memory allocation and code that runs at times I don't really care for 😄).

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Email/EmailNotificationClient.cs
martinothamar: I didn't really intend for it to be used by any other classes than the notification clients, so that's why I didn't parameterize it by name or anything like that.

I'm suprised you categorize `using` as magic, it's a pretty standard part of .NET dev vocabulary isn't it? I only made it to not duplicate the control flow across those two clients, since the App Insights API is a bit bad as you note

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Sms/SmsNotificationClient.cs
martinothamar: Good point! I liked `JsonException`, since the method returns a non-nullable response

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Sms/SmsNotificationClient.cs
martinothamar: Right, but I saw that it always expects to add the `Authorziation` header, whereas we don't for the Notifications API.

> Longer term we should probably consider providing something that looks like.. 

Yeah that sounds like a nice API to have

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Email/EmailNotificationClient.cs
ivarne: I think it is magic when used as a one-off. When reviewing this PR it introduces a new concept that I need to read up on and try to verify correct implementation of the pattern, when it only saves a few lines here and nowhere else (as it isn't generic).

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Sms/SmsNotificationClient.cs
ivarne: It requires `AddHttpClient` not just injecting `HttpClientFactory` though. Otherwise it is nice and composable and even pretty easy to add `AddDependencyLogging(string dependencyName)` to encapsulate all the `TelemetryClient` shenanigans too.

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Telemetry.cs
```diff
@@ -0,0 +1,55 @@
+using System.Diagnostics;
+using Microsoft.ApplicationInsights;
+using Prometheus;
+
+namespace Altinn.App.Core.Infrastructure.Clients.Notifications;
+
+internal static class Telemetry
+{
+    internal static readonly Counter OrderCount = Metrics
+        .CreateCounter("altinn_app_notification_order_request_count", "Number of notification order requests.", labelNames: ["type", "result"]);
```
RonnyB71: Are we with this able to break the metrics down by service owner and app?

ivarne: I would guess those metadata are added when collecting the metrics?

RonnyB71 (approved):
🕹️

src/Altinn.App.Core/Infrastructure/Clients/Notifications/Telemetry.cs
RonnyB71: Could be, I'm not sure.

martinothamar: Do you know where to find the metrics? I looked at the tt02 cluster at the `kube-prometheus-stack` Prometheus config and there is no configuration for externalLabels, where you would typically find this (if labels are added after/during scraping). But find for example `altinn_app_instances_created` in neither Grafana nor App Insights in Azure

martinothamar: As agreed I will test and verify this when working on this PR: https://github.com/Altinn/altinn-studio-charts/pull/41

martinothamar:
@ivarne I've changed things now so they follow the standard conventions like discussed yesterday

src/Altinn.App.Core/Internal/Notifications/Email/IEmailNotificationClient.cs
```diff
@@ -0,0 +1,18 @@
+using Altinn.App.Core.Models.Notifications.Email;
+
+namespace Altinn.App.Core.Internal.Notifications.Email;
+
+/// <summary>
+/// Client for managing Altinn email notifications
+/// </summary>
+public interface IEmailNotificationClient
```
ivarne: Sorry I didn't see this before, but why is the public interface (that I assume we expect app developers to use) in the internal namespace?

ivarne (approved):

src/Altinn.App.Core/Internal/Notifications/Email/IEmailNotificationClient.cs
martinothamar: I've wondered that as well, there are tons of other public interfaces in the internal folder. I asked about this at some point but didn't really get an answer (i.e. seemed accidenal or atleast not intentional)

ivarne: I would hope those interfaces are not intended for direct usage in app code.

martinothamar: Having "pubternal" stuff is not unprecedented, but there is a lot of stuff in there..
Notification interfaces are atleast intended to be public. Me making the implementation classes internal is also a deviation from the convention here..

ivarne: I'm not suggesting to make the interface internal, but to find a place for it where it isn't in a namespace named "Internal"

PR #533, opened 2024-03-21
Csharpier for more consistent formatting

martinothamar:
## Description
A big PR, but just as a proposal :)
`dotnet format` is very limited in what it enforces. Csharpier brings more consistency, for example in placement of linebreaks.
Also added VSCode settings for `formatOnSave` so that we don't have to think about running formatting manually. In this configuration it runs both on build and on save in vscode

## Related Issue(s)
N/A

## Verification
- [x] All tests run green

## Documentation


bjorntore:
I'm certainly pro enforcing the consistent formatting. I don't mind Csharpier's style, as long as everyone agrees.

bjorntore (approved):

.vscode/settings.json
```diff
@@ -1,3 +1,7 @@
 {
-    "dotnet-test-explorer.testProjectPath": "**/*Tests*.csproj"
-}
\ No newline at end of file
+    "dotnet-test-explorer.testProjectPath": "**/*Tests*.csproj",
+    "[csharp]": {
+        "editor.defaultFormatter": "csharpier.csharpier-vscode",
+        "editor.formatOnSave": true
```
bjorntore, line 5: Sure everyone want wants format on save? I have it enabled myself in Rider, but some people might prefer doing it manually. 🤷‍♂️ As long as we run a check in Github for the correct formatting.

martinothamar: True! The MSBuild package will ensure everything is formatted on build anyway. It's also possible to run formatting on git commit hook if we don't want formatting on build either, there's an example of that here: https://github.com/martinothamar/Mediator/blob/1ebfbd6ff73b5b19ff8ccaec2d66d81b3b1d9cc1/.husky/task-runner.json#L6

martinothamar: I just don't like to discover formatting issues only when I've pushed the code, might as well make it impossible to not have it properly formatted 😄

bjorntore: Yea, I've used hook before as well. And I always enable onSave.

ivarne:
I also like consistent formatting.

I kind of like the permissive nature of `dotnet format` because it allows me to shape the code to make it more readable. The typical example for me is when I have a repeating structure with different variable names I want all or none of the calls to be split over multiple lines

```C#
FunctionCall(arg1, arg2, arg3)
FunctionCall(arg4, arg5, arg6)
FunctionCall(arg7, arg8, arg9)
FunctionCall(
    arg10,
    arg11,
    arg12,
)
```
A very strict formatter makes such code seem like there is something different happening when the variable names gets longer.

Another example I use is to have compact part of a function to a one-liner, because it isn't that important for the understanding of the code, and it is short and obvious enough that extracting it to a function doesn't really make sense. Especially in tests I think it makes sense to have a one liner initialisation of some object.

```C#
var obj = new Model{Name="ivar", Age="35"}
```
Instead of blowing it up over many lines.

Another issue is that this PR changes a lot of code, not because it was inconsistent, but just because the style guidelines are different. Especially the rule whether `)` is on the last argument or a new line seems a little noisy to change. I don't have an important opinion on where it should be, but changes that causes it to be harder to traverse git history with `git blame` should have a justification.

martinothamar:
> I kind of like the permissive nature of dotnet format because it allows me to shape the code to make it more readable.

I don't know if it's possible to both have permissiveness and consistent formatting at the same time. In this case, `dotnet format` will never provide consistency for line breaks. I've been using csharpier in some capacity for a while now and it's only happened to me <10 times. In such cases I've changed naming and/or block indentation to make sure all the enumerated cases fit within the column width. If that's also not possible, [csharpier can be told to ignore a set of lines](https://csharpier.com/docs/Ignore), though I've never resorted to this. So I think it's possible to both have consistent line lengths _and_ nice-looking enumerations like that, at least it's never been a big deal in the previous projects where I've used csharpier (or similar formatters for other languages such as prettier, rustfmt, gofmt)

> Instead of blowing it up over many lines.

Csharpier is fine with one-liners like that. It only has to fit within the configured column width (I've used anything from 90-120 before), it can be configured in `.editorconfig` or csharpier specific configfile

> Another issue is that this PR changes a lot of code

Yeah this sucks a bit, for the changed code we would have to go +1 commit further back to see the real change

HauklandJ (approved):
Big fan of format-on-save

.vscode/settings.json
RonnyB71: _“In matters of principle, stand like a rock; in matters of taste, swim with the current.” - Thomas Jefferson_

I'm sure there are things in CSharpier I don't like, but I prefer that we all use the same - and I can live with the things I don't like.

RonnyB71 (approved):

PR #543, opened 2024-04-02
POC for Maskinporten operator

martinothamar:
POC for Maskinporten k8s operator using kubebuilder, using this fork for the app helm charts: https://github.com/Altinn/altinn-studio-charts/compare/main...martinothamar:altinn-studio-charts:feature/maskinporten-reqs

Currently just some scaffolding and local development setup - a secret is reconciled based on the `MaskinportenClient` CRD.
This secret is mounted as a volume into a test application running in a local cluster


martinothamar:
Real work starting at https://github.com/Altinn/altinn-k8s-operator

PR #555, opened 2024-04-05
Adds issue labeler action

danielskovli:
Not entirely sure how to handle the slash in the `status/triage` label name. Adding quotations seem to satisfy the yaml schema, but it's currently unknown if this will satisfy the `issue-labeler` action.


RonnyB71 (approved):
🕹️

PR #563, opened 2024-04-05
Debugs issue labeler action

danielskovli:
Temporarily switches to a label with only standard characters in its name, to eliminate the possibility of config/formatting errors.


danielskovli:
Hmm, feil merge, standby

RonnyB71 (approved):

PR #567, opened 2024-04-07
RFC: Add IFileUploadValidator to mirror the pattern of the new validator classes

ivarne:
Some service owners has asked for validation of uploaded files. Since we decided to use a new approach for the new validation classes, I think it would make sense to provide a new interface `IFileUploadValidator` to provide a way that follows the same patterns as the new validation interfaces to validate file uploads.

## Description
`IFileValidator` and `IFileAnalyser` are registered in custom fields on the `DataType` object in applicationmetadata.json. All the other new validation interfaces uses a different pattern where the validator has a `DataType`/`TaskId` field to configure when it should be executed. 

This PR: 
* Adds a new `IFileUploadValidator ` interface to match the new validation interfaces (including language parameter)
* Adds default implementation `LegacyFileAnalyzerValidator` that ensures `IFileValidator`/`IFileAnalyser` continue to work in a backwards compatible way.
* Consolidates validation functionality into `ValidationService`/`ValidationFactory`.

```C#
/// <summary>
/// Interface for handling validation of files when uploaded.
/// A validation failiure will cause the upload to be rejected.
/// </summary>
public interface IFileUploadValidator
{
    /// <summary>
    /// The data type that this validator should run for. This is the id of the data type from applicationmetadata.json
    /// The string "*" is a special value that means that the validator should run for all data types that don't have app-logic.
    /// </summary>
    string DataType { get; }

    /// <summary>
    /// Returns the group id of the validator.
    /// The default is based on the FullName and DataType fields, and should not need customization
    /// </summary>
    string ValidationSource => $"{this.GetType().FullName}-{DataType}";

    /// <summary>
    /// Validating a an uploaded file to possibly prevent the upload.
    /// The file is considered valid if the list of issues has no entries where <see cref="ValidationIssue.Severity"/> is Error.
    /// </summary>
    Task<List<ValidationIssue>> Validate(Instance instance, DataType dataType, byte[] fileContent, string? filename, string? mimeType, string? language);
}
```

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


bjorntore (approved):

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -187,20 +183,9 @@ public class DataController : ControllerBase
...
-                        fileAnalysisResults = await _fileAnalyserService.Analyse(dataTypeFromMetadata, fileStream, filename);
-                    }
-
-                    bool fileValidationSuccess = true;
-                    List<ValidationIssue> validationIssues = new();
-                    if (FileValidationEnabledForDataType(dataTypeFromMetadata))
-                    {
-                        (fileValidationSuccess, validationIssues) = await _fileValidationService.Validate(dataTypeFromMetadata, fileAnalysisResults);
-                    }
+                    var validationIssues = await _validationService.ValidateFileUpload(instance, dataTypeFromMetadata, fileStream.ToArray(), filename, streamContent.Headers.ContentType!.ToString(), language);
```
martinothamar: Intead of non-null assertion on `streamContent.Headers.ContentType`, should we return a bad request if there is no content type header? Unless that is already being done somewhere?

martinothamar: Nvm, seems like `mimeType` is nullable on `ValidateFileUpload`, so I think it should be `?` instead?

src/Altinn.App.Core/Features/FileAnalyzis/IFileAnalyserFactory.cs
```diff
@@ -1,16 +0,0 @@
-﻿using Altinn.App.Core.Features.FileAnalysis;
-using Altinn.Platform.Storage.Interface.Models;
-
-namespace Altinn.App.Core.Features.FileAnalyzis
-{
-    /// <summary>
-    /// Interface responsible for resolving the correct file analysers to run on against a <see cref="DataType"/>.
-    /// </summary>
-    public interface IFileAnalyserFactory
```
martinothamar: Can we actually delete like this? This would be a breaking change, but the PR is marked as a feature

src/Altinn.App.Core/Features/Validation/Default/LegacyFileAnalyzerValidator.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Threading.Tasks;
+using Altinn.App.Core.Features.FileAnalysis;
+using Altinn.App.Core.Models.Validation;
+using Altinn.Platform.Storage.Interface.Models;
+
+/// <summary>
+/// Implementation of <see cref="IFileUploadValidator"/> that uses the legacy 
+/// <see cref="IFileValidator"/> and <see cref="IFileAnalyser"/>.
+/// </summary>
+public class LegacyFileAnalyzerValidator : IFileUploadValidator
```
martinothamar, line 13: I think this should be `internal` and `sealed`? This was meant to bridge the migration to the the new interfaces right, so we have some expectations from the behavior here?

src/Altinn.App.Core/Features/Validation/Default/LegacyFileAnalyzerValidator.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <inheritdoc/>
+    public async Task<List<ValidationIssue>> Validate(Instance instance, DataType dataType, byte[] fileContent, string? filename, string? mimeType, string? language)
+    {
+        var fileAnalysisResults = new List<FileAnalysisResult>();
+        if (dataType.EnabledFileAnalysers?.Count > 0)
+        {
+            // Run file analysis
+            foreach (var analyserName in dataType.EnabledFileAnalysers)
+            {
+                var analyser = _fileAnalysers.First(a => a.Id == analyserName);
```
martinothamar: nit/question: throw a nicer exception as opposed to the builtin `InvalidOperationException` that `First` might throw. I.e. `throw new InvalidOperationException("Could not find file analyzer with ID: " + analyzerName)`.

My assumption is that `analyserName` comes from the app metadata JSON file. Ignore me if this is already validated on startup or something 😄

martinothamar: another q for my learning purposes since I'm new: I'm not sure if this behavior is defined or communicated somewhere, but in places where we call user code, that user code migh throw exceptions. In this case this shortcircuits the whole request. Another option would be to gather exceptions in a loop and throw an `AggregateException` in the end where we end up with some results and some exceptions. Another would be to make "unexpected exception" a special type of analysis result. If this code was instrumented, we could also signal failure on spans/activities for individual analyzers for example. 

You've probably thought and decided on this stuff before, but it's helpful to know the context and design decisions made for me as a new person

src/Altinn.App.Core/Internal/Validation/IValidationService.cs
```diff
@@ -41,6 +41,12 @@ public interface IValidationService
     /// <returns>List of validation issues for this data element</returns>
     Task<List<ValidationIssue>> ValidateDataElement(Instance instance, DataElement dataElement, DataType dataType, string? language);
 
+    /// <summary>
+    /// Validates a file upload. Will prevent the file from being stored if
+    /// issues with severity Error are found.
+    /// </summary>
+    Task<List<ValidationIssue>> ValidateFileUpload(Instance instance, DataType dataType, byte[] fileContent, string? filename, string? mimeType, string? language);
```
martinothamar: Might be too exotic, but if we used `ReadOnlyMemory<byte>` here that might be a better API in terms of
* Communicating through the typesystem the expected use of `fileContent` here (the validator should only need to read)
* Would allow us to potentially pool buffers in the future (give us more flexibility for managing perf and allocations)

src/Altinn.App.Core/Internal/Validation/ValidationService.cs
```diff
@@ -104,6 +104,33 @@ public async Task<List<ValidationIssue>> ValidateDataElement(Instance instance,
...
 
+    /// <inheritdoc/>
+    public async Task<List<ValidationIssue>> ValidateFileUpload(Instance instance, DataType dataType, byte[] fileContent, string? filename, string? mimeType, string? language)
+    {
+        ArgumentNullException.ThrowIfNull(instance);
+        ArgumentNullException.ThrowIfNull(dataType);
+        ArgumentNullException.ThrowIfNull(fileContent);
+
+        var validators = _validatorFactory.GetFileUploadValidators(dataType.Id);
+        var issuesLists = await Task.WhenAll(validators.Select(async v =>
```
martinothamar: really nitpicking probably: `Task.WhenAll` leads to unbounded concurrency (potentially a lot of tasks are queued on the scheduler, though that is probably never the case in our context). Another option would be `Pallel.ForEachAsync`, where the default level of parallelism is equal the the logical processor count. This doesn't return anything though, so we would have to do the "gather"/"reduce" part on our own

martinothamar (commented):
Pretty limited understanding of the whole context here, but approach looks good to me. I'm uncertain about removing public types and such though. Since it seemed like we agreed to follow semver pretty strictly?

src/Altinn.App.Core/Features/FileAnalyzis/IFileAnalyserFactory.cs
ivarne: I think we can. Very few apps has used the file analysers and there should be no need to have a custom factory.

src/Altinn.App.Api/Controllers/DataController.cs
ivarne: Probably.
```suggestion
                    var validationIssues = await _validationService.ValidateFileUpload(instance, dataTypeFromMetadata, fileStream.ToArray(), filename, streamContent.Headers.ContentType?.ToString(), language);
```
I was under the impression that `DataRestrictionValidation.CompliesWithDataRestrictions` on line 160 would ensure that this was available, but it looks like I was wrong.

src/Altinn.App.Core/Features/Validation/Default/LegacyFileAnalyzerValidator.cs
ivarne: I just followed previous practice. Having this internal would make it more awkward for users to unregister this implementation from the DI container, and they would have to do it in a way that does not cause compiler warnings if this implementation is removed. I don't have a strong opinion either way.

src/Altinn.App.Core/Features/Validation/Default/LegacyFileAnalyzerValidator.cs
ivarne: 1) Yes, totally valid point. I'll add your suggestion
```suggestion
                var analyser = _fileAnalysers.FirstOrDefault(a => a.Id == analyserName)
                    ?? throw new InvalidOperationException("Could not find file analyzer with ID: " + analyzerName);
```

ivarne: I don't think we have a consistent view of how to handle exceptions in extension points. In validations I used this approach which at least logs each exception individually, even though I wanted to do them in parallel because they most likely will call external apis and waiting in sequence does not make sense. You seem to have more experience than me so please do add your suggestions.

src/Altinn.App.Core/Internal/Validation/IValidationService.cs
ivarne: I considered it, but a downside is that it will require a copy for apis that require an array. Maybe that is a risk we should be willing to take, especially considering that we now support apps that might mutate the content before saving.

src/Altinn.App.Core/Internal/Validation/ValidationService.cs
ivarne: Is it really unbounded when it is limited by the number of validators the app developer adds.

I consciously chose this approach because validation is likely to be reasonably fast, unless it makes an external http call, in which case I don't think it makes sense to limit the parallelism.

src/Altinn.App.Core/Features/Validation/Default/LegacyFileAnalyzerValidator.cs
martinothamar: Should users be able to unregister this? I thought it was needed to still support the file analyzer abstraction

src/Altinn.App.Core/Internal/Validation/IValidationService.cs
martinothamar: Yeah this kind of thing works better if its' use is consistent throughout, so might be better to do a larger breaking change once. Atleast most (if not all) of BCL understands ROM/ROS by now

src/Altinn.App.Core/Internal/Validation/ValidationService.cs
martinothamar: Used the word unbounded to indicate that _we_ don't enforce any bound. In terms of defensive programming I tend to try to put bounds on things so that even if users do something that isn't reasonable, the runtime of my code is still fairly predictable. But yeah it seems really really unlikely to ever be an issue here😄 

Yeah I agree that the "ideal bound" increases the more async the workload is,

src/Altinn.App.Core/Features/Validation/Default/LegacyFileAnalyzerValidator.cs
ivarne: I don't see any reason why we should actively prevent users from unregistering anything and replace it with their own interface implementation.

I also can't find any realistic scenario when you'd actually need to unregister this service, but a contrived example might be if you had a badly behaved `IFileAnalyser` in a package that should not be constructed for uploads to all DataTypes, but only the ones where it is actually used. A quick fix would be to replace this implementation that inject `IEnumerable<IFileAnalysr>` with one that fetches from `IServiceProvider` only when it is actually required.

martinothamar: It doesn't prevent unregistering though. My impression from previous discussions was that we would be more intentional about type accessibility going forward, and that any modification to public types is a breaking change. That's what I was considering here when reviewing, and so if I consider the tradeoffs between the ergonomics of unregistering and including this as a public API I personally am more worried about adding another public type

ivarne:
Replaced by `IDataWriteProcessor` with `AbandonAllChanges`

PR #581, opened 2024-04-11
Properly dispose `HttpResponseMessage`  instances from `HttpClient`

martinothamar:
A lot of HTTP responses aren't disposed, this fixes that

## Description
Makes sure all instances of `HttpResponseMessage` are disposed after use. Most changes are simply inserting `using` to make responses `using`-statements. Exceptions:

* The underlying content stream is returned directly to the caller, making it not safe to dispose the response:
  * `PdfGeneratorClient` https://github.com/Altinn/app-lib-dotnet/compare/chore/dispose-http-responses?expand=1#diff-e57271a48749f6cf298b2b3c6f896b7c4048580adbf8489cc61466a56d01a15dR91
  * `DataClient` - https://github.com/Altinn/app-lib-dotnet/compare/chore/dispose-http-responses?expand=1#diff-faf65e5c88ef85660768341399b8faa5ca3a17ce0cd3b8f68d23c6d16331609bR147
* `GetAltinn2Codelist` was refactored slightly to make control flow simpler when disposing the responses
  * This should probably be tested manually even if it is part of a lot of tests

Proposal for future work
*  `PlatformHttpException` should not expose the whole `HttpResponseMessage` object as an instance, since the content now can't be read (since it's disposed). We can create properties for the response information that is safe to read. All we read now is `StatusCode`, which is completely safe
* I saw some places where `JsonSerializerOptions` was new'ed up per request, but these options are meant to be cached as they hold some state/cache for underlying serializers

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/test/Altinn.App.Api.Tests/Mocks/InstanceClientMockSi.cs
+++ b/test/Altinn.App.Api.Tests/Mocks/InstanceClientMockSi.cs
@@ -243,8 +243,9 @@ namespace Altinn.App.Api.Tests.Mocks
 
             if (substatus == null || string.IsNullOrEmpty(substatus.Label))
             {
-                throw await PlatformHttpException.CreateAsync(
-                    new HttpResponseMessage { StatusCode = System.Net.HttpStatusCode.BadRequest });
+                using var response = 
+                    new HttpResponseMessage { StatusCode = System.Net.HttpStatusCode.BadRequest };
+                throw await PlatformHttpException.CreateAsync(response);
             }
 
             string instancePath = GetInstancePath(instanceOwnerPartyId, instanceGuid);
@@ -402,7 +403,7 @@ namespace Altinn.App.Api.Tests.Mocks
             if (!string.IsNullOrEmpty(invalidKey))
             {
                 // platform exceptions.
-                HttpResponseMessage res = new()
+                using HttpResponseMessage res = new()
                 {
                     StatusCode = System.Net.HttpStatusCode.BadRequest,
                     Content = new StringContent($"Unknown query parameter: {invalidKey}")
--- a/test/Altinn.App.Api.Tests/Mocks/InstanceClientMockSi.cs
+++ b/test/Altinn.App.Api.Tests/Mocks/InstanceClientMockSi.cs
@@ -243,7 +243,7 @@ namespace Altinn.App.Api.Tests.Mocks
 
             if (substatus == null || string.IsNullOrEmpty(substatus.Label))
             {
-                using var response = 
+                using var response =
                     new HttpResponseMessage { StatusCode = System.Net.HttpStatusCode.BadRequest };
                 throw await PlatformHttpException.CreateAsync(response);
             }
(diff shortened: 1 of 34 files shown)


HauklandJ (approved):
LGTM -  the using statement is considered a best practice when dealing with objects that implement IDisposable, like HttpResponseMessage

bjorntore (approved):

test/Altinn.App.Api.Tests/Mocks/InstanceClientMockSi.cs
```diff
@@ -243,8 +243,9 @@ public async Task<Instance> UpdateSubstatus(int instanceOwnerPartyId, Guid insta
 
             if (substatus == null || string.IsNullOrEmpty(substatus.Label))
             {
-                throw await PlatformHttpException.CreateAsync(
-                    new HttpResponseMessage { StatusCode = System.Net.HttpStatusCode.BadRequest });
+                using var response = 
+                    new HttpResponseMessage { StatusCode = System.Net.HttpStatusCode.BadRequest };
+                throw await PlatformHttpException.CreateAsync(response);
```
ivarne, line 248: This seems wrong. Doesn't this throw a reference to the disposed message?

It probably won't matter unless the tests tries to access the disposed `Response` property.

ivarne (approved):

test/Altinn.App.Api.Tests/Mocks/InstanceClientMockSi.cs
martinothamar: If I interpret you right you are pointing out what I worry about too about this, which is that the exception will have a reference to a disposed response, whereas it didn't before. Before this PR I think

* Catchers of this exception could potentially read the response, since we didn't dispose it (I'm assuming we're either leaking the underlying resource now or that it has a finalizer that disposes upon GC)

After this PR
* Attempts to do anything with the content of the response in a catch block will throw an exception, while e.g. `StatusCode` still works fine (not guarded by `bool _disposed`)

If you're talking about the order of operations here, which is also potentially confusing, `CreateAsync` is fully completed before the response is disposed (so `CreateAsync` will have no problem reading the content)

martinothamar: This discussion also plays into `HttpCompletionOption`. Default behavior (and the only behavior we use AFAICT) of `HttpClient` is to buffer responses into a `MemoryStream`, in which case the `Dispose` implementation doesn't seem to do anything apart from marking the stream as disposed and no longer readable.

In the `HttpCompletionOption.ResponseHeadersRead` case however, there is more work being done on dispose and I think this is where it is particularly important to clean up.

We could do this cleanup as part of v9 instead, and make a breaking change on `PlatformHttpException` so that it doesn't expose the whole response (only the information which is safely referenced even after dispose)

ivarne: I think you are right. I was just confused why you skipped adding the `using` statements in other places this issue happened, but did here. Thinking more I don't think it's an issue.

ivarne:
Can we have a reduced version of this PR that adds `using` only to the places that doesn't cause breaking change?

martinothamar:
I think like 95% of the functions here end up with `PlatformHttpException.CreateAsync`, so I'm tempted to just close this and redo it when we start 9.0, probably won't take much time then either

Issue #552, opened 2024-04-04
Elements with attributes is not deleted


kentare:
### Description of the bug

When using attributes in the XSD is it not possible to delete the element after a user inserts a value in an input field. Behaviour is present when the element is set as nillable. 

## 1. No value (initial, valid)
![image](https://github.com/Altinn/app-lib-dotnet/assets/24462806/b78c1135-e4a2-4f8c-aeba-d694d5b1cd60)
```xml
<model xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema"/>
```

## 2. User written value (valid)
![image](https://github.com/Altinn/app-lib-dotnet/assets/24462806/11ec8c4c-eaf4-4177-bbfd-773af75a1c44)
```xml
<model xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
   <aarets orid="30320">123</aarets>
</model>
```

## **BUG:** 3. Trying to delete value (invalid, should be the same as no value)
![desimal1](https://github.com/Altinn/app-lib-dotnet/assets/24462806/ff3de4b6-f0d2-4035-851c-44ae7fe24bc4)
```xml
<model xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
    <aarets orid="30320">0</aarets>
</model>
```
Expected and desired behaviour:
The value for this needs to be the same as `1. No value`. `0` is a valid user inputted value, but not a valid value for deletion. 

## Repro from repo
1. Clone minimal reproduction repo: https://altinn.studio/repos/kentare/xmltext-decimal-repro 
2. Start the app
3. Insert a value 
4. Inspect XML
5. Delete value 
6. Inspect XML

### Additional Information

The model is uploaded to altinn.studio to generate the schema and classes. 

The XSD can be found here: https://altinn.studio/repos/kentare/xmltext-decimal-repro/src/branch/master/App/models/model.xsd 

The issue has been discussed here: https://altinn.slack.com/archives/C02EJ9HKQA3/p1710493123111109 

Might be related to: #517 

Discovered in this repo: https://altinn.studio/repos/brg/aarsregnskap-ideelle-202402 

[...]

PR #582, opened 2024-04-11
Split out and improve PrepareModelForXmlStorage from InitializeAltinnRowId

### LINKED ISSUE #517: PATCH: Saving a space and removing it causes 409 crash
### Description of the bug

Typing into a string field, removing the value, adding a space, removing that space (all while waiting for saving in between each step) seems to crash with a 409.

### Steps To Reproduce

https://github.com/Altinn/app-frontend-react/assets/700139/3556b56c-90be-4907-a986-2a28fbd4aa65

### Additional Information

_No response_

ivarne:
This PR has 4 changes
1. Split out `PrepareModelForXmlStorage` from `InitializeAltinnRowId` (just the part that converts `""` to null everywhere)
2. Convert all strings that is `string.IsNullOrWithespace` to null, but only if the `[XmlText]` attribute is set (to match xml serialization)
3. If a class only has a single property without `[BindNever]` and that property has `[XmlText]`, set the parent element to null, as it makes no sense to store a tagg without a value, even if there are `fixed` attributes.

bullet 3 was suggested by @kentare, but now that I think about it, I don't really like that we change behaviour with regards to that issue.

## Related Issue(s)
- Fixes #517
- Part of a solution for #552 (but the issue with xml not allowing `decimal`)

## Verification
- [ ] All tests run green

### DIFF
--- a/src/Altinn.App.Api/Controllers/ActionsController.cs
+++ b/src/Altinn.App.Api/Controllers/ActionsController.cs
@@ -181,6 +181,7 @@ public class ActionsController : ControllerBase
             }
 
             ObjectUtils.InitializeAltinnRowId(newModel);
+            ObjectUtils.PrepareModelForXmlStorage(newModel);
 
             var dataElement = instance.Data.First(d => d.Id.Equals(elementId, StringComparison.OrdinalIgnoreCase));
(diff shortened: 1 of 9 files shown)


bjorntore (approved):

ivarne:
Closing this, as #590 has a better approach that reuses a concept from xml serialisation (`ShouldSerialize*`) instead of implementing some heuristics on `[XmlText]` and `[BindNever]`/`[XmlIgnore]` on all properties.

PR #583, opened 2024-04-12
Make `System.Text.Json.JsonSerializerOptions` instances static

martinothamar:
## Description

`JsonSerializerOptions` should not be created per serializer call, but should be cached and reused as it keeps some cached state internally that it then doesn't have to recreate every time.

Docs: 
https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/configure-options?pivots=dotnet-8-0#reuse-jsonserializeroptions-instances

Improved adherence to some conventions
* Names  = `_jsonSerializerOptions` (when private)
* Statics to the top of the class (along with other statics)
* Target typed new

In the future we should probably consolidate since there is at least some duplication

## Related Issue(s)
- N/A

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Process/ExpressionsExclusiveGateway.cs
+++ b/src/Altinn.App.Core/Internal/Process/ExpressionsExclusiveGateway.cs
@@ -18,7 +18,7 @@ namespace Altinn.App.Core.Internal.Process
     /// </summary>
     public class ExpressionsExclusiveGateway : IProcessExclusiveGateway
     {
-        private static readonly JsonSerializerOptions _jsonSerializerOptions = new JsonSerializerOptions()
+        private static readonly JsonSerializerOptions _jsonSerializerOptions = new()
         {
             AllowTrailingCommas = true,
             ReadCommentHandling = JsonCommentHandling.Skip,
--- a/src/Altinn.App.Core/Internal/Process/ExpressionsExclusiveGateway.cs
+++ b/src/Altinn.App.Core/Internal/Process/ExpressionsExclusiveGateway.cs
@@ -18,6 +18,17 @@ namespace Altinn.App.Core.Internal.Process
     /// </summary>
     public class ExpressionsExclusiveGateway : IProcessExclusiveGateway
     {
+        private static readonly JsonSerializerOptions _jsonSerializerOptions = new JsonSerializerOptions()
+        {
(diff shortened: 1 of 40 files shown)


bjorntore (approved):

src/Altinn.App.Core/Internal/Process/ExpressionsExclusiveGateway.cs
```diff
@@ -18,6 +18,17 @@ namespace Altinn.App.Core.Internal.Process
...
+        private static readonly JsonSerializerOptions _jsonSerializerOptions = new()
+        {
+            AllowTrailingCommas = true,
+            ReadCommentHandling = JsonCommentHandling.Skip,
+            PropertyNameCaseInsensitive = true,
+        };
+        private static readonly JsonSerializerOptions _jsonSerializerOptionsCamelCase = new() 
+        {
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase
+        };
```
ivarne, line 30: Great to see that you preserve existing behaviour when doing these kind of rewrites, even though there is no reason why there should be different serializer options here.

test/Altinn.App.Core.Tests/Helpers/ShadowFieldsConverterTests.cs
```diff
@@ -45,6 +45,8 @@ public void ShouldRemoveShadowFields_WithPrefix()
         Assert.Contains("AltinnSF_", serializedDataWithoutModifier);
 
         var modifier = new IgnorePropertiesWithPrefix("AltinnSF_");
+
+        // TODO: could be cached in a dictionary with the prefix as key, since serializer options _should_ be static/reused
```
ivarne: No, we should not use a dictionary to cache the json serializer in tests, because they won't be reused anyway.

ivarne (approved):
I like `_jsonSerializerOptions`, but I get a hint in Rider when using that name.

Screenshot 2024-04-14 at 21 59 37

test/Altinn.App.Core.Tests/Helpers/ShadowFieldsConverterTests.cs
martinothamar: Removed the comment since I opened a PR for this now: #591

Issue #587, opened 2024-04-12
Support for OpenTelemetry in localtest


HauklandJ:
### Description

Part of the [Monitoring epic](https://github.com/Altinn/app-lib-dotnet/issues/580).

### Additional Information

LGTM stack - [Loki](https://grafana.com/oss/loki/), [Grafana](https://grafana.com/), [Tempo](https://grafana.com/oss/tempo/), [Mimir](https://grafana.com/oss/mimir/)

PR #589, opened 2024-04-12
Add OpenTelemetry configuration and instrumentation to replace AppInsights and prometheus-net

### LINKED ISSUE #586: OpenTelemetry observability
### Description

Part of the [Monitoring epic](https://github.com/Altinn/app-lib-dotnet/issues/580).

### Additional Information

_No response_

HauklandJ:
## Related Issue(s)
- #586
- #601 
- #587
- #593

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -104,6 +109,40 @@ private static void AddApplicationInsights(IServiceCollection services, IConfigu
...
+                        .AddOtlpExporter();
+                })
+                .WithMetrics( x =>
+                {
+                    x.AddRuntimeInstrumentation()
+                        .AddMeter(
+                            "Microsoft.AspNetCore.Hosting",
+                            "Microsoft.AspNetCore.Server.Kestrel",
+                            "System.Net.Http",
+                            "Altinn.App.Api");
```
martinothamar: This will do for the MS/bcl stuff:
```csharp
.AddRuntimeInstrumentation()
.AddHttpClientInstrumentation()
.AddAspNetCoreInstrumentation()
.AddOtlpExporter(
    (_, readerOptions) =>
    {
        if (env.IsDevelopment())
        {
            // Export interval should be set by env var when running in real environments
            // but locally it's nice to receive metrics more often than the default 60s
            readerOptions.PeriodicExportingMetricReaderOptions.ExportIntervalMilliseconds = 10_000;
        }
    }
)
```

src/Altinn.App.Api/Infrastructure/Telemetry/IMetricFactory.cs
```diff
@@ -0,0 +1,58 @@
+using System.Diagnostics.Metrics;
+
+namespace Altinn.App.Api.Infrastructure.Telemetry;
+
+/// <summary>
+/// Provides a factory interface for creating metric meters.
+/// This interface extends the IMeterFactory to simplyfy creation
+/// </summary>
+public interface IMetricFactory : IMeterFactory
```
martinothamar: I don't think we should implement `IMeterFactory` ourselves, we could just wrap it in the implementation (it handles the lifecycle of the underlying meters and such). I think our only goal with this is to let users not provide a name (as the name isn't really important in the way we use it)

src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -104,6 +109,40 @@ private static void AddApplicationInsights(IServiceCollection services, IConfigu
...
+                .WithTracing(tpbuilder => 
+                {
+                    if (env.IsDevelopment())
+                    {
+                        tpbuilder.SetSampler(new AlwaysOnSampler());
+                    }
+
+                    tpbuilder
+                        .AddSource("TODO: find docker-image-tag/assembly-version")
+                        .AddHttpClientInstrumentation()
```
martinothamar: We should go through the options for the instrumentations as well, most of them can be configured to include exceptions as span events, for example:
```csharp
.AddHttpClientInstrumentation(opts => opts.RecordException = true)
```

martinothamar (commented):
Nice! 

We need logging as well, something like 
```csharp
builder.Logging.AddOpenTelemetry(options =>
{
    options.AddOtlpExporter();
});
```

Not sure what we should do about `ActivitySource`. We could inject it directly perhaps, it is built into BCL after all.

In the end I think we want abstractions that
* Make it simple to create traces and metrics
* Let DI contain and manage lifecycle of `Meter`'s and `ActivitySource`'s (then we can also make assertions in tests based on telemetry if we want)

HauklandJ:
Before merge:

remove TelemetryClient,
remove prometheus endpoint and metrics,
telemetry collector must exist and be connected to Azure

martinothamar:
Time to update already 😄  https://github.com/advisories/GHSA-vh2m-22xx-q94f

martinothamar:
I think we should set the `IncludeFormattedMessage` flag to `true`, this "renders" the log message and includes the format as an attribute instead: https://github.com/open-telemetry/opentelemetry-dotnet/blob/57c442cc7b2fe93c81ca877de28f5204fdb11ebb/src/OpenTelemetry/Logs/ILogger/OpenTelemetryLogger.cs#L93

`IncludeFormattedMessage = false`: 
image

`IncludeFormattedMessage = true`: 
image

maskinporten-poc/kubebuilder/test/app/altinn-studio-charts
ivarne, line 1: maskinporten-poc?

martinothamar: #543 is leaking 😄

ivarne: Yes. Submodules are tricky to work with

src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -54,7 +57,22 @@ public static void AddAltinnAppServices(this IServiceCollection services, IConfi
             services.AddAppServices(config, env);
             services.ConfigureDataProtection();
 
-            AddApplicationInsights(services, config, env);
+            var useOpenTelemetrySetting = config["AppSettings:UseOpenTelemetry"];
```
ivarne: Doesn't this perform much of the same function?
```suggestion
            var useOpenTelemetrySetting = config.GetValue<bool>("AppSettings:UseOpenTelemetry");
```

src/Altinn.App.Api/Infrastructure/Middleware/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,81 @@
...
+/// <summary>
+/// Middleware for adding telemetry to the request
+/// </summary>
+internal static class TelemetryMiddleware
+{
+    /// <summary>
+    /// Adds telemetry to the request
+    /// </summary>
+    /// <param name="app">App</param>
+    internal static IApplicationBuilder UseTelemetry(this IApplicationBuilder app)
```
ivarne: Having this `internal` means that an app is required to use `UseAltinnAppCommonConfiguration` in order to apply our telemetry enriching middleware. This means that a developer can't override our standard pipeline. (eg to add middleware in the middle of the pipeline)

Another issue is that this extension does not add telemetry, but just sets some activity tags from the cookie. Should it be named `UseAddTelemetryTagsFromClaims` instead?

src/Altinn.App.Core/Altinn.App.Core.csproj
```diff
@@ -21,8 +21,8 @@
     <PackageReference Include="Microsoft.FeatureManagement.AspNetCore" Version="3.2.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageReference Include="Scrutor" Version="4.2.2" />
-    <PackageReference Include="prometheus-net" Version="8.2.1" />
-    <PackageReference Include="prometheus-net.AspNetCore" Version="8.2.1" />
+    <PackageReference Include="NetEscapades.EnumGenerators" Version="1.0.0-beta08" PrivateAssets="all" ExcludeAssets="runtime" />
```
ivarne, line 23: More dependencies to Core? or is this EnumGenerator strictly a compile time dependency that will never have version conflict with usage in apps?

`OpenTelemetry.Api` is a necessary dependency (I guess).

src/Altinn.App.Core/Features/Telemetry.Instance.cs
```diff
@@ -0,0 +1,144 @@
...
+        activity?.SetTag(Labels.InstanceGuid, instanceGuid);
+        activity?.SetTag(Labels.InstanceOwnerPartyId, instanceOwnerPartyId);
+        return activity;
+    }
+
+    internal static class Instance
+    {
+        private static readonly string _prefix = "Instance";
+
+        internal static readonly string TraceNameGet = $"{_prefix}.Get";
```
ivarne: I think all these work with `const`
```suggestion
        internal const string TraceNameGet = $"{_prefix}.Get";
```

src/Altinn.App.Core/Internal/Validation/ValidationService.cs
```diff
@@ -136,13 +154,16 @@ private Task<List<ValidationIssue>[]> RunDataElementValidators(Instance instance
...
+        using var activity = _telemetry.StartValidateFormDataActivity();
+
         // Locate the relevant data validator services from normal and keyed services
         var dataValidators = _validatorFactory.GetFormDataValidators(dataType.Id)
             .Where(dv => previousData is null || dv.HasRelevantChanges(data, previousData))
             .ToArray();
 
         var issuesLists = await Task.WhenAll(dataValidators.Select(async (v) =>
         {
+            using var activity = _telemetry.StartRunFormDataValidatorActivity();
```
ivarne: I might be confused, but does this functionally replace the `LogDebug` statement on the next line?

Instance id and data element might be tracked at a higher level, but `v.ValidationSource` likely makes a lot of sense to track here (and also instead of `v.GetType().Name` in the `LogDebug` statement.).

ivarne (commented):
A few comments while reading through and trying to understand.

src/Altinn.App.Core/Altinn.App.Core.csproj
martinothamar: `NetEscapades.EnumGenerators` is compile-time only yes, it's not super-important but it does let us avoid allocations when using enums to describe label names

src/Altinn.App.Core/Internal/Validation/ValidationService.cs
martinothamar: Logs are separate from traces (`Activity`s are traces), logs are a simple list of messages while traces form a sort of call-graph. Distributed traces are particularly useful for us since we're in a microservices setting.

Fully agree, lifting tags as high up as possible in the trace hierarchy, and including new information as we go down the stack 👍

src/Altinn.App.Api/Infrastructure/Middleware/TelemetryMiddleware.cs
martinothamar: True! Is `UseAltinnAppCommonConfiguration` optional? Users that don't call `UseAltinnAppCommonConfiguration`, do they know which parts _have_ to be added (if any) and not? Like I would assume authn/z would be a hard requirement?

ivarne: Only requirement is that `docker build` in the folder creates something that starts in kubernetes.

Not sure how much overriding we should support, but making all the middleware public seems sensible to me.

src/Altinn.App.Core/Features/Telemetry.Instance.cs
HauklandJ: Agreed

src/Altinn.App.Core/Altinn.App.Core.csproj
martinothamar: We could avoid `OpenTelemetry.Api` as well (although presumably they will get from Altinn.App.Api anyway), if we just implemented the Exception span/activity event recording code ourselves `RecordException`

ivarne: Is that a lot of code? It would be nice if we didn't place more restrictions on core than strictly necessary.

src/Altinn.App.Api/Infrastructure/Middleware/TelemetryMiddleware.cs
martinothamar: Yeah sounds sensible to me as well, how about `UseTelemetryEnricher` or something? Since it's not strictly tied to tags and claims

src/Altinn.App.Core/Altinn.App.Core.csproj
martinothamar: I'll check it out

src/Altinn.App.Api/Extensions/AltinnAppBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+            "OpenTelemetry is not enabled - set 'UseOpenTelemetry' to 'true' in appsettings to use OpenTelemetry"
+        );
+
+    /// <summary>
+    /// Experimental API for configuring OpenTelemetry logging.
+    /// In newer versions of OpenTelemetry, this API will be improved
+    /// </summary>
+    /// <param name="configure">configuration delegate</param>
+    /// <returns></returns>
+    [Experimental("ALTINN010")]
```
martinothamar: We should consider an analyzer so that we can customize the diagnostic message here

src/Altinn.App.Api/Extensions/WebHostBuilderExtensions.cs
```diff
@@ -15,11 +17,71 @@ public static class WebHostBuilderExtensions
...
+                    Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT")
+                    ?? Environment.GetEnvironmentVariable("DOTNET_ENVIRONMENT")
+                    ?? "Production";
+
+                var (appInsightsKey, appInsightsConnectionString) = ServiceCollectionExtensions.GetAppInsightsConfig(
+                    config,
+                    env
+                );
+                if (
+                    string.IsNullOrWhiteSpace(appInsightsConnectionString) && !string.IsNullOrWhiteSpace(appInsightsKey)
```
martinothamar: Consolidate this logic to `GetAppInsightsConfigForOtel`

src/Altinn.App.Api/Extensions/WebHostBuilderExtensions.cs
```diff
@@ -15,11 +17,71 @@ public static class WebHostBuilderExtensions
...
+                    });
+                }
+                else
+                {
+                    options.AddOtlpExporter();
+                }
+
+                builder.ConfigureServices(services =>
+                {
+                    var appBuilder = services.GetInstanceInServices<AltinnAppBuilder>();
```
martinothamar: Are error messages good enough for this? If AddAltinnAppServices ends up being called after this path

src/Altinn.App.Api/Infrastructure/Middleware/TelemetryEnrichingMiddleware.cs
```diff
@@ -0,0 +1,89 @@
...
+                    else if (claim.Type.Equals(AltinnCoreClaimTypes.AuthenticationLevel))
+                    {
+                        authenticationLevel = Convert.ToInt32(claim.Value);
+                    }
+                }
+
+                trace.Activity.SetTag("user.name", userName);
+                trace.Activity.SetTag("user.id", userId);
+                trace.Activity.SetTag("user.party_id", partyId);
+                trace.Activity.SetTag("user.authentication_level", authenticationLevel);
```
martinothamar: Use const label fields in `Telemetry`?

src/Altinn.App.Api/Infrastructure/Middleware/TelemetryEnrichingMiddleware.cs
```diff
@@ -0,0 +1,89 @@
...
+                    return;
+                }
+
+                string? userName = null;
+                int? userId = null;
+                int? partyId = null;
+                int? authenticationLevel = null;
+                foreach (Claim claim in context.User.Claims)
+                {
+                    if (claim.Type.Equals(AltinnCoreClaimTypes.UserName))
```
martinothamar: Should have explicit `CultureInfo` here

src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -329,16 +328,5 @@ private static void AddFileValidatorServices(IServiceCollection services)
             services.TryAddTransient<IFileValidationService, FileValidationService>();
             services.TryAddTransient<IFileValidatorFactory, FileValidatorFactory>();
         }
-
-        private static void AddMetricsDecorators(IServiceCollection services, IConfiguration configuration)
-        {
-            MetricsSettings metricsSettings =
-                configuration.GetSection("MetricsSettings")?.Get<MetricsSettings>() ?? new MetricsSettings();
```
martinothamar, line 337: MetricsSettigngs should be `Obsolete`

src/Altinn.App.Core/Internal/Validation/ValidationService.cs
```diff
@@ -74,29 +79,31 @@ private Task<List<ValidationIssue>[]> RunTaskValidators(Instance instance, strin
...
                 {
                     _logger.LogError(
                         e,
                         "Error while running validator {validatorName} on task {taskId} in instance {instanceId}",
-                        tv.GetType().Name,
+                        v.GetType().Name,
                         taskId,
                         instance.Id
                     );
+                    activity?.Errored(e);
```
martinothamar, line 106: Need some docs on 
* When do we `RecordException`
* Instrumentation practice when calling into "user code"

src/Altinn.App.Core/Features/Telemetry.cs
```diff
@@ -0,0 +1,158 @@
...
+using Altinn.App.Core.Configuration;
+using Altinn.App.Core.Models;
+using Microsoft.Extensions.Options;
+
+namespace Altinn.App.Core.Features;
+
+/// <summary>
+/// Used for creating traces and metrics for the app.
+/// </summary>
+public sealed partial class Telemetry : IDisposable
```
martinothamar, line 22: Need doc on
* Which part of this is considered public contract (it's not only public members)

src/Altinn.App.Core/Features/Notifications/Email/EmailNotificationClient.cs
```diff
@@ -76,7 +70,8 @@ public async Task<EmailOrderResponse> Order(EmailNotification emailNotification,
                 if (orderResponse is null)
                     throw new JsonException("Couldn't deserialize email notification order response.");
 
-                Telemetry.OrderCount.WithLabels(Telemetry.Types.Email, Telemetry.Result.Success).Inc();
+                _telemetry?.RecordNotificationOrder(_orderType, Telemetry.Notifications.OrderResult.Success);
+                activity?.SetStatus(ActivityStatusCode.Ok);
```
martinothamar: Avoid using status in places like this for now. Status error will be set on the root trace, it's unclear if we need it here (and if we did, we should do it a lot more places as well)

src/Altinn.App.Core/Features/Telemetry.AppOptionsService.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Altinn.App.Core.Features;
+
+partial class Telemetry
+{
+    internal Activity? StartGetOptionsAsyncActivity() =>
+        ActivitySource.StartActivity("AppOptionsService.GetOptionsAsync");
+
+    internal Activity? StartGetOptionsAsyncActivity(InstanceIdentifier instanceIdentifier)
+    {
+        var activity = ActivitySource.StartActivity("AppOptionsService.GetOptionsAsync");
```
martinothamar: Here the span name matches the method name, which sort of introduces coupling (if we change the method name, we should change span name as well). Should span names be more "logical names"/descriptive of the feature or should they match technical method names? Async suffix or not?

src/Altinn.App.Core/Features/Telemetry.ApplicationLanguage.cs
```diff
@@ -0,0 +1,8 @@
+using System.Diagnostics;
+
+namespace Altinn.App.Core.Features;
+
+public partial class Telemetry
```
martinothamar: Normalize to just `partial class Telemetry` everywhere

src/Altinn.App.Core/Features/Telemetry.Authorization.Client.cs
```diff
@@ -0,0 +1,90 @@
...
+    internal Activity? StartClientAuthorizeActionsActivity(
+        Platform.Storage.Interface.Models.Instance instance,
+        List<string> actions
+    )
+    {
+        var activity = ActivitySource.StartActivity($"{_prefix}.AuthorizeActions");
+        if (activity is not null)
+        {
+            activity.SetInstanceId(instance);
+            string actionTypes = string.Join(", ", actions);
```
martinothamar: For complex tags like this, maybe drop or use span event instead?

src/Altinn.App.Core/Features/Telemetry.Authorization.Client.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                activity.SetTag(Labels.TaskId, taskId);
+            }
+        }
+        return activity;
+    }
+
+    internal Activity? StartClientAuthorizeActionsActivity(
+        Platform.Storage.Interface.Models.Instance instance,
+        List<string> actions
```
martinothamar: `actionIds`?

src/Altinn.App.Core/Features/Telemetry.Data.cs
```diff
@@ -0,0 +1,53 @@
...
+    internal Activity? StartDataPatchActivity(Instance instance)
+    {
+        var activity = ActivitySource.StartActivity($"{_prefix}.Patch");
+        activity.SetInstanceId(instance);
+        return activity;
+    }
+
+    internal static class Data
+    {
+        internal static readonly string LabelResult = "result";
```
martinothamar: Standardize how to organize these kinds of fields per partial class

src/Altinn.App.Core/Features/Telemetry.DataClient.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        var activity = ActivitySource.StartActivity($"{_prefix}.UnlockDataElement");
+        activity.SetInstanceId(instanceId);
+        activity?.SetTag(DataClientLabels.DataGuid, dataGuid);
+        return activity;
+    }
+
+    internal static class DataClientLabels
+    {
+        internal const string DataGuid = "DataClient.DataGuid";
```
martinothamar: Maybe consolidate all label names per signal type in one place?

src/Altinn.App.Core/Features/Telemetry.Processes.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    internal void ProcessStarted()
+    {
+        _counters[MetricNameProcessesCreated].Add(1);
+    }
+
+    internal void ProcessEnded(ProcessStateChange processChange)
+    {
+        ArgumentNullException.ThrowIfNull(processChange?.NewProcessState?.Started);
+        ArgumentNullException.ThrowIfNull(processChange?.NewProcessState?.Ended);
```
martinothamar: Maybe just not record duration if they are null?

src/Altinn.App.Core/Features/TelemetryExtensions.cs
```diff
@@ -0,0 +1,148 @@
...
+    /// Sets the Instance GUID as a tag/attribute on the activity/span
+    /// </summary>
+    /// <param name="activity">Activity</param>
+    /// <param name="instance">Instance</param>
+    /// <returns>Activity</returns>
+    public static Activity? SetInstanceId(this Activity? activity, Instance? instance)
+    {
+        if (instance?.Id is not null)
+        {
+            Guid instanceGuid = Guid.Parse(instance.Id.Split("/")[1]);
```
martinothamar: Span-based overload for parsing and better error handling

src/Altinn.App.Core/Infrastructure/Clients/Register/RegisterERClient.cs
```diff
@@ -59,14 +63,16 @@ IAppMetadata appMetadata
             _client = httpClient;
             _accessTokenGenerator = accessTokenGenerator;
             _appMetadata = appMetadata;
+            _telemetry = telemetry;
         }
 
         /// <inheritdoc />
-        public async Task<Organization?> GetOrganization(string OrgNr)
+        public async Task<Organization?> GetOrganization(string orgNr)
```
martinothamar: This is a breaking change technically

src/Altinn.App.Core/Infrastructure/Clients/Storage/InstanceClient.cs
```diff
@@ -104,26 +113,19 @@ public async Task<List<Instance>> GetInstances(Dictionary<string, StringValues>
             );
             QueryResponse<Instance> queryResponse = await QueryInstances(token, apiUrl);
 
-            if (queryResponse.Count == 0)
-            {
-                return new List<Instance>();
```
martinothamar, line 109: I think we should keep this branch?

test/Altinn.App.Core.Tests/Altinn.App.Core.Tests.csproj
```diff
@@ -55,6 +55,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.Testing" Version="8.4.0" />
```
martinothamar: Don't think this is used?

test/Altinn.App.Core.Tests/Mocks/TelemetryFake.cs
```diff
@@ -0,0 +1,166 @@
...
+using Altinn.App.Core.Features;
+using Altinn.App.Core.Models;
+using FluentAssertions;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Altinn.App.Core.Tests.Mocks;
+
+internal sealed record TelemetryFake : IDisposable
```
martinothamar: Maybe rename to `TelemetrySink` or something?

src/Altinn.App.Api/Infrastructure/Middleware/TelemetryEnrichingMiddleware.cs
```diff
@@ -0,0 +1,89 @@
...
+/// </summary>
+public static class TelemetryEnrichingMiddleware
+{
+    /// <summary>
+    /// Adds telemetry to the request
+    /// </summary>
+    /// <param name="app">App</param>
+    public static IApplicationBuilder UseTelemetryEnricher(this IApplicationBuilder app)
+    {
+        app.Use(
```
HauklandJ: wrap in a try catch?

src/Altinn.App.Core/Features/TelemetryExtensions.cs
```diff
@@ -0,0 +1,148 @@
+using System.Diagnostics;
+using Altinn.Platform.Storage.Interface.Models;
+using OpenTelemetry.Trace;
+using Labels = Altinn.App.Core.Features.Telemetry.Labels;
+
+namespace Altinn.App.Core.Features;
```
HauklandJ: consider changing namespace

src/Altinn.App.Core/Features/Telemetry.Notifications.cs
```diff
@@ -0,0 +1,72 @@
...
+using static Altinn.App.Core.Features.Telemetry.Notifications;
+using Tag = System.Collections.Generic.KeyValuePair<string, object?>;
+
+namespace Altinn.App.Core.Features;
+
+partial class Telemetry
+{
+    private void InitNotifications(InitContext context)
+    {
+        InitMetricCounter(
```
HauklandJ, line 20: comment explaining the promethues issue with 0 -> 1

test/Altinn.App.Core.Tests/Mocks/TelemetrySink.cs
```diff
@@ -0,0 +1,166 @@
...
+    {
+        var scope = new Scope();
+        Scopes.TryAdd(scope, default).Should().BeTrue();
+        return scope;
+    }
+}
+
+internal static class TelemetryDI
+{
+    internal static IServiceCollection AddTelemetryFake(this IServiceCollection services)
```
HauklandJ: rename

HauklandJ:
@martinothamar I added a snapshot test poc commit: https://github.com/Altinn/app-lib-dotnet/pull/589/commits/484d2b03856ce4f46aa7bb31bcf09cf771f63146. Some points I would like feedback on: 

- Does adding telemetry tests to already existing tests increase the complexity of the tests too much?
- The verify is for activities as in this example there are no metrics, I was thinking in the case of both we do an additional verify call with CapturedMetrics

martinothamar:
Awesome!

Maybe we should also write down specifically what is considered public contract. I am thinking

* Trace names (`DisplayName`)
* Tag names
* Metric names
* `IdFormat` - important for interoperability, for example App Insights SDK uses W3C

With that in mind we should probably have a custom serializer/model for the snapshot files.
How about a `TelemetrySink.Verify` function that does this (for both activities and metrics in one go)?

> Does adding telemetry tests to already existing tests increase the complexity of the tests too much?

Maybe we should not add the `useTelemetry = false` case, and only test the instrumented path, since in v9 it will always be there? That would be a little less code. If `TelemetrySink` also wraps the verify stuff like suggested above then there wouldn't be too much code to add I think

> The verify is for activities as in this example there are no metrics, I was thinking in the case of both we do an additional verify call with CapturedMetrics

If we just have a single `TelemetrySink.Verify` instance method that outputs 1 file with both activites and metrics, would that work?

martinothamar:
In https://github.com/Altinn/app-lib-dotnet/pull/589/commits/0ba8f5abc4552fde56aa952afcfddf7dcf56f09a I

* Removed `AltinnAppBuilder` and 
* Moved logging configuration into `AddOpenTelemetry`
* Add `APPLICATIONINSIGHTS_CONNECTION_STRING` support: https://learn.microsoft.com/en-us/azure/azure-monitor/app/opentelemetry-configuration?tabs=aspnetcore

The end users can still configure OTel, just by using these DI hooks directly instead

```csharp
void RegisterCustomAppServices(IServiceCollection services, IConfiguration config, IWebHostEnvironment env)
{
    services.ConfigureOpenTelemetryMeterProvider(builder => builder.AddConsoleExporter());
    services.ConfigureOpenTelemetryTracerProvider(builder => builder.AddConsoleExporter());
    services.Configure<OpenTelemetryLoggerOptions>(builder => builder.AddConsoleExporter());
    // In the future the logging API will become:
    // services.ConfigureOpenTelemetryLoggerProvider(builder => builder.AddConsoleExporter());
}
```

Delegates for these kinds of things are executed in order as expected. The way our template for `Program.cs` is setup, the users configuration will be applied before our own (`AddAltinnAppServices`). That is something to think about as some APIs on the builders may have us override something the user configured previously...

I think this is better than returning something from `AddAltinnAppServices`, both because that would make users change code below the `Unless you are sure what you are doing do not change the following code` marker in `Program.cs`, and it leaves us with fewer public APIs to think about (including the `IOpenTelemetryBuilder` that we were a bit worried about). What do you guys think?

ivarne:
I like a simpler and more limited interface without a custom builder (unless we really need it).

> ...as some APIs on the builders may have us override something the user configured previously

Currently we typically use the `TryAdd*` service registration methods, so that it doesn't override previously registered services, so we have kind of a solution for that problem.

martinothamar:
I was referring to the APIs on the OTel builder, e.g. `TraceProviderBuilder.SetSampler`, which currently I think is the only "problematic" API we use. We set the sampler to `AlwaysOnSampler` when running in development environment which overrides whatever may be set in `RegisterCustomAppServices`. Code-based sampling takes precedence over sampling config from `IConfiguration`, so we could maybe set to `always_on` there, or leave it completely to SDK default/app config. So I guess a list of options would be

* Don't configure the sampler in lib at all
* Configure sampling through `IConfiguration` - in memory collection that is overridable from appsettings/env
* Call `AddOpenTelemetry` earlier in the pipeline

Doc on existing sampling in App Insights SDK, adaptive sampling is default: https://learn.microsoft.com/en-us/azure/azure-monitor/app/sampling-classic-api
Defaul in OTel is `ParentBasedSampler(AlwaysOnSampler)`. Reminder for my self: sync with platform team on current sampling in regards to cost. Ideally turning on the OTel feature doesn't increase Azure cost for service owners. Adaptive sampling isn't implemented in the Azure Monitor OTel distribution yet. IMO tail sampling in an agent like OTel collector should also be a good option

HauklandJ:
https://github.com/Altinn/app-lib-dotnet/pull/589/commits/ca80eb99d4412023ae8704b91e93d98135fe6b36 adds a GetSnapshot method to TelemetrySink. This makes adding telemetry testing to existing tests simpler. 
```
SomeTest()
{
  TelemetrySink ts = new();
  ...
  var res = classUnderTest.MethodUnderTest(param_1, ..., param_n, ts);
  ...
  Verify(ts.GetSnapshot());
}
```

ivarne:
> ca80eb9 adds a GetSnapshot method to TelemetrySink. This makes adding telemetry testing to existing tests simpler.

Very nice! For future PR review when we might have a metric removed or added that affects many tests, I think it would be better if you printed them separated by newline so that we get a better git diff. 

Can we assert the values for tags as well? `instance.guid` looks a little strange when it is actually expected to be some known value (as it is in most tests).

HauklandJ:
> For future PR review when we might have a metric removed or added that affects many tests, I think it would be better if you printed them separated by newline so that we get a better git diff.

The Xunit.Verify library generates a diff file when verify is false. Here as an example I altered the expected snapshot to have one less metric and a different guid than the actual. What is then displayed is a nice diff:
image

> Can we assert the values for tags as well? `instance.guid` looks a little strange when it is actually expected to be some known value (as it is in most tests).

Agreed

ivarne:
> What is then displayed is a nice diff:

Your screenshot is cropped so I can't really see if you removed one or more metrics to illustrate how the diff is displayed. If it is common to have more than 3 metrics, I think it makes sense to print them on separate lines to avoid horizontal scrolling.

ivarne:
Nice to see the value! For instance creation tests we can't really hard code the value, so we need some way to ignore it, but not a requirement for this or.

HauklandJ:
> Your screenshot is cropped so I can't really see if you removed one or more metrics to illustrate how the diff is displayed. If it is common to have more than 3 metrics, I think it makes sense to print them on separate lines to avoid horizontal scrolling.

Horizontal scrolling is a good point. I have updated the snapshot creation such that all inserts are on separate lines.

HauklandJ:
>Nice to see the value! For instance creation tests we can't really hard code the value, so we need some way to ignore it, but not a requirement for this or.


Trying to create a return string for snapshot testing was a mistake. Using an object as input is the way to go, an it allows sending a `settings` object as input to the verify method, in which one can ignore values which are never the same. One caveat is that the library tries to be smart with guids. See commit https://github.com/Altinn/app-lib-dotnet/pull/589/commits/97b96ffe3d578f1de871c1a7335d2a1bc7c5af61 where [this verified file](https://github.com/Altinn/app-lib-dotnet/pull/589/commits/97b96ffe3d578f1de871c1a7335d2a1bc7c5af61#diff-d1bd58127596c8b67f74c291c1eaf6ac8a3ab577e1e078d01474cd94ee680255) has stripped the guid as opposed the the string "email" being included in [this verify file](https://github.com/Altinn/app-lib-dotnet/pull/589/commits/97b96ffe3d578f1de871c1a7335d2a1bc7c5af61#diff-56e6a214a6ce618034ffb5e93988c062a83576c04bb7f1e03a82995e550b959b)

ivarne:
> Trying to create a return string for snapshot testing was a mistake. Using an object as input is the way to go, an it allows sending a settings object as input to the verify method

I think we can do much better with custom settings like `.GetSnapshot(ignoredTagValues: ["instance.guid"])`, than a generic `Verify` library can do with settings (the heuristic to just ignore `Guid` is not perfect to say the least). If we add this to many tests it is worth it to make the intention as clear as possible, without copying lots of boilerplate.

HauklandJ:
> I think we can do much better with custom settings like `.GetSnapshot(ignoredTagValues: ["instance.guid"])`, than a generic `Verify` library can do with settings (the heuristic to just ignore `Guid` is not perfect to say the least). If we add this to many tests it is worth it to make the intention as clear as possible, without copying lots of boilerplate.

[Verify.Xunit.Settings](https://github.com/VerifyTests/Verify/blob/main/docs/serializer-settings.md#ignoring-a-type) gives us this. I would advice against reinventing this solution.
The tradeoff is with guids as you point out, however the difference in value for our tests between a value being set to "Guid_1" and being set to "insert guid here" usually boils down to testing that when Guid.Parse("my_guid_to_parse") is done in a test it results in "my_guid_to_parse"

src/Altinn.App.Core/Features/Telemetry.cs
```diff
@@ -0,0 +1,194 @@
...
+        /// Label for the authentication level of the user.
+        /// </summary>
+        public const string UserAuthenticationLevel = "user.authentication_level";
+    }
+
+    internal static class InternalLabels
+    {
+        internal const string Result = "result";
+        internal const string Type = "type";
+        internal const string AuthorizationUserId = "authorization.userid";
```
martinothamar: What makes this different from `Labels.UserId`?

src/Altinn.App.Core/Features/TelemetryExtensions.cs
martinothamar: Tried this, improved some places, worsened others. Keeping as is for now

src/Altinn.App.Core/Altinn.App.Core.csproj
martinothamar: We would have to copy semantic conventions code/config out of otel to do this, I don't think it's worth it

martinothamar:
@ivarne since we've removed `prometheus-net` here, could you remove the subset of changes that affects? Is there any simplification to be made due to this?

martinothamar (approved):
LGTM!

HauklandJ (commented):
lgtmlgtm

PR #590, opened 2024-04-14
Fix a few issues with the new PATCH updates of datamodels

### LINKED ISSUE #517: PATCH: Saving a space and removing it causes 409 crash

### LINKED ISSUE #621: Lokaltest av skjema kræsjer dersom man trykker "enter" flere ganger etter hverandre i en textArea komponent
### Description of the bug

I lokaltest kræsjer skjema dersom man trykker "enter" flere ganger etter hverandre. Samme problem oppstår derimot ikke i tt02.

Her er en video av problemet i lokaltest:
https://github.com/Altinn/app-frontend-react/assets/159779882/ac81161c-8598-4ecd-ae37-9244d5eca0e0

Her er repoet som ble testet i videoen:
https://altinn.studio/repos/ssb/ra1100-01
(autoCreate må settes til "true")

### Steps To Reproduce

1. Last ned repo nevnt i beskrivelsen
2. Sett autoCreate i applicationmetadata.json til "true"
3. Kjør skjema lokalt
4. på nesten hver eneste side i skjemaet er det en textArea komponent. Gå til en side med textArea, trykk deg inn i feltet og trykk på "enter" flere ganger etter hverandre.
5. Resultat: skjemaet kræsjer

### Additional Information

_No response_

ivarne:
Excecute ShouldSerialize* instead of reading attributes

Alternative for https://github.com/Altinn/app-lib-dotnet/pull/582 (the first commit is the same). The difference is how we ensure that the parent object is set to null for objects of the type.

```C#
public class Child
{
    [XmlText]
    public string value { get; set; }

    [XmlAttribute]
    [BindNever]
    public int grid { get; set; } = 123
}
```

https://github.com/Altinn/app-lib-dotnet/pull/582 looks at the [XmlText] and [BindNever] attributes, but this PR looks at the parent object to see if it has a `ShouldSerialize*` method (as is used by xml serialization)

```C#
public class Parent
{
    public Child Child { get; set; }
    
    public bool ShouldSerializeChild() => Child?.value != null
}
```

## Benefits
* Backwards compatible: This just respects a config that always was there.

## Drawbacks
* Studio needs to change model generation in a possibly backwards incompatible way, if their `dataProcessor`/`validator` classes don't check nulls on the wrapper class because they think they never could be null. (now they can)
* Users need to regenerate models if they want the new behaviour (this is required for the `decimal` issue anyway)

## Related Issue(s)
- Alternative for https://github.com/Altinn/app-lib-dotnet/pull/582
- Fixes #517
- Part of a solution for #552 (but studio needs to fix nullable [XmlText] `decimal` and add `ShouldSerialize*` to parent objects)

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Core/Helpers/ObjectUtils.cs
```diff
@@ -30,44 +30,133 @@ public static void InitializeAltinnRowId(object model)
...
+        {
+            // Property has a generic type that is a subtype of List<>
+            if (prop.PropertyType.IsGenericType && prop.PropertyType.GetGenericTypeDefinition() == typeof(List<>))
             {
                 var value = prop.GetValue(model);
                 if (value is null)
                 {
-                    // Initialize IList with null value
+                    // Initialize IList with null value (this is what comes back from xml deserialization)
                     prop.SetValue(model, Activator.CreateInstance(prop.PropertyType));
```
martinothamar, line 111: This is probably besides the point, but I think the first part of this comment is wrong? `Activator.CreateInstance` doesn't return `null` here?

src/Altinn.App.Core/Helpers/ObjectUtils.cs
```diff
@@ -30,44 +30,133 @@ public static void InitializeAltinnRowId(object model)
...
                 {
-                    // Initialize string with null value (xml serialization does not always preserve "")
+                    // Ensure empty strings are set to null
                     prop.SetValue(model, null);
                 }
 
-                // continue recursion over all properties that are not null or value types
-                if (value?.GetType().IsValueType == false)
+                // continue recursion over all properties that are NOT null or value types
+                else if (value.GetType().IsValueType == false)
```
martinothamar: How about `SomeStruct?` types which are really `Nullable<SomeStruct>`. Then `IsValueType = true` and potentially it should be set to null?

src/Altinn.App.Core/Helpers/ObjectUtils.cs
```diff
@@ -30,44 +30,133 @@ public static void InitializeAltinnRowId(object model)
...
 
+    private static void SetToDefaultIfShouldSerializeFalse(object model, PropertyInfo prop, MethodInfo[] methodInfos, object value)
+    {
+        string methodName = $"ShouldSerialize{prop.Name}";
+
+        var shouldSerializeMethod = methodInfos
+            .Where(
+                m => m.Name == methodName
+                     && m.GetParameters().Length == 0
+                     && m.ReturnType == typeof(bool))
```
martinothamar: Should this also check `BindingFlags` to make sure we are getting the instance method? Any constraints around accessibility?

martinothamar (commented):
I'm missing a ton of context here, but threw in some questions atleast 😄 I think I agree on the `BindNever` vs `ShouldSerialize*` approach

> Studio needs to change model generation in a possibly backwards incompatible way

Again I have now context here, should we discuss with the Studio team? Maybe they have ideas

src/Altinn.App.Core/Helpers/ObjectUtils.cs
ivarne: Good catch! Looks like there both private and pulic, instance and static should work for this one, but I'm not sure I found the correct source in the xml libraries of the dotnet runtime.

```
12 results - 6 files

runtime/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/ReflectPropertyDescriptor.cs:
  473  
  474:                     _shouldSerializeMethod = FindMethod(_componentClass, "ShouldSerialize" + Name, args, typeof(bool), publicOnly: false);
  475                      _state[s_bitShouldSerializeQueried] = true;

runtime/src/libraries/System.Management/src/System/Management/WMIGenerator.cs:
  1382                          cmm = new CodeMemberMethod();
  1383:                         cmm.Name = "ShouldSerialize" + PublicProperties[prop.Name].ToString();
  1384                          cmm.Attributes = MemberAttributes.Private | MemberAttributes.Final;

  4110  
  4111:             cmie = new CodeMethodInvokeExpression(new CodePropertyReferenceExpression(new CodeVariableReferenceExpression(contextObject), "PropertyDescriptor"), "ShouldSerializeValue");
  4112              cmie.Parameters.Add(new CodePropertyReferenceExpression(new CodeVariableReferenceExpression(contextObject), "Instance"));

  4175  
  4176:             cmie = new CodeMethodInvokeExpression(new CodePropertyReferenceExpression(new CodeVariableReferenceExpression(contextObject), "PropertyDescriptor"), "ShouldSerializeValue");
  4177              cmie.Parameters.Add(new CodePropertyReferenceExpression(new CodeVariableReferenceExpression(contextObject), "Instance"));

  4207  
  4208:             cmie = new CodeMethodInvokeExpression(new CodePropertyReferenceExpression(new CodeVariableReferenceExpression(contextObject), "PropertyDescriptor"), "ShouldSerializeValue");
  4209              cmie.Parameters.Add(new CodePropertyReferenceExpression(new CodeVariableReferenceExpression(contextObject), "Instance"));

runtime/src/libraries/System.Private.Xml/src/System/Xml/Serialization/Models.cs:
  299              _memberInfo = memberInfo;
  300:             _checkShouldPersistMethodInfo = memberInfo.DeclaringType!.GetMethod($"ShouldSerialize{memberInfo.Name}", Type.EmptyTypes);
  301              _checkShouldPersist = _checkShouldPersistMethodInfo != null;

runtime/src/libraries/System.Private.Xml/src/System/Xml/Serialization/ReflectionXmlSerializationWriter.cs:
  599                      {
  600:                         string methodInvoke = $"ShouldSerialize{m.Name}";
  601                          MethodInfo method = o!.GetType().GetMethod(methodInvoke, BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static | BindingFlags.DeclaredOnly)!;

  631                      {
  632:                         string methodInvoke = $"ShouldSerialize{m.Name}";
  633                          MethodInfo method = o!.GetType().GetMethod(methodInvoke, BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static | BindingFlags.DeclaredOnly)!;

runtime/src/libraries/System.Private.Xml/src/System/Xml/Serialization/XmlSerializationWriter.cs:
  1703                      {
  1704:                         string memberName = $"ShouldSerialize{member.Name}";
  1705                          WriteMethodInfo(typeFullName, typeVariable, memberName, false);

  3190                              Writer.Write("if (");
  3191:                             string methodInvoke = RaCodeGen.GetStringForMethodInvoke("o", fullTypeName, $"ShouldSerialize{m.Name}", mapping.TypeDesc.UseReflection);
  3192                              if (mapping.TypeDesc.UseReflection) methodInvoke = $"(({typeof(bool).FullName}){methodInvoke})";

  3231                          Writer.Write("if (");
  3232:                         string methodInvoke = RaCodeGen.GetStringForMethodInvoke("o", fullTypeName, $"ShouldSerialize{m.Name}", mapping.TypeDesc.UseReflection);
  3233                          if (mapping.TypeDesc.UseReflection) methodInvoke = $"(({typeof(bool).FullName}){methodInvoke})";

runtime/src/libraries/System.Text.Json/tests/System.Text.Json.Tests/Serialization/MetadataTests/DefaultJsonTypeInfoResolverTests.JsonPropertyInfo.cs:
  1091  
  1092:                             Assert.Fail("ShouldSerialize set for value which is not int or string");
  1093                              return false;
```

src/Altinn.App.Core/Helpers/ObjectUtils.cs
ivarne: Can a struct contain a reference to a `string` or `List<>` or references to classes? 
Does struct work with json/xml serialisation? 
If that is possible we might need to add code in order to support people using structs instead of classes for part of their data models.

src/Altinn.App.Core/Helpers/ObjectUtils.cs
ivarne: Yes, It should have been `if it has null value`.

src/Altinn.App.Core/Helpers/ObjectUtils.cs
martinothamar: There are no limitations on what a struct can contain AFAIK. A little mini-test I made locally here makes it look like that is supported, i.e. this works: 
image

src/Altinn.App.Core/Helpers/ObjectUtils.cs
martinothamar: Nice, from what I could tell `ReflectionXmlSerializationWriter` sounds like the most common impl, and that basically any method works as long as it is not inherited (`DeclaredOnly`)? Could probably just copy the bindingflags from there. I couldn't find this documented anywhere heh

src/Altinn.App.Core/Helpers/ObjectUtils.cs
ivarne: I think you are correct then. Any suggestions for how I might solve https://github.com/Altinn/app-lib-dotnet/pull/538 then?

The issue is that `decimal` has cyclic references so just traversing all properties fails.

A recursion counter would fix it, and should probably be combined with providing binding flags to the `GetProperties()` call.

martinothamar: Hmm. 

If we only think about  the `decimal` and `DateTime` case for a second, is using `BindingFlags` enough? I don't quite understand how they lead to recursive behavior here.

If the goal is to more generally avoid infinite recursion, which is probably a good idea (I guess users could define self-referential types themselves as well?), I think we could use something like `HashSet<(obj Instance, string Property)>` to keep track of "visited nodes".

ivarne: ~~It seems to be fixed when only reading instance and public properties. The cyclic reference was in a private or static property.~~ (Typo when testing. The issue is with DateTime having an instance property to get the `Date` or `Time` part and that can be applied infinite number of times.)

I opted for a simple depth counter to avoid infinite recursion. I used the same recursion limit as System.Text.Json at 64 levels. A hash set will obviously fail faster, but also cary a bigger memory footprint for the common case without cyclic references. Isn't this one of the cases when you need a WeakHashSet?

src/Altinn.App.Core/Helpers/ObjectUtils.cs
ivarne: It works with inheritance also, but the `$"ShouldSerialize{prop.Name}"` must be in the same class as the property it wants to control serialisation of. I don't think it is worth adding that extra indirection check in our copy of the code. System.Xml caches this calculation, so they don't pay the penalty every execution.

src/Altinn.App.Core/Helpers/ObjectUtils.cs
ivarne: Newest version ignores all types in the `System` namespace for any recursion and have a recursion depth limit that causes a readable exception instead of a stack overflow. A hash set will only detect cyclic references, but as C# has computed properties they can just create infinite new references at runtime anyway.

If user defined data model types needs this kind of infinite recursive properties, they will have an issue with json serialisation. Maybe we could respect `[JsonIgnore]` or something like that?

martinothamar: Right, seems like S.T.J special cases primitives with handwritten converters (`JsonPrimitiveConverter`), so I guess that combined with the max depth solves it there. I like the `System` prefix solution 👍

martinothamar: > Maybe we could respect [JsonIgnore] or something like that?

How about `[XmlIgnore]` as well in that case? Using the same type for both XML and JSON, is the expectation from users that they always use matching sets of attributes where applicable? As in, if you use `JsonIgnore`, you must also use `XmlIgnore`?

martinothamar (approved):

src/Altinn.App.Core/Helpers/ObjectUtils.cs
ivarne: I think the `System` prefix and an obvious Exception on infinite recursive types is the best we can do. Adding an option to have ignored properties with infinite recursion seems to be over engineering that I assume nobody will ever need.

ivarne:
@martinothamar Would you take a second look at this (especially the changes for https://github.com/Altinn/app-lib-dotnet/issues/621 to ensure that xml serialization on windows does not insert `\r` before newlines), close the treads (if you agree) and approve again?

Note:
I went with `NewLineHandling.None` instead of the arguably better `NewLineHandling.Entitize` (escape newlines as in json) to limit backwards compatibility issues for receiving systems.

martinothamar (approved):
LGTM

PR #591, opened 2024-04-15
Improve API for getting JSON options that ignore prefixes during serialization

martinothamar:
## Description
During #583 I tried to make all `JsonSerializerOptions` instances static. There is one place left after that where instances aren't cached, which is `IgnorePropertiesWithPrefix` along with the associated test. This PR fixes that by caching instances in a dictionary

Considerations
* If consumers require differing options set on `JsonSerializerOptions` (apart from the modifier), that would have to be part of the cache key
* Could have chosen to go with `ConcurrentDictionary`, but that doesn't provide mutual exclusion for instantiating the options to cache
* The cache is unbounded and items never expire, could use a `MemoryCache` instead, depends on expected invocation pattern. Same problem as above in terms instantiation

## Related Issue(s)
- N/A

## Verification
- [x] All tests run green

## Documentation


martinothamar:
I like that this manages the lifecycle of the options for you (even though that makes configuration kind of limited/bad...). But in the end this feels like a silly API to have public at all IMO

ivarne (approved):
Functionally this looks like it will work fine and , but a little nitpick on the API is that I don't understand why it makes sense for `JsonHelper` to export a public function that returns `JsonSerializerOptions`. Wouldn't it make more sense to have `JsonHelper.SerializeIgnorePrefix(object, prefix)` that returns the serialised data as a string.

martinothamar:
I'm fine with that as well, maybe it belongs alongside `JsonSerializerPermissive` instead then? `JsonSerializerIgnorePrefix`. Another option is perhaps to just not offer this as a public API, only obsolete the old one with a message saying to implement it themselves?

I searched through public altinn repos and couldn't find a single use, not from other projects on Github either (studio etc).

I think I like this best:
* Implement `SerializeIgnorePrefix` as an internal function
* Obsolete the public type with a message saying to implement it themselves

ivarne:
I agree this is a great candidate for functionality that can be `internal`.

HauklandJ (approved):

PR #594, opened 2024-04-16
Fix issue where AppInsights SDK doesn't load/configure correctly due to keyed services being registered in DI

martinothamar:
## Description
A bug report came in, where logs/telemetry wasn't being shipped to AppInsights.
Upon inspecting the `IServiceCollection` registrations after calling `AddApplicationInsightsTelemetry`, we discovered that everything until (and including) these lines of registrations succeeded:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/4626e3632323b741a2bd56d3d529c283cf7626c9/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Extensions/ApplicationInsightsExtensions.cs#L134

However the `TelemetryClient` itself was never registered in DI. In the code linked above, if there is an exception, it is (relatively silently) swallowed and registered on the event source for AppInsights.

We traced the event source in a local test app:
```console
dotnet trace collect --providers Microsoft-ApplicationInsights-AspNetCore -- ./bin/Debug/net8.0/Altinn.App
```

Which could be loaded into PerfView

image

We see this error:

```
HasStack="True" ThreadID="2,558,902" ProcessorNumber="0" errorMessage="System.InvalidOperationException: This service descriptor is keyed. Your service provider may not support keyed services. at Microsoft.Extensions.DependencyInjection.ServiceDescriptor.ThrowKeyedDescriptor() 
at Microsoft.Extensions.DependencyInjection.ServiceDescriptor.get_ImplementationFactory() 
at Microsoft.Extensions.DependencyInjection.ApplicationInsightsExtensions.<>c__28`2.<AddSingletonIfNotExists>b__28_0(ServiceDescriptor o) 
at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate) 
at Microsoft.Extensions.DependencyInjection.ApplicationInsightsExtensions.AddSingletonIfNotExists[TService,TImplementation](IServiceCollection services) 
at Microsoft.Extensions.DependencyInjection.ApplicationInsightsExtensions.AddCommonTelemetryModules(IServiceCollection services) at Microsoft.Extensions.DependencyInjection.ApplicationInsightsExtensions.AddApplicationInsightsTelemetry(IServiceCollection services)" appDomainName="Altinn.App" 
```

Which means that AppInsights thinks that our DI container (Scrutor) doesn't support keyed services. I think that's right? Since there is an open issue on that: https://github.com/khellang/Scrutor/pull/209

This PR replaces keyed services by just injecting multiple services, and introducing a second marker interface to distinguish between them instead.

Other options:
* Use a different DI container (for example the built in one), which supports keyed services right now?
* ?

## Related Issue(s)
- N/A

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Altinn.App.Api.csproj
+++ b/src/Altinn.App.Api/Altinn.App.Api.csproj
@@ -48,4 +48,9 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <NoWarn>$(NoWarn);1591</NoWarn>
   </PropertyGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="$(AssemblyName).Tests" />
+    <InternalsVisibleTo Include="DynamicProxyGenAssembly2" />
+  </ItemGroup>
 </Project>
--- a/src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
@@ -73,7 +73,13 @@ namespace Altinn.App.Api.Extensions
             services.AddMetricsServer(config);
         }
 
-        private static void AddApplicationInsights(IServiceCollection services, IConfiguration config, IWebHostEnvironment env)
+        /// <summary>
+        /// Adds Application Insights to the service collection.
+        /// </summary>
+        /// <param name="services">Services</param>
+        /// <param name="config">Config</param>
+        /// <param name="env">Environment</param>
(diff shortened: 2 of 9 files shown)


martinothamar:
We can probably also just register AppInsights earlier in the DI container, since the issue is provoked when AppInsights scans the registry.. https://github.com/microsoft/ApplicationInsights-dotnet/blob/4626e3632323b741a2bd56d3d529c283cf7626c9/NETCORE/src/Shared/Extensions/ApplicationInsightsExtensions.cs#L297
But that would feel less solid

HauklandJ (approved):

Issue #460, opened 2024-02-27
Signatures are not deleted when process is reversed


ivarne:
## Description of the bug

When an app has two signature steps and the second one is refused, the old first signature object stays and the result is 3 instead of 2 signature objects, one of which is invalid

## Discussion
Discussion and proposed new feature is documented in #617

tjololo:
Implementing this feature request in Storage will resolve this issue: https://github.com/Altinn/altinn-storage/issues/351

bjorntore:
Am I reading the code correctly: We will delete any data element where a reference exists with RelationType="GeneratedFrom" and ValueType="Task" when the process task it was connected to is "ended"? See https://github.com/Altinn/app-lib-dotnet/blob/main/src/Altinn.App.Core/Internal/Process/ProcessTasks/Common/ProcessTaskFinalizer.cs#L60.

Is that not a bit aggressive?

Maybe we instead should configure on the data type if it should be auto deleted when the process step is ended/abandoned?

If we want to always delete data elements with such a relation if "reject" is called on a task, maybe we should stick it in AbandonTaskEventHandler?

For Signature specifically, it could also have been solved by implementing this in the SigningProcessTask.Abandon method.

ivarne:
> For Signature specifically, it could also have been solved by implementing this in the SigningProcessTask.Abandon method.

No, that will not solve the issue when the signature task is not the last task in the process, and someone wants to go back past a previous signature task to change something.

bjorntore:
> > For Signature specifically, it could also have been solved by implementing this in the SigningProcessTask.Abandon method.
> 
> No, that will not solve the issue when the signature task is not the last task in the process, and someone wants to go back past a previous signature task to change something.

I should have written "maybe". :P

ivarne:
I have previously asked the exact same question

danielskovli:
This has been implemented in https://github.com/Altinn/app-lib-dotnet/pull/617 and merged to main.

PR #617, opened 2024-04-25
Delete signatures if task process is reversed

danielskovli:
## Description
If a signature step is cancelled/abandoned, eg. the user navigates backwards in the flow and/or modifies the signed data, the signature should be invalidated (deleted).

In practice, this is not very straightforward to implement because we must ensure that the user perfoming the cleanup-action has the appropriate rights for the storage claim. 

Ideally the cleanup should occur during `Initialize` for each task, meaning any and all tasks would be able to dispose of stale data on their own. This would take place by simply deleting all associated `DataElements` before proceeding to generate new ones.

Because of the challenge regarding user permissions, this patch offers a compromise. Cleanup is performed by `ProcessEngine.Next`, immediately before calling `HandleAction` on the request. At this point the user permissions have been validated, because the user is actually on their way _out_ of a task, but crucially the event takes place _before_ new data (signatures, pdf) is generated.

This feels a bit counter intuitive, and isn't the ideal approach (as outlined above), but the result is mostly the same as performing cleanup during `Initialize`.

## Edge cases
- A user completes a signature task, before returning to a previous step where they alter some data. Instead of completing the signature step again, they abandon the process. The stale signatures now still persist in datastorage. But the instance is/should be marked as incomplete, so does this matter? The concern here is that the generated signature/pdf gets used somehow and that the consumer does not validate the signature(s).

## Test setup
- Modified [app-localtest](https://github.com/Altinn/app-localtest) setup as per this [altinn-storage patch](https://github.com/Altinn/altinn-storage/pull/392/files)
- [A test app that's capable of rejecting a signature step](https://dev.altinn.studio/repos/danielskovli/signering-ds), running with a local version of app-lib and the [code contained in this PR](https://github.com/Altinn/app-lib-dotnet/tree/feat/460-delete-signatures-on-process-reversed)
- **Verify:** Signing and rejecting several times produces exactly one signature [with modified app-localtest], as opposed to multiple signatures [original app-localtest]

## Test results
✅ After the update of [package Altinn.Platform.Storage.Interface to v3.25.0](https://github.com/Altinn/altinn-storage/releases/tag/Storage.Interface-3.25.0) it has been confirmed that existing behavior with current altinn-storage service is preserved, and new desired behavior is achieved with a [proposed change in altinn-storage](https://github.com/Altinn/altinn-storage/pull/404).

## Related Issue(s)
- #460
- Altinn/altinn-storage#351

## Verification
- [x] All tests run green

## Documentation


danielskovli:
Reverts initial scaffolding in light of https://github.com/Altinn/altinn-storage/pull/392#discussion_r1580581687

src/Altinn.App.Core/Internal/Process/ProcessTasks/Common/ProcessTaskInitializer.cs
```diff
@@ -137,4 +138,41 @@ private async Task UpdateDataValuesOnInstance(Instance instance, string dataType
...
+        AppIdentifier appIdentifier = new(instance.AppId);
+        InstanceIdentifier instanceIdentifier = new(instance);
+        var dataElements =
+            instance
+                .Data?.Where(de =>
+                    de.References?.Exists(r => r.ValueType == ReferenceType.Task && r.Value == taskId) is true
+                )
+                .ToList() ?? [];
+
+        _logger.LogInformation("Found {} stale data element(s) to delete", dataElements.Count);
```
martinothamar:
```suggestion
        _logger.LogInformation("Found {Count} stale data element(s) to delete", dataElements.Count);
```
Message template properties should be named, and convention is PascalCase for these

danielskovli:
Original concept of `SignRequest.TaskId` was reimplemented as `SignRequest.GeneratedFromTask` after various internal discussions related to compatibility between app-lib versions and altinn-storage versions.

bjorntore (approved):

Issue #600, opened 2024-04-18
Layout & Expression changes due to supporting multiple data models


bjosttveit:
### Description

Supporting multiple data models in app frontend should mostly work with app-lib-dotnet out-of-the-box, as it will fetch and update each data model independently.

Where things could (and probably will) go wrong is where app-lib does its own layout parsing and expression evaluation. The relevant changes are that `dataModelBindings` now optionally can be defined using an object specifying both the path, and the dataType, like this:
```json
"dataModelBindings": {
  "simpleBinding": {
    "property": "some.field",
    "dataType": "dataModelName"
  }
}
```

Additionally, to support multiple data models in `dataModel` expressions, a third optional parameter has been added, making the following valid:
```json
["dataModel", "some.field", "dataModelName"]
```

`component` expressions will also be able to reference any data model through its `dataModelBindings`, but the configuration there has not changed.

Meaning that in order to support this, it is no longer enough to have the "default" data model (defined in layout-sets) available, as the layout and expressions can reference any data model defined in the app.

The `ExpressionValidator` kind of already supports this, insofar as they are already defined on a per-data-model basis. However, it currently still needs to be able to ignore hidden data, and therefore still relies on being able to parse the layout and run all expressions.

There are no breaking changes in the config, meaning that existing apps should work in the same way as before. However, when actually trying to use and reference multiple data models I suspect that this will cause the app backend to crash when it tries to parse the layout and run expressions.

App-Frontend issue [#1709](https://github.com/Altinn/app-lib-dotnet/issues/1157) again comes to mind, as offloading the work of removing hidden data, and updating the backend validators accordingly would go a long way to resolving these issues. We could even say that referencing other data models in expression validation is not allowed, meaning we would not need to fetch potentially all data models when validating a single data model with expression validation.

The other option is of course updating the logic in app-lib to support the extended `dataModelBindings` functionality, and making several / all data models available when running expressions there.

### In scope
* Reference current sub-form/datatype ie. expressions that points to components within the same layout.
* Reference other datatype that has 1 as it's maxCount

### Out of scope
* Reference sub-forms (underskjema) ie. datatypes with more than 1 as maxCount.

### Additional Information
* This should be implemented as a opt-in solution leaving existing implementations working as-is.
* This will today be triggered by enabling one of the following (which are false by default):
	* Required
	* ExpressionValidation
	* RemoveHiddenData

The code where "things happen" are:
	* LayoutEvaluator
[...]

PR #619, opened 2024-04-27
Support backend validation for multiple data models

ivarne:
New type `record struct ModelBinding` to replace the current use of `string` in a way that ensures a structured way of storing the optional reference to an extra data model.

```C#
public readonly record struct ModelBinding
{
    public ModelBinding()
    {
        DataType = "default";
    }

    public required string Field { get; init; }
    public string DataType { get; init; }

    // Consider removing the implicit operator. It makes the existing tests work with much less changes.
    public static implicit operator ModelBinding(string field)
    {
        return new ModelBinding { Field = field };
    }
}
```

I'm pretty happy about the minimal changes required for everything but the actual lookup of data to work.

## Remaining work
* Make the `IDataModel` implementations aware of multiple data models so that lookups don't just ignore them.
    * Decide on a caching approach. Should we use a caching version `IDataClient` registered as `scoped` in DI? Or do we have other options?

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/600

## Verification
- [ ] All tests run green

## Documentation


martinothamar:
Do you want me to review the whole thing or just the `ModelBinding` struct for now? 

Would `ModelBindingSelector` be a better name? To me that sounds more specific and clear about what it does/is used for, in that it represents the selection/query into the datamodel. I think the structure is good. It doesn't really validate the inputs, so in terms of correctness I don't think it improves the "primitive obsession" situation that we have (I don't think there are any good ways to actually do that well atm anyway)

I think we have to have the implicit operator since we're doing `string` -> `ModelBinding` change on `IDataModelAccessor`? Re not having breaking changes

ivarne:
> Do you want me to review the whole thing or just the ModelBinding struct for now?

I want any feedback I can get, apart from "Hi! I made a test app, how do I test this new behavior", because it isn't complete yet.

src/Altinn.App.Core/Helpers/DataModel/DataModel.cs
```diff
@@ -94,15 +95,15 @@ public DataModel(object serviceModel)
     }
 
     /// <inheritdoc />
-    public string[] GetResolvedKeys(string key)
+    public ModelBinding[] GetResolvedKeys(ModelBinding key)
     {
         if (_serviceModel is null)
         {
-            return new string[0];
+            return Array.Empty<ModelBinding>();
```
martinothamar: nit: `[]` lowers to the same code

src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
```diff
@@ -105,7 +106,23 @@ bool defaultReturn
...
+        {
+            [string field] => new ModelBinding { Field = field },
+            [string field, string dataType] => new ModelBinding { Field = field, DataType = dataType },
+            [ModelBinding binding] => binding,
+            [null] => throw new ExpressionEvaluatorTypeErrorException("Cannot lookup dataModel null"),
+            _
+                => throw new ExpressionEvaluatorTypeErrorException(
+                    $"""Expected ["dataModel", ...] to have 1-2 argument(s), got {args.Length}"""
+                )
+        };
```
martinothamar, line 116: nice 😎

src/Altinn.App.Core/Internal/Expressions/LayoutEvaluatorState.cs
```diff
@@ -311,7 +311,8 @@ private void GetModelErrorsForExpression(Expression? expr, BaseComponent compone
                 );
                 return;
             }
-            if (!_dataModel.VerifyKey(binding))
+            var dataType = expr.Args.ElementAtOrDefault(1)?.Value as string;
+            if (!_dataModel.VerifyKey(new ModelBinding { Field = binding, DataType = dataType }))
```
martinothamar, line 320: NRT warning on `dataType`. What happens if `dataType` is different from the `_dataModel._serviceModel` instance here? Seems like there's an assumption that they should always be matching ("*" or type ID). Should `DataModel` always verify the that input `ModelBinding.DataType`? Should `DataType` be nullable? It's unclear to me what the invariants should be here

src/Altinn.App.Core/Models/Layout/ModelBinding.cs
```diff
@@ -0,0 +1,17 @@
+namespace Altinn.App.Core.Models.Layout;
+
+public readonly record struct ModelBinding
+{
+    public ModelBinding()
+    {
+        DataType = "default";
```
martinothamar: What happens if someone actually `default` as a datatype ID? Maybe we should have some other marker for this, maybe just have `DataType` be nullable?

src/Altinn.App.Core/Models/Layout/PageComponentConverter.cs
```diff
@@ -446,6 +446,34 @@ private BaseComponent ReadComponent(ref Utf8JsonReader reader, JsonSerializerOpt
...
+            if (reader.TokenType != JsonTokenType.PropertyName)
+            {
+                throw new JsonException();
+            }
+
+            var propertyName = reader.GetString()!;
+            reader.Read();
+            modelBindings[propertyName] = reader.TokenType switch
+            {
+                JsonTokenType.String => new ModelBinding { Field = reader.GetString()! },
```
martinothamar: IMO we shouldn't use null-forgiving operator this much, because it's unclear why it should/could be null (and generally detracts from the point of NRT). Some options I could think of

* Make sure we use nullability annotations and null-checks (and related attributes (NotNullWhen etc)) correctly
* throw an exception with a better message as opposed to getting nullref
* use `Debug.Assert` as a last resort if we can't convince the compiler something is nonnull (but should always be nonnull)

src/Altinn.App.Core/Helpers/IDataModel.cs
```diff
@@ -19,17 +16,17 @@ public interface IDataModelAccessor
     /// "Bedrifter[1].Ansatte.Alder", will fail, because the indicies will be reset
     /// after an inline index is used
     /// </remarks>
-    object? GetModelData(string key, ReadOnlySpan<int> indicies = default);
+    object? GetModelData(ModelBinding key, ReadOnlySpan<int> indicies = default);
```
martinothamar: Another tricky one in terms of not making breaking changes. Someone could have implemented this interface and will get a broken build for this. I don't think anyone has though (couldn't find anything when searching)

src/Altinn.App.Core/Helpers/DataModel/DataModel.cs
ivarne: Nice, I’ve not (yet) fully embraced that syntax, but it is growing on me.

There is a little bit of a setback every time someone or a linter suggests not using `var`

src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
ivarne: Pattern matching is sometimes both concise and readable 🥳

src/Altinn.App.Core/Models/Layout/PageComponentConverter.cs
ivarne: Totally agree, but in this case it’s the Utf8JsonReader that have an api that can’t be properly annotated, so there is not much we can do.

src/Altinn.App.Core/Helpers/IDataModel.cs
ivarne: 🤫🤫🤫

src/Altinn.App.Core/Models/Layout/ModelBinding.cs
ivarne: I went back and forth on this issue 10 times while writing this, but a naming collision seems like something I should’ve made impossible.

src/Altinn.App.Core/Internal/Expressions/LayoutEvaluatorState.cs
ivarne: This part of the code is only in preparation for proper backend validation of json files, and sadly it is not yet used

ivarne: DataType is nullable now, so this warning is gone

ivarne:
@martinothamar @bjosttveit

Da virker alle eksisterende tester og det er egentlig klart for review. Har ikke begynt å skrive tester for den nye funksjonaliteten ennå.

Har inkludert `["language"]` expression funksjonen, fordi den krever mer context var naturlig å legge med når jeg uansett måtte endre `LayoutEvaluatorStateInitializer.Init` funksjonen for å støtte flere modeller.

src/Altinn.App.Core/Features/Validation/Default/ExpressionValidator.cs
```diff
@@ -92,7 +95,7 @@ IAppMetadata appMetadata
                     rowIndices: DataModel.GetRowIndices(resolvedField),
                     rowLength: null
                 );
-                var positionalArguments = new[] { resolvedField };
+                var positionalArguments = new object[] { resolvedField };
```
bjosttveit, line 98: In app-frontend, I added the dataTypeId as a second positional argument to be able to make generic validation definitions work with multiple data models.

https://github.com/Altinn/app-frontend-react/blob/092186ddde46021a89a8943761ffab8abda811b8/src/features/validation/expressionValidation/ExpressionValidation.tsx#L60

Then you can do this to refer to the "current" field being validated like this:
image

ivarne: Not sure I fully understand, but is this what you look for?
```suggestion
                var positionalArguments = new object[] { resolvedField, baseField.DataType };
```
or

```suggestion
                var positionalArguments = new object[] { resolvedField, dataElement.DataType };
```

ivarne: How have you extended validation to support more data models? They are still in separate files per data model, which makes the data model in argv implicit from the filename?

Found a bug in the current implementation, because I don't traverse `*validation.json` to ensure that I prefetch all references to external models. This means that lookups in types not referenced in the layout will read `null` even if there is a value

bjosttveit: Yeah we could say that the data type should be implied by the file. I wasn't sure if we should do that or not though, since there would be a different "default" than in the layout files. But if we agree on that then I will change it in app-frontend as well.

ivarne: Isn't the premise for expression validation that it is different from layout, because it is tied to the data type and not the component?

bjosttveit: Absolutely, as long as we are on the same page between backend and frontend

src/Altinn.App.Core/Internal/Expressions/ILayoutEvaluatorStateInitializer.cs
```diff
@@ -0,0 +1,21 @@
+using Altinn.Platform.Storage.Interface.Models;
+
+namespace Altinn.App.Core.Internal.Expressions;
+
+/// <summary>
+/// Interface for initializing a <see cref="LayoutEvaluatorState" /> from dependency injection services
+/// </summary>
+public interface ILayoutEvaluatorStateInitializer
```
martinothamar, line 8: Why are we making this a public interface? Do we expect users to make use of these? I tried to find usages of `LayoutEvaluatorStateInitializer` as well and couldn't find any

src/Altinn.App.Core/Models/Expressions/Expression.cs
```diff
@@ -10,23 +11,52 @@ namespace Altinn.App.Core.Models.Expressions;
 /// All props are marked as nullable, but a valid instance has either <see cref="Function" /> and <see cref="Args" /> or <see cref="Value" />
 /// </remarks>
 [JsonConverter(typeof(ExpressionConverter))]
-public sealed class Expression
+public record struct Expression
```
martinothamar: nice, could be `readonly` as well

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
```diff
@@ -7,20 +7,20 @@ namespace Altinn.App.Core.Models.Layout.Components;
...
         string type,
-        IReadOnlyDictionary<string, string>? dataModelBindings,
-        IEnumerable<BaseComponent> children,
-        IEnumerable<string>? childIDs,
-        Expression? hidden,
-        Expression? required,
-        Expression? readOnly,
+        IReadOnlyDictionary<string, ModelBinding>? dataModelBindings,
+        IReadOnlyCollection<BaseComponent> children,
+        IReadOnlyCollection<string>? childIDs,
```
martinothamar, line 20: When do we use `IReadOnlyCollection` as opposed to `IEnumerable`?

src/Altinn.App.Core/Internal/Expressions/LayoutEvaluatorStateInitializer.cs
```diff
@@ -9,34 +11,91 @@ namespace Altinn.App.Core.Internal.Expressions;
...
+        _dataAccessor = dataAccessor;
         _frontEndSettings = frontEndSettings.Value;
     }
 
     /// <summary>
     /// Initialize LayoutEvaluatorState with given Instance, data object and layoutSetId
     /// </summary>
-    public virtual Task<LayoutEvaluatorState> Init(
+    [Obsolete("Use the overload with ILayoutEvaluatorStateInitializer instead")]
+    public Task<LayoutEvaluatorState> Init(
```
martinothamar, line 39: Not virtual any more, so noone implements this?

src/Altinn.App.Core/Helpers/DataModel/DataModel.cs
```diff
@@ -128,16 +155,17 @@ internal static string JoinFieldKeyParts(string? currentKey, string? key)
     /// <summary>
     /// Get the row indices from a key
     /// </summary>
-    public static int[]? GetRowIndices(string key)
+    public static int[]? GetRowIndices(ModelBinding key)
```
martinothamar, line 176: In this case we don't actually need the whole key (and its public), but maybe it's worth it for the consistency

src/Altinn.App.Core/Models/Layout/Components/BaseComponent.cs
```diff
@@ -11,24 +11,24 @@ namespace Altinn.App.Core.Models.Layout.Components;
 /// Includes <see cref="DataModelBindings" /> that will be initialized to an empty dictionary
 /// for components that don't have them.
 /// </remarks>
-public class BaseComponent
+public record BaseComponent
```
martinothamar, line 14: I'm assuming we don't rely on reference equality anywhere?

src/Altinn.App.Core/Models/Layout/LayoutModel.cs
```diff
@@ -54,4 +52,42 @@ public IEnumerable<BaseComponent> GetComponents()
                     nodes.Push(n);
         }
     }
+
+    /// <summary>
+    /// Get all external model references used in the layout model
+    /// </summary>
+    public IEnumerable<string> GetExternalModelReferences()
```
martinothamar: Is there a good use-case for this being exposed public?

src/Altinn.App.Core/Models/Layout/PageComponentConverter.cs
```diff
@@ -467,8 +474,45 @@ private BaseComponent ReadComponent(ref Utf8JsonReader reader, JsonSerializerOpt
...
+        if (reader.TokenType != JsonTokenType.StartObject)
+        {
+            throw new JsonException("Expected StartObject token for \"dataModelBindings\"");
+        }
+
+        while (reader.Read() && reader.TokenType != JsonTokenType.EndObject)
+        {
+            if (reader.TokenType != JsonTokenType.PropertyName)
+            {
+                throw new JsonException();
```
martinothamar, line 504: Could we add a message to these exceptions?

src/Altinn.App.Core/Models/Layout/PageComponentConverter.cs
```diff
@@ -467,8 +474,45 @@ private BaseComponent ReadComponent(ref Utf8JsonReader reader, JsonSerializerOpt
...
+            }
+
+            // ! Token type is PropertyName so value is a string
+            var propertyName = reader.GetString()!;
```
martinothamar, line 508: Throw an exception here as well? Since we do that below

src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -300,10 +302,10 @@ private static void AddAppOptions(IServiceCollection services)
 
     private static void AddProcessServices(IServiceCollection services)
     {
-        services.TryAddTransient<IProcessEngine, ProcessEngine>();
+        services.TryAddScoped<IProcessEngine, ProcessEngine>();
```
martinothamar, line 327: These are potentially breaking as well right? I tried to search some usage but hopefully this is always created from within a scope

src/Altinn.App.Core/Helpers/DataModel/DataModel.cs
```diff
@@ -398,11 +423,14 @@ public void RemoveField(string key, RowRemovalOption rowRemovalOption)
...
     /// <inheritdoc />
-    public bool VerifyKey(string key)
+    public bool VerifyKey(ModelBinding key)
     {
-        return VerifyKeyRecursive(key.Split('.'), 0, _serviceModel.GetType());
+        return VerifyKeyRecursive(key.Field.Split('.'), 0, ServiceModel(key).GetType());
     }
 
+    /// <inheritdoc />
+    public DataElement DefaultDataElement { get; }
```
martinothamar, line 477: What's the usecase for this being public?

src/Altinn.App.Core/Internal/Data/CachedFormDataAccessor.cs
```diff
@@ -0,0 +1,125 @@
...
+using Altinn.Platform.Storage.Interface.Models;
+
+namespace Altinn.App.Core.Internal.Data;
+
+/// <summary>
+/// Class that caches form data to avoid multiple calls to the data service for a single validation
+///
+/// Must be registered as a scoped service in DI container
+/// </summary>
+internal class CachedFormDataAccessor : ICachedFormDataAccessor
```
martinothamar: Could make this `sealed` as well

src/Altinn.App.Core/Internal/Data/ICachedFormDataAccessor.cs
```diff
@@ -0,0 +1,21 @@
+using Altinn.Platform.Storage.Interface.Models;
+
+namespace Altinn.App.Core.Internal.Data;
+
+/// <summary>
+/// Use this in your validators, dataProcessors to get form data from the cache
+///
+/// Note that this is a scoped service and can't be used in singleton or transient services
+/// </summary>
+public interface ICachedFormDataAccessor
```
martinothamar, line 10: Should this be in some other folder istead? We've been putting most stuff in the `Features` folder. Since we encourage use of this in user code I don't think it should be in `Internal` folder or namespace

src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
```diff
@@ -22,18 +23,15 @@ bool defaultReturn
...
+            ArgumentNullException.ThrowIfNull(context.Component);
             var expr = property switch
             {
-                "hidden" => context.Component?.Hidden,
-                "hiddenRow"
-                    => context.Component is RepeatingGroupComponent repeatingGroup ? repeatingGroup.HiddenRow : null,
-                "required" => context.Component?.Required,
+                "hidden" => context.Component.Hidden,
+                "hiddenRow" when context.Component is RepeatingGroupComponent repeatingGroup
+                    => repeatingGroup.HiddenRow,
```
martinothamar, line 31: This is a change of behavior right? Before you would get null if there was a `hiddenRow` property on a component that was not a `RepeatingGroup`, but now we will end up with the exception in the fallthrough branch?

martinothamar (commented):
Made one pass, will make another once tests come in 😄

src/Altinn.App.Core/Models/Layout/Components/GridComponent.cs
ivarne: I used it to avoid warnings for multiple enumeration.

src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
ivarne: Yes, but I see it as highly unlikely that it actually causes a problem.

src/Altinn.App.Core/Features/Validation/Default/ExpressionValidator.cs
ivarne: So, do we add `["argv", 1]` or not. I'm fine either way, but we need to do the same.

src/Altinn.App.Core/Helpers/DataModel/DataModel.cs
ivarne: I think it is worth it for consistency. Users will have to adapt for all the other methods that changed, and if we release with the implicit cast this will just provide more options.

src/Altinn.App.Core/Helpers/DataModel/DataModel.cs
ivarne: Nothing! (at least that I can think of)

I needed it to get the data element for `ValidationIssue`, but it did not solve the full issue, so I added [`LayoutEvaluatorState.GetDataElement(ModelBinding field)`](https://github.com/Altinn/app-lib-dotnet/blob/f7f1687ce82fe5990ac7a1b7cc3df492289412b7/src/Altinn.App.Core/Internal/Expressions/LayoutEvaluatorState.cs#L384-L392) instead. I should probably just move that implementation into the data model and not rely on the `_instanceContext`

src/Altinn.App.Core/Internal/Data/CachedFormDataAccessor.cs
ivarne: Done

src/Altinn.App.Core/Internal/Data/ICachedFormDataAccessor.cs
ivarne: Makes sense, but this should be in the same namespace as `IDataClient` and `IDataService`. I made a v9 issue for that https://github.com/Altinn/app-lib-dotnet/issues/677

src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
ivarne: Impressive catch! I think this behaviour is better, and the diff will only crash very specific and likely erroneous code.

src/Altinn.App.Core/Internal/Expressions/ILayoutEvaluatorStateInitializer.cs
ivarne: My intended way to interact with the layout / expression machinery is to inject `(I)LayoutEvaluatorStateInitializer` to initialise a `LayoutEvaluatorState` and pass it to one of the static methods in `LayoutEvaluator`. This has been used a lot by SSB to run `RemoveHiddenData` before doing calculations.

src/Altinn.App.Core/Internal/Expressions/LayoutEvaluatorStateInitializer.cs
ivarne: It was virtual just for tests. I don't see any reason why someone would override this.

src/Altinn.App.Core/Models/Expressions/Expression.cs
ivarne: I didn't know a class/record could be readonly. Just that I could remove the` {set; }` implementations to make it immutable. Fixed

src/Altinn.App.Core/Models/Layout/Components/BaseComponent.cs
ivarne: I'm not sure I understand. Wouldn't reference equality still work?

The question is if we have use cases for value equality or if I just made it a record because someone on Youtube said it.

src/Altinn.App.Core/Models/Layout/LayoutModel.cs
ivarne: There wasn't a good use case for exposing anything of this layout /expressions tools, until someone wanted to do something this code can do. `ILayoutStateInitializer` is likely the only place this will be used.

src/Altinn.App.Core/Models/Layout/PageComponentConverter.cs
ivarne: You can have exceptions with messages if you suggest a helpful text that explains that despite the fact that `Utf8JsonReader` promises to throw itself on invalid json, but didn't.

src/Altinn.App.Core/Features/Validation/Default/ExpressionValidator.cs
bjosttveit: I will make the default one follow the file, so I suppose it is not necessary to do both?

bjosttveit: I have removed argv 1 now

src/Altinn.App.Core/Models/Layout/Components/BaseComponent.cs
martinothamar: Manually calling `ReferenceEquals` still works, but I was referring to the default `Equals` and `==` stuff that this changes. So it was just to make sure we don't use this as a key in a dictionary or something. I'm guessing we don't, and that we don't expect users to rely on this either

martinothamar (approved):
Generally LGTM 👍 

In terms of `public` vs `internal` - I don't think we should add more public members unless we have an agreed-upon use-case with consumers already. Didn't seem like that was the case, so according to previous discussions I think those should be made internal. If not, I think we should revisit the decision in the team, so we have shared understanding

src/Altinn.App.Core/Models/Layout/ModelBinding.cs
```diff
@@ -0,0 +1,26 @@
...
+
+/// <summary>
+/// Wrapper type for a model binding with optional data type specification
+/// </summary>
+public readonly record struct ModelBinding
+{
+    /// <summary>
+    /// The field in the model the binding is for
+    /// </summary>
+    public required string Field { get; init; }
```
bjosttveit, line 14: I have called it `property` in app-frontend, and so it is currently unable to deserialize, so I guess we need to agree on what this should be called as well 😅

ivarne: ValidationIssue uses `Field`, so we use that.
https://github.com/Altinn/app-lib-dotnet/blob/06f0e26a11dca56d0ec784bf5fb376bf0ceefdb5/src/Altinn.App.Core/Models/Validation/ValidationIssue.cs#L49

src/Altinn.App.Core/Internal/Expressions/LayoutEvaluatorStateInitializer.cs
```diff
@@ -9,34 +11,94 @@ namespace Altinn.App.Core.Internal.Expressions;
...
+        }
+
+        var extraModels = await Task.WhenAll(dataTasks);
+
+        return new LayoutEvaluatorState(
+            new DataModel(
+                extraModels.Append(
+                    KeyValuePair.Create(defaultDataElement, await _dataAccessor.Get(instance, defaultDataElement))
+                )
+            ),
```
bjosttveit: The `DataModel` constructor crashes with `System.ArgumentException: An item with the same key has already been added.` if you reference the default data model explicitly in data model bindings or expressions somewhere, as it will then be present in `GetExternaModelReferences()` before being added again at the end.

This also causes the default model to be fetched twice so it should not be fixed by checking the already added models inside of the `DataModel` constructor. I suggest you add the default data type in the same way as the "external" ones to ensure uniqueness.

ivarne: I think this should be fixed in 21426412.

src/Altinn.App.Core/Features/Validation/Default/ExpressionValidator.cs
```diff
@@ -102,29 +105,33 @@ IAppMetadata appMetadata
...
+                                var validationIssue = new ValidationIssue
+                                {
+                                    Field = resolvedField.Field,
+                                    DataElementId = resolvedField.DataType,
+                                    Severity = validation.Severity ?? ValidationIssueSeverity.Error,
+                                    CustomTextKey = validation.Message,
+                                    Code = validation.Message,
+                                    Source = ValidationIssueSources.Expression,
+                                };
+                                validationIssues.Add(validationIssue);
```
bjosttveit, line 126: The `DataElementId` should not be set to the data type id?

ivarne: No, that would not be very helpful. Good catch!

PR #622, opened 2024-04-30
Signing optional data types

### LINKED ISSUE #595: Signering av valgfrie dataTypes
### Description

Vi har behov for å signere dataelementer, dersom de er med. Altså valgfrie dataTypes.

Når man setter opp signering, definerer man hvilke dataTypes som skal signeres. Dersom dataTypes i denne listen er valgfrie - eks de er konfigurert med minCount=0, får man feil pga manglende dataelement for signering.
Signeringen tar tilsynelatende ikke hensyn til konfigurasjonen av dataTypes i applicationmetadata.json.

Eksempel: Valgfri dataType
image

Signering task
image

Process next med signering - uten dataelement av typen IG
image

Process next med signering - med dataelement av typen IG på instansen
image

### Additional Information

_No response_

HauklandJ:
## Related Issue(s)
- #595 

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Core/Extensions/CollectionExtensions.cs
```diff
@@ -0,0 +1,23 @@
+namespace Altinn.App.Core.Extensions;
+
+/// <summary>
+/// Extensions for Collections.
+/// </summary>
+public static class CollectionExtensions
```
martinothamar: Don't think this should be public

martinothamar: I think we should add `NotNullWhenAttribute` to let the compiler know that `enumerable` is not null when the method returns false

src/Altinn.App.Core/Features/Action/SigningUserAction.cs
```diff
@@ -92,17 +100,39 @@ public async Task<UserActionResult> HandleAction(UserActionContext context)
...
+    {
+        var signatureIsConfigured =
+            currentTask.ExtensionElements?.TaskExtension?.SignatureConfiguration?.SignatureDataType is not null;
+        if (dataTypesToSign.IsNullOrEmpty() || !signatureIsConfigured)
+        {
+            return false;
+        }
+
+        var dataElementMatchExists = dataElements.Any(de =>
+            dataTypesToSign!.Any(dt => string.Equals(dt.Id, de.DataType, StringComparison.OrdinalIgnoreCase))
```
martinothamar: I think we should avoid null-forgiving operator as much as possible. In this case, using `NotNullWhenAttribute` on `IsNullOrEmpty` might be enough for it to realize it's not null?

Where we can't help the compiler with the flow analysis, I think it's better to either use
`Debug.Assert` or a manually thrown exception with a message describing the assumption rather than ending up with nullref

src/Altinn.App.Core/Features/Action/SigningUserAction.cs
```diff
@@ -92,17 +100,39 @@ public async Task<UserActionResult> HandleAction(UserActionContext context)
...
 
+    private static bool ShouldSign(
+        ProcessTask currentTask,
+        List<DataElement> dataElements,
+        List<DataType>? dataTypesToSign
+    )
+    {
+        var signatureIsConfigured =
+            currentTask.ExtensionElements?.TaskExtension?.SignatureConfiguration?.SignatureDataType is not null;
+        if (dataTypesToSign.IsNullOrEmpty() || !signatureIsConfigured)
```
martinothamar: Is the `IsNullOrEmpty` function worth it in this PR? when we have concrete type like a `List<>` I think we might as well use `dataTypesToSign is null or []` - codegen for that is a null check and `Count` check

HauklandJ: I saw that .IsNullOrEmpty was used elsewhere as well (PdfService) - by using Microsoft.IdentityModel.Tokens, line 73/74.

HauklandJ: Looking at it again, I think the implementation in PdfService was looking to use string.IsNullOrEmpty(), and instead did stringvar.IsNullOrEmpty by accident (which works as string is char[]). With this in mind, it may not be worth it

src/Altinn.App.Core/Features/Action/SigningUserAction.cs
HauklandJ: Whether or not we should use null-forgiving operators in cases where the compilers flow analysis "fails" should perhaps be a team rule?
I lean towards not using it, simply because it can be hard to spot

martinothamar: yeah agreed!

src/Altinn.App.Core/Extensions/CollectionExtensions.cs
HauklandJ: resolved as the class will be removed.

test/Altinn.App.Core.Tests/Features/Action/SigningUserActionTests.cs
```diff
@@ -31,7 +31,59 @@ public async void HandleAction_returns_ok_if_user_is_valid()
...
+        // Arrange
+        UserProfile userProfile = new UserProfile()
+        {
+            UserId = 1337,
+            Party = new Party() { SSN = "12345678901" }
+        };
+        var appMetadata = new ApplicationMetadata("org/id")
+        {
+            // Optional because MinCount == 0
+            DataTypes = [new DataType { Id = "model", MinCount = 1 }]
```
martinothamar: Is the comment wrong or is the value wrong here?

src/Altinn.App.Core/Features/Action/SigningUserAction.cs
```diff
@@ -92,17 +99,39 @@ public async Task<UserActionResult> HandleAction(UserActionContext context)
...
         List<DataElement> dataElements,
-        List<string> dataTypesToSign
+        List<DataType>? dataTypesToSign
     )
     {
         var connectedDataElements = new List<DataElementSignature>();
-        foreach (var dataType in dataTypesToSign)
+        if (dataTypesToSign is null or [])
+            return connectedDataElements;
+        foreach (var dataType in dataTypesToSign!)
```
martinothamar: null-forgiving operator not needed here

src/Altinn.App.Core/Features/Action/SigningUserAction.cs
```diff
@@ -92,17 +99,39 @@ public async Task<UserActionResult> HandleAction(UserActionContext context)
...
+            currentTask.ExtensionElements?.TaskExtension?.SignatureConfiguration?.SignatureDataType is not null;
+        if (dataTypesToSign is null or [] || !signatureIsConfigured)
+        {
+            return false;
+        }
+
+        var dataElementMatchExists = dataElements.Any(de =>
+            dataTypesToSign.Any(dt => string.Equals(dt.Id, de.DataType, StringComparison.OrdinalIgnoreCase))
+        );
+        var allDataTypesAreOptional = dataTypesToSign!.All(d => d.MinCount == 0);
```
martinothamar: null-forgiving operator not needed here

test/Altinn.App.Core.Tests/Features/Action/SigningUserActionTests.cs
HauklandJ: Nice, the comment actually helped uncover a bug. This test was a false positive as the mincount = 1 makes the datatype mandatory, BUT because instance.Data.Datatype includes the Id of the mandatory datatype then its ok.

martinothamar (approved):

PR #625, opened 2024-05-06
Treat build warnings as errors (in CI and optionally locally), fix and suppress all warnings

martinothamar:
## Description
I've noticed in recent PRs there are still a lot of `!` and warnings sneaking through. This PR makes warnings errors, and removed NRT-related errors

* preserves existing behavior as a first priority
* helps the compiler flow analysis where we can
* change annotations where we can
* supress the rest

Also removed some `async void` functions..

In the future (next time we do a breaking change), we should revisit all instances of `#nullable disable`, 
Maybe we should also use something like `Nullable.Extended.Analyzer` to disallow null-forgiving operator. 

1st commit - project config changes
2nd commit - remove `async void`
3rd commit - fix and suppress warnings where needed

## Related Issue(s)
- N/A

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,8 +1,18 @@
 <Project>
+    <PropertyGroup>
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+        <LangVersion>latest</LangVersion>
+    </PropertyGroup>
+
     <ItemGroup>
         <PackageReference Include="CSharpier.MsBuild" Version="0.28.2">
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
         </PackageReference>
+        <!-- TODO: enable this at some point, when we've got a handle on NRTs -->
+        <!-- <PackageReference Include="Nullable.Extended.Analyzer" Version="1.15.6169">
+            <PrivateAssets>all</PrivateAssets>
+            <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+          </PackageReference> -->
     </ItemGroup>
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,8 @@
 <Project>
-    <PropertyGroup>
+    <PropertyGroup Condition="'$(DOTNET_TREATWARNINGSASERRORS)' == 'true'">
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    </PropertyGroup>
+    <PropertyGroup>
         <LangVersion>latest</LangVersion>
     </PropertyGroup>
 
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,8 @@
 <Project>
-    <PropertyGroup Condition="'$(DOTNET_TREATWARNINGSASERRORS)' == 'true'">
+    <PropertyGroup Condition="'$(DOTNET_TREATWARNINGSASERRORS)' == 'true' OR '$(CI)' == 'true'">
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     </PropertyGroup>
+
     <PropertyGroup>
         <LangVersion>latest</LangVersion>
     </PropertyGroup>
--- a/src/Altinn.App.Api/Helpers/RequestHandling/DataRestrictionValidation.cs
+++ b/src/Altinn.App.Api/Helpers/RequestHandling/DataRestrictionValidation.cs
@@ -154,7 +154,10 @@ namespace Altinn.App.Api.Helpers.RequestHandling
         /// </summary>
         public static string? GetFileNameFromHeader(StringValues headerValues)
         {
-            ContentDispositionHeaderValue contentDisposition = ContentDispositionHeaderValue.Parse(headerValues);
+            string? headerValue = headerValues;
+            ArgumentNullException.ThrowIfNull(headerValue, nameof(headerValues));
+
+            ContentDispositionHeaderValue contentDisposition = ContentDispositionHeaderValue.Parse(headerValue);
             string? filename = contentDisposition.FileNameStar ?? contentDisposition.FileName;
 
             // We actively remove quotes because we don't want them replaced with '_'.
--- a/src/Altinn.App.Core/Configuration/AppSettings.cs
+++ b/src/Altinn.App.Core/Configuration/AppSettings.cs
@@ -65,6 +65,7 @@ namespace Altinn.App.Core.Configuration
         // TODO: can this be removed?
         // Env var being set is ServiceRepositorySettings__BaseResourceFolderContainer, but this prop is not used anywhere
 #nullable disable
+        [Obsolete("This is not used, and will be removed in the next major version")]
         public string BaseResourceFolderContainer { get; set; }
 
 #nullable restore
--- a/src/Altinn.App.Core/Configuration/AppSettings.cs
+++ b/src/Altinn.App.Core/Configuration/AppSettings.cs
@@ -215,7 +215,5 @@ namespace Altinn.App.Core.Configuration
         /// Enable the functionality to run expression validation in backend
         /// </summary>
         public bool ExpressionValidation { get; set; } = false;
-
-        public string RandomThing { get; set; }
     }
 }
--- a/src/Altinn.App.Core/Configuration/AppSettings.cs
+++ b/src/Altinn.App.Core/Configuration/AppSettings.cs
@@ -6,6 +6,7 @@ namespace Altinn.App.Core.Configuration
     /// <summary>
     /// Class that represents the ServiceRepositorySettings
     /// </summary>
+    // TODO: IOptions validation so that we know which of these properties are required
     public class AppSettings
     {
         /// <summary>
@@ -61,7 +62,12 @@ namespace Altinn.App.Core.Configuration
         /// <summary>
         /// Gets or sets the BaseResourceFolderContainer that identifies where in the docker container the runtime can find files needed
         /// </summary>
+        // TODO: can this be removed?
+        // Env var being set is ServiceRepositorySettings__BaseResourceFolderContainer, but this prop is not used anywhere
+#nullable disable
         public string BaseResourceFolderContainer { get; set; }
+
+#nullable restore
 
         /// <summary>
         /// Gets or sets The name of the FormLayout json file Name
@@ -142,6 +148,7 @@ namespace Altinn.App.Core.Configuration
         /// <summary>
         /// Open Id Connect Well known endpoint
         /// </summary>
+#nullable disable
         public string OpenIdWellKnownEndpoint { get; set; }
 
         /// <summary>
@@ -153,6 +160,8 @@ namespace Altinn.App.Core.Configuration
         /// Name of the cookie for runtime
         /// </summary>
         public string RuntimeCookieName { get; set; }
+
+#nullable restore
 
         /// <summary>
         /// Option to disable csrf check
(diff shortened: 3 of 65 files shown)


danielskovli (approved):
As a new addition to this team and codebase, perhaps take my approval with a grain of salt. But for what it's worth, I think all the changes in this PR are great refactors and puts in place important building blocks going forward.

As you mention, I agree that we ideally should bubble up `?` nullable types and handle these properly instead of telling the compiler to close its eyes. But where this fits in the roadmap for breaking changes for public properties and methods, I don't know.

Directory.Build.props
```diff
@@ -1,8 +1,18 @@
 <Project>
+    <PropertyGroup>
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+        <LangVersion>latest</LangVersion>
+    </PropertyGroup>
```
ivarne, line 8: Personally I think it makes sense to add this config only for CI, and allow developers to comment out a code block locally (for a quick sanity check) without removing using statements and fixing all warnings before being able to run tests.

martinothamar: That diagnostic doesn't seem to be warning severity by default (atleast on my machine in VSCode it's only a suggestion), but I get what you mean. So in the pic below, the unused using is no problem, but a variable that isn't used get complaints

image

Personally I'm fine either way

ivarne: Pro (only CI):
* more flexibility to run before all issues are fixed
* No need for ! or other warning supressions to test while working
Con:
* One more reason to be surprised when CI comes back with failiure, but we have this with sonar blocking us from adding new warnings now, and I don't think it is much of a problem

src/Altinn.App.Api/Helpers/RequestHandling/DataRestrictionValidation.cs
```diff
@@ -154,7 +154,10 @@ public static class DataRestrictionValidation
         /// </summary>
         public static string? GetFileNameFromHeader(StringValues headerValues)
         {
-            ContentDispositionHeaderValue contentDisposition = ContentDispositionHeaderValue.Parse(headerValues);
+            string? headerValue = headerValues;
```
ivarne, line 157: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.stringvalues.op_implicit?view=net-8.0#microsoft-extensions-primitives-stringvalues-op-implicit(microsoft-extensions-primitives-stringvalues)-system-string

>Defines an implicit conversion of a given StringValues to a string, with multiple values joined as a comma separated string.

Shouldn't we rather take the `Single` (or first or last) header, instead of a comma separated concatenation?

src/Altinn.App.Core/Configuration/AppSettings.cs
```diff
@@ -61,8 +62,13 @@ public class AppSettings
         /// <summary>
         /// Gets or sets the BaseResourceFolderContainer that identifies where in the docker container the runtime can find files needed
         /// </summary>
+        // TODO: can this be removed?
+        // Env var being set is ServiceRepositorySettings__BaseResourceFolderContainer, but this prop is not used anywhere
+#nullable disable
         public string BaseResourceFolderContainer { get; set; }
 
+#nullable restore
```
ivarne, line 71: If it is never referenced we can at least add [Obsolete] to it. It seems wrong anyway, because folder locations are specified by the developer in their dockerfile and the platform environment variables does not know the layout.

I tend to favor ` = default!;` instead of `#nullable disable` on a small region in a class, but now that I have read the full PR I kind of like the temporary look of what you did.

src/Altinn.App.Core/Configuration/AppSettings.cs
```diff
@@ -186,8 +195,11 @@ public class AppSettings
         /// <summary>
         /// Gets or sets the version of the application.
         /// </summary>
+#nullable disable
         public string AppVersion { get; set; }
 
+#nullable restore
+
```
ivarne, line 203: Here we can just use `string.Empty` or `"not set from environement"`

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
```diff
@@ -389,13 +390,15 @@ public byte[] GetRuleHandlerForSet(string id)
...
-            byte[] filedata = null;
+            byte[]? filedata = null;
             if (File.Exists(fileName))
             {
                 filedata = File.ReadAllBytes(fileName);
             }
 
+#nullable disable
             return filedata;
+#nullable restore
```
ivarne, line 401: We can change the signature for private methods.

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
```diff
@@ -411,7 +414,9 @@ private byte[] ReadFileContentsFromLegalPath(string legalPath, string filePath)
                 return File.ReadAllBytes(fullFileName);
             }
 
+#nullable disable
             return null;
+#nullable restore
```
ivarne, line 419: Change the signature for private methods

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
```diff
@@ -451,6 +453,7 @@ Stream stream
             string apiUrl = $"{_platformSettings.ApiStorageEndpoint}instances/{instanceIdentifier}/data/{dataGuid}";
             string token = _userTokenProvider.GetUserToken();
             StreamContent content = new StreamContent(stream);
+            ArgumentNullException.ThrowIfNull(contentType);
```
ivarne, line 456: use `#nullable disable` here to postpone the issue for a cleanup. Makes no sense to declare `string? contentType` and just throw a deliberate exception.

src/Altinn.App.Core/Infrastructure/Clients/Storage/InstanceClient.cs
```diff
@@ -99,7 +99,7 @@ public async Task<List<Instance>> GetInstances(Dictionary<string, StringValues>
 
             foreach (var queryParameter in queryParams)
             {
-                foreach (string value in queryParameter.Value)
+                foreach (string? value in queryParameter.Value)
                 {
                     apiUrl.Append($"&{queryParameter.Key}={value}");
```
ivarne, line 105: We must add escaping here.

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
```diff
@@ -132,7 +132,11 @@ private async Task<string> GetLanguage()
...
         if (userId != null)
         {
-            UserProfile userProfile = await _profileClient.GetUserProfile((int)userId);
+            UserProfile? userProfile = await _profileClient.GetUserProfile((int)userId);
+            if (userProfile is null)
+            {
+                throw new Exception("Could not get user profile.");
+            }
 
             if (!string.IsNullOrEmpty(userProfile.ProfileSettingPreference?.Language))
```
ivarne, line 139: I would just add an extra `?` here, instead of throwing an exception for missing profile. The default is good enough.

src/Altinn.App.Core/Internal/Process/ProcessNavigator.cs
```diff
@@ -37,7 +37,11 @@ ILogger<ProcessNavigator> logger
     public async Task<ProcessElement?> GetNextTask(Instance instance, string currentElement, string? action)
     {
         List<ProcessElement> directFlowTargets = _processReader.GetNextElements(currentElement);
-        List<ProcessElement> filteredNext = await NextFollowAndFilterGateways(instance, directFlowTargets, action);
+        List<ProcessElement> filteredNext = await NextFollowAndFilterGateways(
+            instance,
+            directFlowTargets.Cast<ProcessElement?>().ToList(),
```
ivarne:
```C#
List<ProcessElement?> directFlowTargets = _processReader.GetNextElements(currentElement)!;
List<ProcessElement> filteredNext = await NextFollowAndFilterGateways(instance, directFlowTargets, action);
```
No reason to make a copy

test/Altinn.App.Api.Tests/Altinn.App.Api.Tests.csproj
```diff
@@ -5,7 +5,7 @@
 		<Nullable>enable</Nullable>
 		<ImplicitUsings>enable</ImplicitUsings>
 		<IsPackable>false</IsPackable>
-		<NoWarn>$(NoWarn);CS1591;CS0618</NoWarn>
+		<NoWarn>$(NoWarn);CS1591;CS0618;CS7022</NoWarn>
 		<!--
 			CS1591: Don't warn for missing XML doc
 			CS0618: This is a test project, so we usually continue testing [Obsolete] apis
```
ivarne, line 11: CS7022: The entry point of the program is global code; ignoring 'Program.Main(string[])' entry point

ivarne (approved):
Very good work!

I have a few comments for improvemnts, but mostly things I noticed that could be argued are out of scope for this PR.

test/Altinn.App.Api.Tests/Altinn.App.Api.Tests.csproj
ivarne:
```suggestion
			CS0618: This is a test project, so we usually continue testing [Obsolete] apis
			CS7022: The entry point of the program is global code; ignoring 'Program.Main(string[])' entry point
```

src/Altinn.App.Api/Helpers/RequestHandling/DataRestrictionValidation.cs
martinothamar: I didn't want to change behavior in this PR

src/Altinn.App.Core/Configuration/AppSettings.cs
martinothamar: Yeah it's temporary, idea being that we already use `#nullable disable` a lot of places, so that for the next major bump we will just search through all of these. I also want to minimize null-forgiving operator as I kind of want to use `Nullable.Extended.Analyzer` too

src/Altinn.App.Core/Configuration/AppSettings.cs
martinothamar: That would change behavior though

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
martinothamar: I know, but then I would have to deal with this 2 places instead of here

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
martinothamar: This would just be moving the directive... I can do it but it doesn't improve anything

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
martinothamar: This value is immediately passed to `MediaTypeHeaderValue.Parse`, which will throw this exception, so this just saves us having to annotate further. Current behavior is preserved

src/Altinn.App.Core/Infrastructure/Clients/Storage/InstanceClient.cs
martinothamar: Will do! 🕐

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
martinothamar: Prefer not to change behavior here, like we discussed

src/Altinn.App.Core/Internal/Process/ProcessNavigator.cs
martinothamar: You mentioned if a direct cast would be better/clearer, to which I agree.. But surprisingly that does produce a warning 
image

PR #636, opened 2024-05-08
Remove/fix or suppress null-forgiving operator usage in `src/`, enable warning/build errors

martinothamar:
## Description
As a sort of followup on #625, this PR 

* fixes/removes null-forgiving operator usage where possible
* suppresses with TODO comment where fixing would require making breaking changes (API or behavior)

only in `src/` folder for now

## Related Issue(s)
- N/A

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Features/Action/SigningUserAction.cs
+++ b/src/Altinn.App.Core/Features/Action/SigningUserAction.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics.CodeAnalysis;
 using Altinn.App.Core.Internal.App;
 using Altinn.App.Core.Internal.Process;
 using Altinn.App.Core.Internal.Process.Elements;
@@ -75,14 +76,15 @@ public class SigningUserAction : IUserAction
                 ?.DataTypes?.Where(d => dataTypeIds.Contains(d.Id, StringComparer.OrdinalIgnoreCase))
                 .ToList();
 
-            if (ShouldSign(currentTask, context.Instance.Data, dataTypesToSign))
+            var shouldSign = ShouldSign(currentTask, context.Instance.Data, dataTypesToSign);
+            if (shouldSign.Value)
             {
                 var dataElementSignatures = GetDataElementSignatures(context.Instance.Data, dataTypesToSign);
                 SignatureContext signatureContext =
                     new(
                         new InstanceIdentifier(context.Instance),
                         currentTask.Id,
-                        currentTask.ExtensionElements?.TaskExtension?.SignatureConfiguration?.SignatureDataType!,
+                        shouldSign.SignatureDataType,
                         await GetSignee(context.UserId.Value),
                         dataElementSignatures
                     );
@@ -100,24 +102,28 @@ public class SigningUserAction : IUserAction
         );
     }
 
-    private static bool ShouldSign(
+    private readonly record struct ShouldSignResult(
+        [property: MemberNotNullWhen(true, "SignatureDataType")] bool Value,
+        string? SignatureDataType
+    );
+
+    private static ShouldSignResult ShouldSign(
         ProcessTask currentTask,
         List<DataElement> dataElements,
         List<DataType>? dataTypesToSign
     )
     {
-        var signatureIsConfigured =
-            currentTask.ExtensionElements?.TaskExtension?.SignatureConfiguration?.SignatureDataType is not null;
-        if (dataTypesToSign is null or [] || !signatureIsConfigured)
+        var signatureDataType = currentTask.ExtensionElements?.TaskExtension?.SignatureConfiguration?.SignatureDataType;
+        if (dataTypesToSign is null or [] || signatureDataType is null)
         {
-            return false;
+            return new(false, null);
         }
 
         var dataElementMatchExists = dataElements.Any(de =>
             dataTypesToSign.Any(dt => string.Equals(dt.Id, de.DataType, StringComparison.OrdinalIgnoreCase))
         );
         var allDataTypesAreOptional = dataTypesToSign.All(d => d.MinCount == 0);
-        return dataElementMatchExists || allDataTypesAreOptional;
+        return new(dataElementMatchExists || allDataTypesAreOptional, signatureDataType);
     }
 
     private static List<DataElementSignature> GetDataElementSignatures(
--- a/src/Altinn.App.Core/Features/Action/SigningUserAction.cs
+++ b/src/Altinn.App.Core/Features/Action/SigningUserAction.cs
@@ -76,15 +76,16 @@ public class SigningUserAction : IUserAction
                 ?.DataTypes?.Where(d => dataTypeIds.Contains(d.Id, StringComparer.OrdinalIgnoreCase))
                 .ToList();
 
-            var shouldSign = ShouldSign(currentTask, context.Instance.Data, dataTypesToSign);
-            if (shouldSign.Value)
+            if (
+                GetDataTypeForSignature(currentTask, context.Instance.Data, dataTypesToSign) is string signatureDataType
+            )
             {
                 var dataElementSignatures = GetDataElementSignatures(context.Instance.Data, dataTypesToSign);
                 SignatureContext signatureContext =
                     new(
                         new InstanceIdentifier(context.Instance),
                         currentTask.Id,
-                        shouldSign.SignatureDataType,
+                        signatureDataType,
                         await GetSignee(context.UserId.Value),
                         dataElementSignatures
                     );
@@ -102,12 +103,7 @@ public class SigningUserAction : IUserAction
         );
     }
 
-    private readonly record struct ShouldSignResult(
-        [property: MemberNotNullWhen(true, "SignatureDataType")] bool Value,
-        string? SignatureDataType
-    );
-
-    private static ShouldSignResult ShouldSign(
+    private static string? GetDataTypeForSignature(
         ProcessTask currentTask,
         List<DataElement> dataElements,
         List<DataType>? dataTypesToSign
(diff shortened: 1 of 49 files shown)


src/Altinn.App.Core/Models/Validation/ValidationIssue.cs
```diff
@@ -71,7 +71,10 @@ public class ValidationIssue
         /// </remarks>
         [JsonProperty(PropertyName = "source")]
         [JsonPropertyName("source")]
-        public string Source { get; set; } = default!;
+#nullable disable
+        public string Source { get; set; }
+
+#nullable restore
```
ivarne, line 77: See `<remarks>`:

```suggestion
        public string Source { get; set; } = "[]"
```

We need a permanent solution for this. It can't be required, but it makes no sense if marked as nullable, as we set it for all types before returning.

src/Altinn.App.Core/Internal/Language/ApplicationLanguage.cs
```diff
@@ -44,6 +44,7 @@ public async Task<List<Models.ApplicationLanguage>> GetApplicationLanguages()
             {
                 await using (FileStream fileStream = new(fileInfo.FullName, FileMode.Open, FileAccess.Read))
                 {
+                    // ! TODO: find a better way to deal with deserialization errors here, rather than adding nulls to the list
```
ivarne, line 56: It currently trows in all other cases than when the file contains a the four letters `null`

src/Altinn.App.Core/Internal/Patch/PatchService.cs
```diff
@@ -76,7 +76,9 @@ IAppModel appModel
         var patchResult = jsonPatch.Apply(oldModelNode);
         if (!patchResult.IsSuccess)
         {
-            bool testOperationFailed = patchResult.Error!.Contains("is not equal to the indicated value.");
+            bool testOperationFailed =
+                patchResult.Error?.Contains("is not equal to the indicated value.")
+                ?? throw new Exception("JsonPatch failed, but no error message was provided");
```
ivarne: I think this issue has been fixed in v3 of JsonPatch. https://github.com/gregsdennis/json-everything/pull/611

src/Altinn.App.Core/Internal/Patch/PatchService.cs
```diff
@@ -135,7 +137,7 @@ IAppModel appModel
         return new DataPatchResult { NewDataModel = result.Ok, ValidationIssues = validationIssues };
     }
 
-    private static ServiceResult<object, string> DeserializeModel(Type type, JsonNode patchResult)
+    private static ServiceResult<object, string> DeserializeModel(Type type, JsonNode? patchResult)
```
ivarne, line 152: Doesn't this change cause a warning on the next line? Seems like I forgot to add annotation for the `.Result` after `.Success` is true

src/Altinn.App.Core/Internal/Process/ProcessTasks/Common/ProcessTaskFinalizer.cs
```diff
@@ -164,7 +164,9 @@ DataType dataType
             else
             {
                 // Remove the shadow fields from the data
-                data = JsonSerializer.Deserialize(serializedData, modelType)!;
+                data =
+                    JsonSerializer.Deserialize(serializedData, modelType)
+                    ?? throw new JsonException("Could not deserialize back datamodel after removing shadow fields");
```
ivarne:
```suggestion
                    ?? throw new JsonException("Data was \"null\"");
```

src/Altinn.App.Core/Features/Action/SigningUserAction.cs
```diff
@@ -99,24 +101,28 @@ public async Task<UserActionResult> HandleAction(UserActionContext context)
         );
     }
 
-    private static bool ShouldSign(
+    private readonly record struct ShouldSignResult(
+        [property: MemberNotNullWhen(true, "SignatureDataType")] bool Value,
+        string? SignatureDataType
+    );
+
+    private static ShouldSignResult ShouldSign(
```
ivarne: This rewrite took a while to understand. Wouldn't a `private static string? GetDataTypeForSignature(---)` combined with `if (dataTypeForSignature is null)` do the same trick?

src/Altinn.App.Core/Implementation/PrefillSI.cs
```diff
@@ -99,11 +99,18 @@ IHttpContextAccessor httpContextAccessor
...
+                    profilePrefill.ToObject<Dictionary<string, string>>()
+                    ?? throw new Exception("Unexpectedly failed to convert profilePrefill JToken to dictionary");
 
                 if (userProfileDict.Count > 0)
                 {
-                    int userId = AuthenticationHelper.GetUserId(_httpContextAccessor.HttpContext!);
+                    var httpContext =
+                        _httpContextAccessor.HttpContext
+                        ?? throw new Exception(
+                            "HttpContext is unexpectedly null, need HttpContext to get current user"
```
ivarne: You wrote this better in a previous location
```suggestion
                            "Could not get HttpContext - must be in a request context to get current user");
```

ivarne (approved):
Not sure of the value of banning `!` but otherwise this looks good to me. I wrote a few comments, mostly to ensure I was awake and really looked at all the changes.

src/Altinn.App.Core/Models/Validation/ValidationIssue.cs
martinothamar: Did read the remark, but still opted for keeping the same behavior. In the second to last commit I implemented what you suggested and added some test assertions to verify. Does that look OK to you?

src/Altinn.App.Core/Internal/Language/ApplicationLanguage.cs
martinothamar: Indeed! I added another comment to say that, let me know if this was not what you meant

src/Altinn.App.Core/Internal/Patch/PatchService.cs
martinothamar: Nice! But what are you asking for here?

Current behavior = throw NRE
this PR = throw contextualized exception

src/Altinn.App.Core/Internal/Patch/PatchService.cs
martinothamar: `Deserialize` here is an extension method where the receiver is `JsonNode?`. It ends up here and is handled as expected: https://github.com/dotnet/runtime/blob/b15e35138eaa43e242d5eaa8680f9cb3367b5ef6/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializer.Read.Node.cs#L183-L186

src/Altinn.App.Core/Internal/Process/ProcessTasks/Common/ProcessTaskFinalizer.cs
martinothamar: I still think the existing context is useful, but extended the message to include what you suggested

src/Altinn.App.Core/Features/Action/SigningUserAction.cs
martinothamar: Very good point, that ended up a lot simpler 😄 Latest commit implements that

martinothamar:
Great comments as always!

> Not sure of the value of banning `!`

I should have stated my intentions in the description, so it is not to ban `!`, but rather raise the bar for using it. I find that in the vast majority of `!`-uses (in any codebase I've seen) end up in one of the following categories

* Throw NREs, so we can improve the code by throwing contextualized exceptions, `ArgumentNullException` etc
* Are easily fixed by helping the compiler flow analysis with attributes or other means
* Can be fixed by improving the domain modelling

So my hope is that this diagnostic being emitted will lead us to better habits essentially, and actually deal with the nulls. Even if we ignore the above and it is supressed - we have to give a reason in a comment

src/Altinn.App.Core/Models/Validation/ValidationIssue.cs
martinothamar: Actually, I remember now I searched for `new ValidationIssue` and found that `Source` is not at all always non-null. So I whole-heartedly regret that commit 😄 

Here is one example
image
So now `Source` will be `"[]"` as opposed to null. I don't think this makes any sense. This is a classic case where the warning that was emitted should have led to a better design as opposed to being ignored/suppressed IMO

src/Altinn.App.Core/Internal/Process/ProcessEventHandlingDelegator.cs
```diff
@@ -72,6 +72,7 @@ IEnumerable<IProcessTask> processTasks
                         case InstanceEventType.process_StartEvent:
                             break;
                         case InstanceEventType.process_StartTask:
+                            // ! TODO: figure out why taskId can be null here
```
bjorntore, line 75: It's because some of the events don't have taskIds (StartEvent and EndEvent). Code can be restructured to avoid the ! though.

martinothamar: Ahh I see. What would be the change here? Something we can do now (non-breaking) ? I'm happy to try, or I could make a better comment here if it's something we should do later

bjorntore: I can give it a shot and commit here.

src/Altinn.App.Core/Internal/Patch/PatchService.cs
martinothamar: I did not read that PR properly. Fixed 😄

PR #641, opened 2024-05-09
Cleanup analyzers and StyleCop config, enable builtint .NET static analysis

martinothamar:
## Description
Originally wanted to cleanup config related to analysers and StyleCop. When I had consolidated config into .editorconfig mostly and referenced Stylecop.Analyzers everywhere there were a ton of warnings (~1k), so the existing config wasn't working (perhaps it was only enabled in Api).

Since Stylecop mostly relates to code style and formatting, maybe we can just remove it now that we have Csharpier which kind of eliminates this concern? There is obviously some stuff that Csharpier doesn't fix, such as naming convention (which I want to fix with .editorconfig).

My recent changes of this flavour have been mainly aimed at increasing the quality bar for _new_ code, but for these kinds of changes that isn't really possible, so in this PR I have also implemented codefixes for currently `Minimum` analysis level. Ideally we will work our way up to `Recommended` (while supressing the stuff we don't agree with ofc)

## Related Issue(s)
- N/A

## Verification
- [x] All tests run green

## Documentation


ivarne (approved):
Very nice to see this kind of work!!!

I'm a little sceptical of `static` methods everywhere it is possible, because of two issues:
* I'm afraid it might not always be correct when implementing parts that will be used by reflection. It would probably be a bug in the reflection code instead.
* It might cause issues when developing because I have not yet referenced the class property, (or temporary commented it out, or expecting to hit `extract interface` at some point), causing a less pleasant dev experience.

Having the editor suggest making such methods static would be great though. Just not as a requirement/warning.

PR #644, opened 2024-05-13
RFC: Add paramter to process/next for frontend to warnings to ignore

ivarne:
Currently we ignore all warnings on process/next, but that means that some warnigns will never be seen and confirmed by the end user.

This adds an optional list that frontend can specify so that warnings that are not explicitly ignored is not ignored.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


ivarne:
@olemartinorg seems to prefer that we just ignore warnings when the validation system fails to present them to the user before hitting submit.

PR #651, opened 2024-05-16
Roslyn analyzer setup included in Altinn.App.Core, analyzer for `IHttpContextAccessor.HttpContext` usage

martinothamar:
## Description

This PR used to include some analysis to catch configuration mistakes. Back when I worked on this we agreed that we wanted more "normal" config parsing and validation code that could be shared between a Roslyn analyzer and runtime validation. Now this PR contains just the basic project setup (kept the test and bench setup and fixtures so that we can easily get started with new stuff). Includes

* Tests for AppImplementationFactory analysis
* App analyzer for `IHttpContextAccessor.HttpContext` usage

Docs: https://github.com/Altinn/altinn-studio-docs/pull/2116

image

------------

### Old description

This PR includes a Roslyn analyzer setup that we can use for various things.
To start, this analyzes the `applicationmetadata.json` file and reports errors when it isn't able to resolve `classRef` to its corresponding C# class

image

The analyzer (and its' `Newtonsoft.Json` dependency) is packaged alongside `Altinn.App.Core` in the analyzers folder

image

I use the `4.1.0` version of the Roslyn libraries for the analyzer. It's supported by [almost all](https://learn.microsoft.com/en-us/visualstudio/extensibility/roslyn-version-support?view=vs-2022) of the VS 2022 versions and has been included in the .NET SDK since 6.0 I think.

Referencing locally

```csproj
<ProjectReference Include="../../../app-lib-dotnet/src/Altinn.App.Core/Altinn.App.Core.csproj" />
<ProjectReference Include="../../../app-lib-dotnet/src/Altinn.App.Api/Altinn.App.Api.csproj">
    <CopyToOutputDirectory>lib\$(TargetFramework)\*.xml</CopyToOutputDirectory>
</ProjectReference>
<ProjectReference Include="../../../app-lib-dotnet/src/Altinn.App.Analyzers/Altinn.App.Analyzers.csproj" PrivateAssets="all" ReferenceOutputAssembly="false" OutputItemType="Analyzer" />
```

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Analyzers/MetadataAnalyzer.cs
```diff
@@ -0,0 +1,91 @@
...
+                out var projectDir
+            )
+        )
+        {
+            context.ReportDiagnostic(Diagnostic.Create(Diagnostics.ProjectNotFound, Location.None));
+            return;
+        }
+
+#pragma warning disable RS1035 // Do not use APIs banned for analyzers
+        // On compilation end, it is not totally unreasonable to read a file
```
ivarne: Working around the issue of all apps adding `<AdditionalFiles Include="app/config/applicationmetadata.json" />` (or a wildcard for all json files if it works.)

See: https://github.com/dotnet/roslyn/blob/main/docs/analyzers/Using%20Additional%20Files.md

src/Altinn.App.Analyzers/MetadataAnalyzer.cs
```diff
@@ -0,0 +1,91 @@
...
+            return;
+        }
+
+        try
+        {
+            var jsonText = File.ReadAllText(file);
+#pragma warning restore RS1035 // Do not use APIs banned for analyzers
+            var metadata = JObject.Parse(jsonText) ?? throw new Exception("Deserialization returned 'null'");
+            var dataTypes =
+                (metadata.GetValue("dataTypes") as JArray) ?? throw new JsonException("Failed to parse 'dataTypes'");
```
ivarne: Why Newtonsoft instead of system.text.json? I think we should try to standardise (or mostly I don't want to learn the library that seems to be moved away from).

ivarne (commented):
Cool...

We should probably make these Errors instead of Warnings unless we find a way to make CI build run with `warnings as errors`, because lots of apps has many warnings and these pretty serious issues will just drown.

src/Altinn.App.Analyzers/MetadataAnalyzer.cs
martinothamar: Yep! We could make a TODO for v9 to add that into the project files through the CLI. I don't think we can do anything about that now?

src/Altinn.App.Analyzers/MetadataAnalyzer.cs
martinothamar: Simply because System.Text.Json [has a bunch of `netstandard2.0` dependencies](https://www.nuget.org/packages/System.Text.Json/9.0.0-preview.3.24172.9#dependencies-body-tab) whereas `Newtonsoft.Json` has none 😄 It's a bit of a hassle as transitive dependencies all the way down have to be copied. I gave it a shot when building this out but didn't want to invest a ton of time into it atm

martinothamar:
> We should probably make these Errors instead of Warnings unless we find a way to make CI build run with warnings as errors, because lots of apps has many warnings and these pretty serious issues will just drown.

If we want to deploy this before v9 (I would like to), then we can't make them errors I think.

ivarne:
I think we can make compile errors in broken apps?

src/Altinn.App.Analyzers/MetadataAnalyzer.cs
ivarne: 😭

martinothamar:
This PR is ready for proper review now!

Testing infra was as usual not trivial at all, and took the majority of time, but I'm really happy with the result.
There is a global xunit collection to run all tests serially. The initialization of the fixture (building the Roslyn workspace etc) takes relatively long compared to the actual tests so I think that's the right tradeoff? See output below 
image
This is a "hot" build, building the workspace took 2s and the tests that use them take take less than a second at least. This approach also gives us some flexibility as we can make physical changes to the project files safely.
The only diagnostic not tested is the `ProjectFileNotFound` one. I considered the new [IFileProvider](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.fileproviders.ifileprovider?view=net-8.0) stuff to be able to fake the file system access/checks but that would bring in more dependencies into the analyzer which is a bit of a hassle. Could also build our own as the amount of code isn't huge. The output of the diagnostics also use Verify snapshots. I've used the `Verify.SourgeGenerators` package before but since we only want to verify snapshots here I opted for copying out only the `Diagnostic` JSON converter which was used there. To verify error handling I opted for injectable delegates that are invoked from the analyzer project. I added some comments to the `AltinnTestAppFixture` and the `ProjectModification` files to explain more of the testing approach.

Another change is using [PolySharp](https://github.com/Sergio0694/PolySharp) to polyfill some stuff such as records. That enabled me to use a pattern that I'm really fond of, especially in places where precise error handling is very important. There is a `ApplicationMetadataResult` type that functions as a sort of union type using a closed type hierarchy. I use an [analyzer](https://github.com/shuebner/ClosedTypeHierarchyDiagnosticSuppressor) to suppress exhaustiveness check diagnostics. Code:

Callee: https://github.com/Altinn/app-lib-dotnet/pull/651/files#diff-481c24765c42685ba9fdf20ea3d628bbfada6d7c92ccc1e915fb8fed7d8f9269R6-R59
Caller: https://github.com/Altinn/app-lib-dotnet/pull/651/files#diff-7c717ef7bb4564e83aed692d56aca3056beef2b6b1bcfc66af0137c4663bd344R67-R98

Note that the caller doesn't need a default case in the switch statement, yet all error conditions are propery handled.

I also tried to add proper locations to diagnostics by capturing lineinfo from `Newtonsoft.Json` while serializing, seems to work for the current tests.

@ivarne  re
> I think we can make compile errors in broken apps?

I think this is a reasonable perspective for sure, but another would be that making those error severity would change build-time behavior. So some apps that are broken now still build, but that changes if we make diagnostics error severity right now. Another approach would be to keep them as warning now, but convert them to errors for v9. Those who want to be strict can turn on `WarningsAsErrors`, so we still enable that in this scenario. We agreed that we wanted to be strict about breaking changes, so that's why I opted for the current approach. I'm open different solutions, but we should probably consider these subtleties

martinothamar:
Since we're trying to improve DevEx with these changes, I thought I'd verify the perf hit of adding static analysis (since this will run per compilation for apps). Added a benchmarking harness, with running the `MetadataAnalyzer` as a first benchmark. Benchmark results:

image

CPU profile:

image

martinothamar:
A new case that popped up - a user had mistakenly created a `Setting.json` file in the `ui/` folder as opposed to `Settings.json` 😄

ivarne:
> A new case that popped up - a user had mistakenly created a `Setting.json` file in the `ui/` folder as opposed to `Settings.json` 😄 

Or even funnier, they could name it `settings.json` instead of `Settings.json` and see that everything work locally on windows, but have the config missing in docker in production. (I have done that)

src/Altinn.App.Core/Altinn.App.Core.csproj
```diff
@@ -44,8 +44,17 @@
...
+    <None Include="../Altinn.App.Analyzers/bin/$(Configuration)/netstandard2.0/Altinn.App.Analyzers.dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="true" />
+    <!-- <None Include="../Altinn.App.Analyzers/bin/$(Configuration)/netstandard2.0/System.Text.Json.dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="true" /> -->
+    <None Include="packaging/Altinn.App.Core.props" Pack="true" PackagePath="build/" Visible="true" Condition="'$(MinVerVersionOverride)'==''" />
+    <None Include="packaging/Altinn.App.Core.Experimental.props" Pack="true" PackagePath="build/" Visible="true" Condition="'$(MinVerVersionOverride)'!=''" />
   </ItemGroup>
 
+  <Target Name="CreateExperimentalPropsFile" AfterTargets="Build" Condition="'$(MinVerVersionOverride)'!=''">
+    <Move SourceFiles="packaging/Altinn.App.Core.props" DestinationFiles="packaging/Altinn.App.Core.Experimental.props"/>
+  </Target>
+
```
ivarne, line 68: Comment

src/Altinn.App.Core/packaging/Altinn.App.Core.props
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <PropertyGroup>
+    <IsAltinnApp>true</IsAltinnApp>
+  </PropertyGroup>
+  <ItemGroup>
+    <AdditionalFiles Include="config/**" />
+    <AdditionalFiles Include="ui/**/*.json" />
+  </ItemGroup>
+</Project>
```
ivarne, line 18: Comment why

test/Altinn.App.Analyzers.Tests/Fixtures/DiagnosticJsonConverter.cs
```diff
@@ -0,0 +1,52 @@
+using Microsoft.CodeAnalysis;
+
+namespace Altinn.App.Analyzers.Tests.Fixtures;
+
+// Based on: https://github.com/VerifyTests/Verify.SourceGenerators/blob/18eb1cd67ebf76803f164751ecb27397a4c11951/src/Verify.SourceGenerators/Converters/DiagnosticConverter.cs
```
ivarne, line 5: Ikke fornøyd med default snapshot

test/Altinn.App.Analyzers.Tests/_snapshots/MetadataAnalyzerTests.AppMetadata_Deserialization_Failure_Diagnostic.verified.txt
```diff
@@ -0,0 +1,15 @@
...
+  {
+    Id: ALTINN0012,
+    Title: Altinn app metadata file couldn't be parsed,
+    Severity: Warning,
+    WarningLevel: 1,
+    Location: ExternalFile({CurrentDirectory}App/config/applicationmetadata.json@1:1),
+    HelpLink: https://docs.altinn.studio/app/development/analysis/rules/ALTINN0012,
+    MessageFormat: Could not parse application metadata file: '{0}' {1},
+    Message:
+Could not parse application metadata file: 'AppMetadata_Deserialization_Failure_Diagnostic: On appmetadata deserialization'    STACKTRACE
```
ivarne: delete snapshots

ivarne (approved):

PR #653, opened 2024-05-16
Split switch that handled both process task events and other non task events

bjorntore:


test/Altinn.App.Core.Tests/Internal/Process/ProcessEventHandlingTests.cs
```diff
@@ -533,17 +533,7 @@ public async Task UpdateProcessAndDispatchEvents_EndEvent_confirmation_instance_
...
-                    CurrentTask = new()
-                    {
-                        ElementId = "Task_2",
-                        AltinnTaskType = "confirmation",
-                        Name = "Bekreft",
-                        Flow = 2
-                    },
-                    EndEvent = "EndEvent"
-                }
+                ProcessInfo = new() { CurrentTask = null, EndEvent = "EndEvent" }
```
bjorntore, line 536: As far as I can tell EndEvent will not contain a current task, so this test was configured a bit wrong.

src/Altinn.App.Core/Internal/Process/ProcessEventHandlingDelegator.cs
```diff
@@ -109,6 +73,63 @@ IEnumerable<IProcessTask> processTasks
...
+            ProcessElementInfo currentTask,
+            Dictionary<string, string>? prefill
+        )
+        {
+            string taskId = currentTask.ElementId;
+            string altinnTaskType = currentTask.AltinnTaskType;
+
+            switch (eventType)
+            {
+                case InstanceEventType.process_StartTask:
```
bjorntore, line 88: Might be better to do command pattern instead of these switches, but that's a challenge for later.

martinothamar (approved):
Perfect! 🚀

bjorntore:
Hang on a bit, don't merge. Think I might need to check if it's a process task a bit differently.

bjorntore:
I'll redo this later. Closing PR to avoid merge issues with other cleanup PRs.

PR #657, opened 2024-05-26
Refactor to remove null-forgiving operator related to `AltinnPaymentConfiguration.PaymentDataType`

martinothamar:
## Description
While rebasing #636 based on the new payment stuff I noticed an opportunity to restructure some things around `AltinnPaymentConfiguration` so that we

* Don't have the same validation logic several places
* Express the validated state in a type `ValidAltinnPaymentConfiguration`

which in the end let's us remove the null-forgiving operator where `AltinnPaymentConfiguration.PaymentDataType` is used

## Related Issue(s)
- N/A

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Process/Elements/AltinnExtensionProperties/AltinnPaymentConfiguration.cs
+++ b/src/Altinn.App.Core/Internal/Process/Elements/AltinnExtensionProperties/AltinnPaymentConfiguration.cs
@@ -1,4 +1,6 @@
+using System.Diagnostics.CodeAnalysis;
 using System.Xml.Serialization;
+using Altinn.App.Core.Internal.App;
 
 namespace Altinn.App.Core.Internal.Process.Elements.AltinnExtensionProperties
 {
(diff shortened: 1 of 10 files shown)


HauklandJ (approved):

src/Altinn.App.Core/Internal/Process/Elements/AltinnExtensionProperties/AltinnPaymentConfiguration.cs
```diff
@@ -12,5 +14,45 @@ public class AltinnPaymentConfiguration
         /// </summary>
         [XmlElement("paymentDataType", Namespace = "http://altinn.no/process")]
         public string? PaymentDataType { get; set; }
+
+        internal ValidAltinnPaymentConfiguration Validate()
+        {
+            List<string>? errorMessages = null;
```
bjorntore, line 20: Honestly don't really need this list of error message anymore. Had more config values before, hence the list, but now that we only check one value it's not really needed. MIGHT be handy in the future. Up to you if you want to simply it.

bjorntore (approved):

src/Altinn.App.Core/Internal/Process/Elements/AltinnExtensionProperties/AltinnPaymentConfiguration.cs
martinothamar: Ahh I see. Since it was a bit of a hassle convincing the compiler of the nullability flow, I'll keep it as is for now so I remember this works 😄

PR #659, opened 2024-05-27
Tests for telemetry enriching middleware, make ASP.NET Core spans root, misc improvements

martinothamar:
## Description
We didn't have any tests for the telemetry middleware. I added tests and suggest some improvements

* Add `AuthenticateMethod`, `OrganisationName` and `OrganisationNumber` as tags on root trace when available
* Make ASP.NET Core spans always root the root of the trace
* Make label names more consistent - `.` separator as opposed to `_` which we still had in some cases
* Run `WebApplicationFactory` tests with `UseOpenTelemetry` = `true`
  * Since the `TelemetrySink` is now used in both Core tests and Api tests, I created a common project to keep it 
* Make the telemetry middleware type itself internal

### Making ASP.NET Core spans always root

The default behavior in ASP.NET Core (or is it OTel? I don't remember anymore) is that ASP.NET Core spans extract the tracecontext from the incoming request. Usually this is nice and convenient, but it can also be a bit scary. At that point, clients control
  * Trace size - i.e. have many spans, and this has implication for infrastructure. For example Tempo has a max trace limit, and it's hard to make configuration like this consistent throughout infra and apps. Keeping control of where the traces begin is not a definitive solution but it's going to be a lot harder to mess up (and we would be messing up)
  * Baggage - baggage is always passed on to child spans, so in practice clients of app APIs can inject arbitrary content into the diagnostics machinery which can slow things down among other things..  

To illustrate the problem I made some dummy code in an app tracking the `main` branch

image

The dummy endpoint is exposed from an app which is shipping telemetry to the localtest LGTM stack. In this case the client has HTTP client instrumentation turned on. It created a loop of 1000 iterations, sending 1000 HTTP requests within a single root span. The app happily accepts this and in Grafana we can see the resulting trace

image

Baggage is not passed on through OTLP as attributes, but it's still taking up memory and being propagated throughout.
So even though this PR just splits those spans into separate traces (the actual span data ends up with the same size), but we avoid ingesting and propagating random baggage. See the `OtelPropagator` and `AspNetCorePropagator` types for impl.

## Related Issue(s)
- N/A

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -259,6 +270,69 @@ public Task StartAsync(CancellationToken cancellationToken)
             public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
         }
 
+        internal sealed class OtelPropagator : TextMapPropagator
+        {
+            private readonly TextMapPropagator inner;
```
HauklandJ: Perhaps `_inner` to avoid `this.`?

HauklandJ (approved):
Nice!

PR #663, opened 2024-05-29
Fix issue where datamodel decimal properties are stored with greater precision than is supported by frontend

martinothamar:
## Description
Suggested fix for resolving issues around decimal precisions across systems

A datamodel generated by Studio might look like this:
image

In some cases, apps have calculated fields that are decimal, which can lead to a lot of precision being used.
When this data is stored, all 128bits of `decimal` precision come with, but the frontend which lives in JS land
will only have 64bit numbers. So when data is posted back (for whatever reason, maybe they shouldn't in this scenario), the `test` JSON patch operation will fail because the field has higher precision coming from storage than it has from the client request.

Alternatives
* Keep storing 128bit precision, but make JSON patch more tolerant instead (comparison based on double precision)
* Don't allow decimal properties in datamodels
* ???

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/issues/2066

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Helpers/ObjectUtils.cs
+++ b/src/Altinn.App.Core/Helpers/ObjectUtils.cs
@@ -1,5 +1,6 @@
 using System.Collections;
 using System.Reflection;
+using System.Text.Json.Serialization;
 using System.Xml.Serialization;
 
 namespace Altinn.App.Core.Helpers;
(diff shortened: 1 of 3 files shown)


martinothamar:
@nkylstad @mirkoSekulic do you have any input or ideas on this?

ivarne:
Tried my luck at writing some tests, and obviously found a few issues and previous errors and locations for improved comments.

1. The same problem with rounding in javascript happens to `long` so I included it as well.
2. This is fixing a Json problem, so we should not round properties with `[JsonIgnore]`

```c#
public string StringBackingField { get; set; }

[JsonIgnore]
public decimal DecimalView { get { decimal.Parse(StringBackingField )} set {StringBackingField = value.ToString()} }

```

ivarne:
@martinothamar I don't love this solution, but it does solve some very real problems that likely won't ever go away for existing apps. Can you add a comment with the justification for not doing this?

martinothamar:
Conclusion was posted in: https://github.com/Altinn/app-frontend-react/issues/2066#issuecomment-2161140745
Solution being updated docs, and not special casing large numeric types, that was my interpretation anyway

PR #666, opened 2024-05-30
Remove `Scrutor` as dependency, refactor `ProfileClient`

martinothamar:
## Description
In #589 we migrated the protobuf-net metrics that were used in service decorators to use `System.Diagnostics` stuff in `Telemetry`. So now there was only cache decorator for `ProfileClient` left. This PR moves caching inside the client and removed the package dep

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Infrastructure/Clients/Profile/ProfileClient.cs
+++ b/src/Altinn.App.Core/Infrastructure/Clients/Profile/ProfileClient.cs
@@ -11,10 +11,25 @@ using Altinn.Common.AccessTokenClient.Services;
 using Altinn.Platform.Profile.Models;
 using AltinnCore.Authentication.Utils;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace Altinn.App.Core.Infrastructure.Clients.Profile;
+
+internal static class ProfileClientDI
+{
+    public static IServiceCollection AddProfileClient(this IServiceCollection services)
+    {
+        services.AddTransient<IProfileClient>(sp => new ProfileClientCachingDecorator(
+            ActivatorUtilities.CreateInstance<ProfileClient>(sp),
+            sp.GetRequiredService<IMemoryCache>(),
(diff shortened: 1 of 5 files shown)


martinothamar:
P.S. could not find any usage of Scrutor API from search across apps

src/Altinn.App.Core/Infrastructure/Clients/Profile/ProfileClient.cs
```diff
@@ -61,12 +68,38 @@ public class ProfileClient : IProfileClient
         _appMetadata = appMetadata;
         _accessTokenGenerator = accessTokenGenerator;
         _telemetry = telemetry;
+
+        _cache = new MemoryCache(new MemoryCacheOptions());
+        _cacheEntryOptions = new MemoryCacheEntryOptions()
+        {
+            AbsoluteExpirationRelativeToNow = TimeSpan.FromSeconds(cacheSettings.Value.ProfileCacheLifetimeSeconds)
+        };
     }
```
ivarne: The old code gets the memory cache from DI.

Won't this approach disable the cache in an app when the service is registered with `.AddHttpClient<>`?

test/Altinn.App.Core.Tests/Infrastructure/Clients/Profile/ProfileClientTests.cs
```diff
@@ -0,0 +1,190 @@
...
+    [Fact]
+    public async Task Returns_Test_Profile_Cached()
+    {
+        const int userId = 1234;
+        await using var fixture = BuildFixture(() => new UserProfile { UserId = userId, });
+
+        var client = fixture.ServiceProvider.GetRequiredService<IProfileClient>();
+
+        var profile1 = await client.GetUserProfile(userId);
+        var profile2 = await client.GetUserProfile(userId);
```
ivarne: Resolve the client from different scopes to ensure that it caches across scopes.

ivarne (changes requested):
I think there is a difference in the implementation here.

martinothamar:
True, there were breaking changes as well, since those types are public... Went with a different approach

HauklandJ (approved):

ivarne (approved):

ivarne:
I don't think we need to have stable dependencies in DI constructors, but this is a fine solution.

PR #669, opened 2024-06-03
Implement IMaskinportenClient interface

### LINKED ISSUE #602: Implement IMaskinportenClient interface
### Description

In relation to #492, app-lib-dotnet needs to expose an `IMaskinportenClient` interface along with an implementation.

Assuming that the app developer has configured the required Maskinporten scopes via Altinn Studio (publish page), the app should be able to receive an `IMaskinportenClient` instance at runtime. This instance will handle Maskinporten authentication and expose a `.GetAccessToken` method or simila; for use with `HttpClient` calls to protected APIs.

### Additional Information

Should we also expose a factory method that returns a pre-authenticated `HttpClient` instance? If so, consider `IHttpClientFactory` [[1](https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests), [2](https://medium.com/@imstevenphan/leveraging-ihttpclientfactory-in-asp-net-e1970b9719f5)]

### LINKED ISSUE #714: MaskinportenClient implementation cannot use HybridCache because it is currently in preview
## Description
Nuget is refusing to release a stable version of Altinn.App.* because the new Maskinporten implementation uses a package which is currently only available as a [preview release](https://www.nuget.org/packages/Microsoft.Extensions.Caching.Hybrid).

## Status
[The current implementation](https://github.com/Altinn/app-lib-dotnet/pull/669) has been reverted from main, because it prevented a scheduled nuget release. This code on this branch is feature-complete, but suffers from the issue described above.

## Proposed resolution
Either rewrite the implementation or hold off release until this HybridCache is available as a stable release, which will be when [.NET 9.0](https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-9.0?view=aspnetcore-8.0#new-hybridcache-library) is released later this year. At which point the package should be stable, or we could upgrade the framework to get the functionality without an additional dependency.

## Related PR
- https://github.com/Altinn/app-lib-dotnet/pull/715

## Related issues
- https://github.com/Altinn/app-lib-dotnet/issues/681

This new issue will serve as the new tracking for this feature. The old issue has been closed, and may as well stay closed.

danielskovli:
## Description

Implements IMaskinportenClient interface as per [proof of concept project](https://github.com/danielskovli/maskinporten-authentication-dotnet).

## Related Issue(s)
- #602 

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Api/Extensions/WebHostBuilderExtensions.cs
```diff
@@ -27,8 +30,42 @@ public static void ConfigureAppWebHost(this IWebHostBuilder builder, string[] ar
...
+                    string jsonFile = Path.GetFileName(jsonAbsolutePath);
+
+                    configBuilder.AddJsonFile(
+                        provider: new PhysicalFileProvider(jsonDir),
+                        path: jsonFile,
+                        optional: true,
+                        reloadOnChange: true
+                    );
+                }
+
```
ivarne, line 35: Her tenker jeg du bør kjøre `Extract Method` på den nye koden og gi den et navn for å få bedre kodestruktur.

src/Altinn.App.Core/EFormidling/Implementation/EformidlingStatusCheckEventHandler2.cs
```diff
@@ -25,7 +25,11 @@ public class EformidlingStatusCheckEventHandler2 : IEventHandler
     private readonly IEFormidlingClient _eFormidlingClient;
     private readonly ILogger<EformidlingStatusCheckEventHandler2> _logger;
     private readonly IHttpClientFactory _httpClientFactory;
+
+#pragma warning disable CS0618
```
ivarne:
```suggestion
#pragma warning disable CS0618 // 'member' is obsolete:
```

src/Altinn.App.Core/EFormidling/Implementation/EformidlingStatusCheckEventHandler2.cs
```diff
@@ -36,7 +40,9 @@ public class EformidlingStatusCheckEventHandler2 : IEventHandler
         IEFormidlingClient eFormidlingClient,
         IHttpClientFactory httpClientFactory,
         ILogger<EformidlingStatusCheckEventHandler2> logger,
+#pragma warning disable CS0618
```
ivarne:
```suggestion
#pragma warning disable CS0618 // 'member' is obsolete:
```

src/Altinn.App.Core/Features/Maskinporten/Converters/JsonWebKeyConverter.cs
```diff
@@ -0,0 +1,78 @@
...
+                $"The JsonWebKey is invalid after deserialization, not all required properties were found: {validationResult}"
+            );
+    }
+
+    /// <inheritdoc/>
+    public override void Write(Utf8JsonWriter writer, JsonWebKey value, JsonSerializerOptions options)
+    {
+        throw new NotImplementedException();
+    }
+}
```
ivarne, line 74: Er dette for å slippe `JsonSerializer.Deserialize<JsonWebKey>(jwkString, new JsonSerializerOptions{PropertyNameCaseInsensitive = true,})`

src/Altinn.App.Api/Extensions/HttpClientBuilderExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </para>
+    /// </summary>
+    /// <param name="builder">The Http client builder</param>
+    /// <param name="scopes">One or more scopes to claim authorization for with Maskinporten</param>
+    public static IHttpClientBuilder UseMaskinportenAuthorization(
+        this IHttpClientBuilder builder,
+        params string[] scopes
+    )
+    {
+        return builder.AddHttpMessageHandler(provider => new MaskinportenDelegatingHandler(scopes, provider));
```
ivarne: Konvensjonen her er vel vanligvis å bruke `provider.GetRequiredSerivce<>()` i builder metoden eller  `ActivatorUtilities.CreateInstance<MaskinportenDelegatingHandler>(scopes)` og ikke sende `ServiceProvider` ned til klassen.
https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities.createinstance?view=net-8.0

src/Altinn.App.Core/Features/Maskinporten/MaskinportenClient.cs
```diff
@@ -0,0 +1,273 @@
...
+namespace Altinn.App.Core.Features.Maskinporten;
+
+/// <inheritdoc/>
+public sealed class MaskinportenClient : IMaskinportenClient
+{
+    /// <summary>
+    /// The margin to take into consideration when determining if a token has expired (seconds).
+    /// <remarks>This value represents the worst-case latency scenario for <em>outbound</em> connections carrying the access token.</remarks>
+    /// </summary>
+    internal const int TokenExpirationMargin = 30;
```
ivarne, line 19: Can this be a percent value? Eg use the token for 50% (or 70%) of its advertised lifespan. Could also be configurable so that it is easier to change if we encounter problems.

src/Altinn.App.Core/Features/Maskinporten/MaskinportenClient.cs
```diff
@@ -0,0 +1,273 @@
...
+    /// <summary>
+    /// Instantiates a new <see cref="MaskinportenClient"/> object.
+    /// </summary>
+    /// <param name="options">Maskinporten settings.</param>
+    /// <param name="httpClientFactory">HttpClient factory.</param>
+    /// <param name="logger">Optional logger interface.</param>
+    public MaskinportenClient(
+        IOptionsMonitor<MaskinportenSettings> options,
+        IHttpClientFactory httpClientFactory,
+        ILogger<MaskinportenClient>? logger = default
```
ivarne: Why is the logger optional?

src/Altinn.App.Core/Features/Maskinporten/MaskinportenClient.cs
```diff
@@ -0,0 +1,273 @@
...
+        {
+            _logger?.LogDebug("Waiting for token request to resolve with Maskinporten");
+            return lazy.Value;
+        }
+
+        _logger?.LogDebug(
+            "Using cached access token which expires at {ExpiresAt}",
+            ((MaskinportenTokenResponse)result).ExpiresAt
+        );
+        return Task.FromResult((MaskinportenTokenResponse)result);
```
ivarne: Jeg blir veldig forvirret av denne cache implementasjonen. Når returnerer den `MaskinportenTokenResponse` og når returnerer den `Lazy<Task<MaskinportenTokenResponse>>`

test/Altinn.App.Core.Tests/Features/Maskinporten/Models/MaskinportenSettingsTest.cs
```diff
@@ -0,0 +1,75 @@
...
+                    "e": "",
+                    "use": "sig",
+                    "kid": "test-kid",
+                    "qi": "E-oyO4HWOVxD_d7xZxFltTZDz6ZtLPZFB_KYXYeVFDrO8KZE9kFb4TNlFvrFjv-dHtpNey95gqtOtdNMwdVdZbAKbDdo5LYSJ_rk-4ZVsDusq7FCJ5nmmrxfQ5yNEPqHwgdUfs50v_fV1x8SEDjfWzaaVK5ltqPiUXtpTTLBQIg",
+                    "dp": "yVvJ6y6VgjfszjldBFNv_qHwlz58MJw5sg_mcBfJX_4Tp-pzReNy42xeGXnkuOaM2qE6tGcw5y5tgmV8XUxRiyV-R3y5WbpVFBwOGu6i1vkTxaiZXM3oAz5vz2oUQrJIgO1bzXa28NxtbFQrq1jw4G4Tpjzcqlqc06QGqXzn0vk",
+                    "alg": "RS256",
+                    "dq": "B5CI7dhAfvhsq9FE35b5oZ6SxlDT4ZT0XTqVVM-fp3Op0JDUpgGfazyqtXm6M98UNhxBlkj2Yq8f7PW7HHbwe_tgWPuKeUs4OSZGpnfCrFrnbps79suYdew4dK6NWkwz-MDMJRvPlrk2XNqA32xmmAsaVkkH67CNlM2AaZ0La2E",
+                    "n": "sy9DZ1U6jfP1UBN2EZTD1DPkajdZsFsXGGVHfbJmH5MFwXbtKlwV_jYjz58YIj1n48OxH7f-Ldgc-fBLz45QU1HbDZPij7q3uYm1ZMTGkPqkY8kHX51TsFOEqzVhNfyc6yVsjlj5KPyyxLyAcx6ixiE2K8vIeuKMVbZCZt605L39ENUsiQ-cfnqp-zo1ihU5xJOQaWV9pGuG4XoLAUIktF6_YPF4pFmSWRHk5aURUfTCvo11n3EUBYJUiJb8AqUt3yqGSoV-4wXir-9oRNjDUtE_QA3eErGKCebtUd6oxzcXcHiGY0npKxt7JQti3jTZRcnkScmmP-XvrQzB6kzSCQ"
+                }
```
ivarne: Legg på en kommentar om hva du har endret for å trigge valideringsfeil?

ivarne (commented):
Ser bra ut. Noen raske kommentarer (håper de er gode).

#### WebApplicationFactory
Savner en full test der du bruker WebApplicationFactory til å hente token til en `IOptionProvider`, for å teste hele løpet, men det blir kanskje for mye mocking av litt mange http klienter til at det egentlig gir verdi. (mulig CustomWebApplicationFactory.SendAsync gjør at du kan teste med flere http klienter og at maskinporten delegating handler kommer i forkant.)
https://github.com/Altinn/app-lib-dotnet/blob/0161134de61bf4349d9934f8877aa53b6d35bbf7/test/Altinn.App.Api.Tests/CustomWebApplicationFactory.cs#L212-L215



#### Tid
Har ikke gjort det selv, men det finnes interfaces for å gjøre det enklere å teste cache expiration med tid (uten Task.Delay som gjør testene veeeldig trege)

https://andrewlock.net/exploring-the-dotnet-8-preview-avoiding-flaky-tests-with-timeprovider-and-itimer/

src/Altinn.App.Api/Extensions/HttpClientBuilderExtensions.cs
danielskovli: Litt av det her henger igjen fra før jeg hadde fullstendig comit til dependency injection. Bra feedback, jeg fikser

src/Altinn.App.Core/Features/Maskinporten/Converters/JsonWebKeyConverter.cs
danielskovli: Dette er fordi den versjonen av JsonWebKey vi har tilgjengelig i app-lib-dotnet ikke har støtte for serialisering direkte. Ved oppgradering til en nyere pakke som støtter dette, brekker veldig mye eksisterende funksjonalitet.

Jeg gikk ikke i dybden på hvordan det eventuelt kunne fikses i de andre situasjonene, fordi det relaterer til ting vi egentlig skriver litt ut med denne nye maskinporten integrasjonen.

JWK props er definert ganske sterkt i diverse RFCs, så jeg bare laget en egen parser.

src/Altinn.App.Core/Features/Maskinporten/MaskinportenClient.cs
danielskovli: I have had this conversation with @martinothamar as well. I don't personally believe a relative value makes any sense, mainly because:

We added a margin here to account for outbound latency. The inbound latency between Maskinporten and the library code has already been accounted for. The outbound latency for authorized http calls aren't really related to a percentage of when the token expires. They are related to an absolute time when the token is considered expired. What we want to assure here, is that we don't ship off a token that expires in transit.

In my mind, this requires a fixed value and does not relate to the relative lifespan of the token. 30 seconds is obviously a generous buffer, but at the same time is lower than the default http client timeout (100 seconds): https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.timeout?view=net-8.0

src/Altinn.App.Core/Features/Maskinporten/MaskinportenClient.cs
danielskovli: It shouldn't be, good catch 👍 Another throwback to olden times code

src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -95,6 +98,30 @@ IWebHostEnvironment env
...
+    /// <param name="configureOptions">
+    /// Action delegate that provides <see cref="MaskinportenSettings"/> configuration for the <see cref="MaskinportenClient"/> service
+    /// </param>
+    public static IServiceCollection ConfigureMaskinportenClient(
+        this IServiceCollection services,
+        Action<MaskinportenSettings> configureOptions
+    )
+    {
+        services.RemoveOptions<MaskinportenSettings>();
+        services.AddOptions<MaskinportenSettings>().Configure(configureOptions).ValidateDataAnnotations();
```
martinothamar, line 120: I don't know if this is expected behavior when using `Configure...` APIs like this, usually all configure alls are applied in order. So a user could apply some config in 1 call and the rest of the config in another

src/Altinn.App.Api/Extensions/WebHostBuilderExtensions.cs
```diff
@@ -27,8 +30,59 @@ public static void ConfigureAppWebHost(this IWebHostBuilder builder, string[] ar
...
+    {
+        if (services.GetOptionsDescriptor<MaskinportenSettings>() is null)
+        {
+            services
+                .AddOptions<MaskinportenSettings>()
+                .BindConfiguration("MaskinportenSettings")
+                .ValidateDataAnnotations();
+        }
+
+        services.AddSingleton<IMaskinportenClient, MaskinportenClient>();
```
martinothamar: I wonder if this part should will fit better in `AddAltinnAppServices` or `AddAppServices` instead? Since all the other app services come from there

src/Altinn.App.Core/Features/Maskinporten/Converters/JsonWebKeyConverter.cs
```diff
@@ -0,0 +1,78 @@
...
+/// Reads a JSON blob containing a jwk, converting it to a <see cref="JsonWebKey"/> instance
+/// </summary>
+public class JsonWebKeyConverter : JsonConverter<JsonWebKey>
+{
+    /// <inheritdoc/>
+    public override JsonWebKey? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        if (reader.TokenType != JsonTokenType.StartObject)
+        {
+            throw new JsonException();
```
martinothamar: nit: Should we have exception messages here?

src/Altinn.App.Core/Features/Maskinporten/Converters/JsonWebKeyConverter.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Text.Json.Serialization;
+using Altinn.App.Core.Features.Maskinporten.Models;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Altinn.App.Core.Features.Maskinporten.Converters;
+
+/// <summary>
+/// Reads a JSON blob containing a jwk, converting it to a <see cref="JsonWebKey"/> instance
+/// </summary>
+public class JsonWebKeyConverter : JsonConverter<JsonWebKey>
```
martinothamar: I think this can be `internal sealed`?

src/Altinn.App.Core/Features/Maskinporten/Exceptions/MaskinportenException.cs
```diff
@@ -0,0 +1,18 @@
+namespace Altinn.App.Core.Features.Maskinporten.Exceptions;
+
+/// <summary>
+/// Generic Maskinporten related exception. Something went wrong, and it was related to Maskinporten.
+/// </summary>
+public class MaskinportenException : Exception
```
martinothamar: nit: `abstract`?

src/Altinn.App.Core/Features/Maskinporten/Exceptions/MaskinportenTokenExpiredException.cs
```diff
@@ -0,0 +1,18 @@
+namespace Altinn.App.Core.Features.Maskinporten.Exceptions;
+
+/// <summary>
+/// An exception that indicates the access token has expired when it was in fact expected to be valid
+/// </summary>
+public class MaskinportenTokenExpiredException : MaskinportenException
```
martinothamar: nit: concrete exceptions `sealed`? Since they are meant to model the failure modes of our code, I think it's safe to say users should inherit from them

src/Altinn.App.Core/Features/Maskinporten/Models/MaskinportenSettings.cs
```diff
@@ -0,0 +1,181 @@
...
+    [Required]
+    [JsonPropertyName("key")]
+    [JsonConverter(typeof(JsonWebKeyConverter))]
+    public required JsonWebKey Key { get; set; }
+}
+
+/// <summary>
+/// Serialization wrapper for a JsonWebKey object
+/// </summary>
+internal sealed record JwkWrapper
```
martinothamar: nit: maybe this is a good candidate for being a struct, seeing as it's a temporary local, so we can avoid any heap alloc

test/Altinn.App.Api.Tests/Telemetry/TelemetryConfigurationTests.cs
```diff
@@ -259,7 +230,7 @@ public async Task OpenTelemetry_Registers_Correctly_When_Enabled()
             new("AppSettings:UseOpenTelemetry", "true"),
         ];
         Telemetry? telemetry = null;
-        await using (var app = BuildApp(configData))
+        await using (var app = AppBuilder.Build(configData: configData))
```
martinothamar, line 233: Nice 🥇

martinothamar (commented):
👏  

I made a small change to `UseMaskinportenAuthorization` and the related test since I found it easier than to explain, hope that's OK

src/Altinn.App.Core/Features/Maskinporten/MaskinportenClient.cs
danielskovli: I'll have another stab at clarifying the intent in this method, and perhaps add some locking as was brainstormed the other day. TBC.

src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
danielskovli: In our case, I'm not sure if that's possible due to the `[Required]` attributes and how we call `ValidateDataAnnotations()` after loading the config.

This was because we wanted the JSON to be validated... but if I'm not remembering wrong, this means we have to provide a valid object each time if manually configuring.

That said, I'd be happy to remove the `RemoveOptions` call if that goes against common convention (of which I know very little, unfortunately).

src/Altinn.App.Api/Extensions/WebHostBuilderExtensions.cs
danielskovli: That's a good point. I think because I had the "add json" and the "add service" wrapped in the same method in the POC, I naturally wanted them to stay close when porting over. What you're saying makes sense though, I'll make it happen.

src/Altinn.App.Core/Features/Maskinporten/Converters/JsonWebKeyConverter.cs
danielskovli: Probably. I couldn't come up with anything better than "your JSON is wrong, fix it"... but maybe that's better than nothing 😄

src/Altinn.App.Core/Features/Maskinporten/Models/MaskinportenSettings.cs
danielskovli: This is where my my somewhat limited track record in .NET shows through. What you're saying makes sense though, I'll look into it.

src/Altinn.App.Core/Features/Maskinporten/Converters/JsonWebKeyConverter.cs
ivarne: Expected object for "JsonWebKey". Found {JsonTokenType}

danielskovli:
@ivarne 
> Savner en full test der du bruker WebApplicationFactory til å hente token til en `IOptionProvider`, for å teste hele løpet

Jeg synes vi har ganske bra dekning på alle komponentene nå, men det jeg savner selv er vel en _faktisk_ implementasjon med en test-mottaker som kan ta imot en request og verifisere at `Authorization: Bearer xxx` stemmer.

All annen mocking blir bare ... mer mocking. Ta gjerne en titt på den litt [oppdaterte testen](https://github.com/Altinn/app-lib-dotnet/pull/669/files#diff-9ee77abeaa34defb508602eefdfa30c001744b920bb25e7304d0185e6607146cR92) her og se om det hjelper noe på magefølelsen.

> Tid [...] Har ikke gjort det selv, men det finnes interfaces for å gjøre det enklere å teste cache expiration med tid 

Veldig interessert i dette! Skal ta en titt 😄

danielskovli:
@martinothamar 
> I made a small change to UseMaskinportenAuthorization and the related test since I found it easier than to explain, hope that's OK

Very appreciated and more than okay! 🙏

src/Altinn.App.Core/Features/Maskinporten/Models/MaskinportenSettings.cs
danielskovli: Implemented as a record struct, seems to work as intended :)

src/Altinn.App.Core/Features/Maskinporten/MaskinportenClient.cs
danielskovli: New implementation with `LazyRefreshCache` seems more clear to me and also solves a few edge case concerns. Considering this resolved for now.

HauklandJ:
TODO: instrument

src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
martinothamar: I think the validations run after all configures/postconfigures are ran: https://github.com/dotnet/runtime/blob/c1cebefa8b0522a120a455d0b768fc297134523d/src/libraries/Microsoft.Extensions.Options/src/OptionsFactory.cs#L55-L92

martinothamar:
> All annen mocking blir bare ... mer mocking

Helt enig, synes mocking har en tendens til å bli veldig verbose, og det er vanskelig å få oversikt over hva man faktisk tester in the end (fant en test bug senest i forrige uke relatert til dette...).

Noen som har erfaring med https://github.com/WireMock-Net/WireMock.Net/wiki/What-Is-WireMock.Net ?
Det er jo fortsatt en slags Mock, men kanskje det funker bedre i praksis. Vi burde også være flinke til å injecte failures - timeouts, auth osv

ivarne:
Har brukt fattigmans varianten av wiremock i testing av pdf clienten. Se SendAsync på baseklassen vår for webapplicationfactory.

martinothamar:
Inntrykket mitt var at man testet standard `HttpClient` også med den, dvs det sendes faktisk requests, mulig det er feil, men da får man ihvertfall dekning på hvordan man håndterer exceptions o.l. som kommer fra standard `HttpClient`

src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
danielskovli: I've removed the `RemoveOptions` call, but opted to leave the `public required T ...` in the model, since this makes the downstream code a bit cleaner. So in practice you couldn't initialize a partial configuration, then later come back and finish the job. 

I have, however, validated that the last value added overrides any previous values. Eg. the unit test for this covers adding two configs and validating that the last config added is the one retrieved with `IOptionsMonitor.CurrentValue`.

src/Altinn.App.Core/Helpers/LazyRefreshCache.cs
```diff
@@ -0,0 +1,224 @@
...
+                            }
+                        }
+
+                        // Remove the Lazy Task from the _taskCache because it is no longer needed
+                        // and need to be gone when the cache is refreshed
+                        _taskCache.TryRemove(key, out var _);
+
+                        return value;
+                    }
+                    catch (Exception e)
```
HauklandJ: nit: No need to assign var if not to be used.`catch (Exception e)` ->`catch (Exception)`

test/Altinn.App.Core.Tests/Features/Maskinporten/Delegates/MaskinportenDelegatingHandlerTest.cs
```diff
@@ -0,0 +1,63 @@
...
+            scopes,
+            new MaskinportenTokenResponse
+            {
+                TokenType = "Bearer",
+                Scope = "-",
+                AccessToken = "jwt-content-placeholder",
+                ExpiresIn = -1
+            }
+        );
+        var httpClient = new HttpClient(handler);
```
HauklandJ, line 25: Some instances here of disposables not being disposed. Adding a `using` e.g. `using var httpClient = new HttpClient(handler);` fixes this.

danielskovli: I was just going with the flow based on various docs and posts, like this: https://stackoverflow.com/questions/50912160/should-httpclient-instances-created-by-httpclientfactory-be-disposed

Essentially the httprequesthandler is what you normally want to dispose, but in this case, the factory will ensure its _not_ disposed (on purpose, because of our middleware).

So I don't really know if we should dispose the client here. Chatter online indicates leave it alone, but feel free to make a firm recommendation here if you have an opinion after reading this response :)

test/Altinn.App.Core.Tests/Features/Maskinporten/Models/MaskinportenTokenResponseTest.cs
```diff
@@ -0,0 +1,35 @@
+using System.Text.Json;
+using Altinn.App.Core.Features.Maskinporten.Models;
+using FluentAssertions;
+
+namespace Altinn.App.Core.Tests.Features.Maskinporten.Models;
+
+public class MaskinportenTokenResponseTest
+{
+    [Fact]
+    public void ShouldDeserializeFromJsonCorrectly()
```
HauklandJ, line 10: Seems to me that we are simply testing System.Text.Json here? I guess one could argue that we are testing that our model is implemented correctly?

danielskovli: The thought behind  this test was to see if we implement our model (and System.Text.Json markup) in a valid way as per a known response from Maskinporten. 

If they change their response without telling us, we won't catch it. If someone messes with our model in a breaking manner, the test will fail. Best I could come up with.

test/Altinn.App.Core.Tests/Features/Maskinporten/Delegates/MaskinportenDelegatingHandlerTest.cs
HauklandJ: Can't argue with famous actor Chris Pratt.

HauklandJ: I dug a bit deeper into this as I have disposed httpClients before and do it instinctively.[ Microsoft does it in their example code](https://learn.microsoft.com/en-us/dotnet/core/extensions/httpclient-factory#basic-usage), but they do not explain why..

martinothamar (approved):

ivarne:
Ja, så langt jeg kan se er wiremock skrevet i java og kjører i en ekstern prosess på et eget portnummer (antakelig i docker). Da får du full testing av all http relatert kode i din app, siden det bare blir serveren som er en mock.

Kan godt hende at vi har behov der wiremock er riktig løsning, men jeg er litt tvilende. Så lenge C# lar oss lage en `MockHttpMessageHandler` der vi selv implementerer `SendAsync` som transofrmerer en `HttpRequestMessage` til en `HttpResponseMessage` er jeg usikker på om full http stack mot en mocket server egentlig gir oss noe mer i unit test sammenheng. Det eneste måtte være et bedre api for å registrere flere handlere for ulike api requester i samme test, men det er ganske enkelt å forbedre det jeg skrev også til å ha en routing (enten hand rolled `Dictionary<Regex,Handler>` eller se om det finnes noe nuget pakker).

src/Altinn.App.Core/Features/Maskinporten/MaskinportenClient.cs
```diff
@@ -0,0 +1,272 @@
...
+        CancellationToken cancellationToken = default
+    )
+    {
+        string formattedScopes = FormattedScopes(scopes);
+        DateTimeOffset referenceTime = _timeprovider.GetUtcNow();
+
+        _telemetry?.StartGetAccessTokenActivity(_options.CurrentValue.ClientId, formattedScopes);
+
+        var result = await _tokenCache.GetOrCreateAsync(
+            formattedScopes,
```
ivarne: Er det garantert at `formattedScopes` er unikt og ikke vil gi en navnekollisjon nå som du bruker er delt cache?

danielskovli: Forsåvidt ingen garanti for det, når jeg tenker meg om. Kan padde litt med en unik identifier kanskje?

ivarne: Ja, det er nok lurt. Utrolig vanskelig å feilsøke cache key kollisjoner.

danielskovli: Jeg elsker salt både i maten og i koden, så det her blir bra

bjorntore:
Setting ignore-for-release since this had to be reverted from main due to a preview dependency, which is not allowed.

There will be a new PR for this later.

PR #675, opened 2024-06-06
Set `AnalysisMode` to `Recommended`, fix resulting warnings

martinothamar:
## Description
We did `Minimum` level last time, now trying to get to recommended

* Suppressed warnings that would lead to breaking change, or was nitpicky
* A couple of parameter names that got warnings
* Lots of StringComparison and int parsing stuff
  * `OrdinalIgnoreCase` when operating on URLs
  * `Ordinal` where `StringComparison` was missing
  * Some `CurrentCulture` variants I switched to `InvariantCulture` 

We were/are a bit inconsistent on when to ignore case or not, so please review those bits in particular

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/DataController.cs
+++ b/src/Altinn.App.Api/Controllers/DataController.cs
@@ -1,3 +1,4 @@
+using System.Globalization;
 using System.Net;
 using System.Security.Claims;
 using System.Text.Json;
@@ -133,7 +134,7 @@ public class DataController : ControllerBase
             Application application = await _appMetadata.GetApplicationMetadata();
 
             DataType? dataTypeFromMetadata = application.DataTypes.First(e =>
-                e.Id.Equals(dataType, StringComparison.InvariantCultureIgnoreCase)
+                e.Id.Equals(dataType, StringComparison.Ordinal)
             );
 
             if (dataTypeFromMetadata == null)
@@ -302,7 +303,9 @@ public class DataController : ControllerBase
                 return NotFound($"Did not find instance {instance}");
             }
 
-            DataElement? dataElement = instance.Data.First(m => m.Id.Equals(dataGuid.ToString()));
+            DataElement? dataElement = instance.Data.First(m =>
+                m.Id.Equals(dataGuid.ToString(), StringComparison.Ordinal)
+            );
 
             if (dataElement == null)
             {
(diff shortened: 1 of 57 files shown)


src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -133,7 +134,7 @@ IPatchService patchService
             Application application = await _appMetadata.GetApplicationMetadata();
 
             DataType? dataTypeFromMetadata = application.DataTypes.First(e =>
-                e.Id.Equals(dataType, StringComparison.InvariantCultureIgnoreCase)
+                e.Id.Equals(dataType, StringComparison.Ordinal)
```
martinothamar: Like this one, we were using `InvariantCultureIgnoreCase`, but some other places where the `Id` is compared we are using defaults

src/Altinn.App.Api/Infrastructure/Telemetry/IdentityTelemetryFilter.cs
```diff
@@ -42,17 +43,17 @@ public void Process(ITelemetry item)
                 int? partyId = ctx.User.GetPartyIdAsInt();
                 int authLevel = ctx.User.GetAuthenticationLevel();
 
-                request.Properties.Add("partyId", partyId.ToString());
-                request.Properties.Add("authLevel", authLevel.ToString());
+                request.Properties.Add("partyId", partyId?.ToString(CultureInfo.InvariantCulture) ?? "");
```
ivarne, line 46: This changes the telemetry property from `null` to `""` when it can't be found?

ivarne (approved):
Sad to see all those `CultureInfo` just polluting our code, but I agree that they kind of have to be there.

src/Altinn.App.Api/Infrastructure/Telemetry/IdentityTelemetryFilter.cs
martinothamar: [`Nullable<T>` returns `""` when there is no value](https://github.com/dotnet/runtime/blob/c897ce62f21b061f8a53f6629fea97526cbb09a8/src/libraries/System.Private.CoreLib/src/System/Nullable.cs#L67)

danielskovli (commented):
LGTM 👍 💯

danielskovli:
Haha, this was already merged. Dang, and I thought I was helping.

martinothamar:
you doubled my confidence 😄

PR #700, opened 2024-06-27
Feature/external api

### LINKED ISSUE #698: Backend endepunkt for å hente data fra eksternt api

cammiida:
## Related Issue(s)
- #698

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/ExternalApiController.cs
+++ b/src/Altinn.App.Api/Controllers/ExternalApiController.cs
@@ -31,6 +31,7 @@ public class ExternalApiController : ControllerBase
     /// <param name="instanceOwnerPartyId">The instance owner party id</param>
     /// <param name="instanceGuid">The instance guid</param>
     /// <param name="externalApiId">The id of the external api</param>
+    /// <param name="queryParams">The query parameters to pass to the external api endpoint</param>
     /// <returns>The data for the external api</returns>
     [HttpGet]
     [Authorize(Policy = AuthzConstants.POLICY_INSTANCE_READ)]
@@ -40,23 +41,27 @@ public class ExternalApiController : ControllerBase
     public async Task<IActionResult> Get(
         [FromRoute] int instanceOwnerPartyId,
         [FromRoute] Guid instanceGuid,
-        [FromRoute] string externalApiId
+        [FromRoute] string externalApiId,
+        [FromQuery] Dictionary<string, string> queryParams
     )
     {
-        _logger.LogInformation("Getting data for external api with id {ExternalApiId}", externalApiId);
-
         var instanceIdentifier = new InstanceIdentifier(instanceOwnerPartyId, instanceGuid);
 
         try
         {
-            var externalApiData = await _externalApiService.GetExternalApiData(externalApiId, instanceIdentifier);
+            var externalApiData = await _externalApiService.GetExternalApiData(
+                externalApiId,
+                instanceIdentifier,
+                queryParams
+            );
             return Ok(externalApiData);
         }
         catch (Exception e)
         {
             if (e is KeyNotFoundException)
             {
-                return NotFound();
+                _logger.LogWarning(e, "External api not found");
+                return StatusCode(404, "External api not found");
             }
 
             return StatusCode(500, "An error occurred while fetching data from an external api.");
--- a/src/Altinn.App.Api/Controllers/ExternalApiController.cs
+++ b/src/Altinn.App.Api/Controllers/ExternalApiController.cs
@@ -1,12 +1,15 @@
+using Altinn.App.Core.Constants;
 using Altinn.App.Core.Features.ExternalApi;
+using Altinn.App.Core.Models;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 
 namespace Altinn.App.Api.Controllers;
 
 /// <summary>
-/// Represents the DataLists API.
+/// Represents the External API Controller.
 /// </summary>
-[Route("{org}/{app}/api/external")]
+[Route("{org}/{app}/instances/{instanceOwnerPartyId:int}/{instanceGuid:guid}/api/external")]
 [ApiController]
 public class ExternalApiController : ControllerBase
 {
@@ -14,7 +17,7 @@ public class ExternalApiController : ControllerBase
     private readonly IExternalApiService _externalApiService;
 
     /// <summary>
-    /// Create new instance of the <see cref="ActionsController"/> class
+    /// Creates a new instance of the <see cref="ExternalApiController"/> class
     /// </summary>
     public ExternalApiController(ILogger<ExternalApiController> logger, IExternalApiService externalApiService)
     {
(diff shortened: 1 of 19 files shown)


src/Altinn.App.Core/Features/ExternalApi/ExternalApiFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <param name="externalApiId"></param>
+    public IExternalApiClient GetExternalApiClient(string externalApiId)
+    {
+        IExternalApiClient? client = _externalApiClients.FirstOrDefault(e =>
+            e.Id.Equals(externalApiId, StringComparison.OrdinalIgnoreCase)
+        );
+
+        if (client == null)
+        {
+            _logger.LogWarning("No external client found for external API with id {ExternalApiId}", externalApiId);
```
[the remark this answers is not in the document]
martinothamar: In theory we could try to limit the accepted size of `externalApiId`, but I don't think this is a big issue 😄

src/Altinn.App.Core/Features/ExternalApi/ExternalApiFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <summary>
+    /// Finds the implementation of <see cref="IExternalApiClient"/> based on the external api id
+    /// </summary>
+    IExternalApiClient GetExternalApiClient(string externalApiId);
+}
+
+/// <summary>
+/// Factory class for resolving <see cref="IExternalApiClient"/> implementations
+/// </summary>
+public class ExternalApiFactory : IExternalApiFactory
```
martinothamar: Could this be `internal sealed`?

src/Altinn.App.Core/Features/ExternalApi/ExternalApiFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+    public IExternalApiClient GetExternalApiClient(string externalApiId)
+    {
+        var client = _externalApiClients.FirstOrDefault(e =>
+            e.Id.Equals(externalApiId, StringComparison.OrdinalIgnoreCase)
+        );
+
+        if (client == null)
+        {
+            _logger.LogWarning("No external client found for external API with id {ExternalApiId}", externalApiId);
+            throw new KeyNotFoundException($"No external api client found for external API with id {externalApiId}");
```
martinothamar: Should we return `null` here instead? Returning `IExternalApiClient?` sSeems simpler than throwing an exception, it's not documented in xmldocs either

martinothamar (commented):
nice cleanup of the policy stuff 🥇

src/Altinn.App.Core/Features/ExternalApi/ExternalApiFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    /// <inheritdoc/>
+    /// <param name="externalApiId"></param>
+    public IExternalApiClient GetExternalApiClient(string externalApiId)
+    {
+        var client = _externalApiClients.FirstOrDefault(e =>
+            e.Id.Equals(externalApiId, StringComparison.OrdinalIgnoreCase)
+        );
+
+        if (client == null)
```
danielskovli: Consider using the `is` operator

src/Altinn.App.Core/Features/ExternalApi/ExternalApiFactory.cs
danielskovli: Should we roll our own specific exception here? I usually do, to make debugging easier down the line, but I don't know if we have a team policy regarding this.

danielskovli: On second glance, I see that you're catching it in the controller. In which case it doesn't really matter what the exception is 🤷

danielskovli (approved):
Looks good! 👍

src/Altinn.App.Core/Features/ExternalApi/ExternalApiFactory.cs
cammiida: Hm, ser ut som det er litt forskjellige meninger på nettet. Blant annet at man eventuelt kan returnere 204. Men om man ikke finner det eksterne api-et er det jo en feil, så syns kanskje ikke 200-serien er helt riktig? 🤔 og ved å returnere `200(null)` så lurer jeg på hvordan brukeren skal vite at et eksternt api med den satte id-en ikke eksisterer og ikke at resultatet av å hente data fra det eksterne api-et er `null`?

cammiida: Whoops engelsk 😅 
With regards to the documentation, do you have any suggestions on how to document it? It's not an exception, so I don't think that would be right. Saw something about a `<response>` tag, but that isn't documentet in the [dotnet documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/) and doesn't work with intellisense.

martinothamar: I was only referring to the `IExternalApiFactory.GetExternalApiClient` method, not the HTTP status code. I see that it is 404 now, but isn't the issue that the input is wrong? Should it be  400? Reasons for my `IExternalApiFactory` suggestion:

* Exceptions tend to be used for exceptional cases (unexpected conditions), is someone trying to refer to a non-existing external API exceptional?
* I think `IExternalApiClient? GetExternalApiClient(string externalApiId);` more clearly communicates the behavior, it let's the caller know just from the method signature that it either returns what was asked for or `null`
* Exceptions are poor for performance

XML doc for exception is `<exception cref="KeyNotFoundException">some message</exception>`

cammiida: Ahh, I see I was looking at the controller, not the factory. My bad! I think it makes sense to return null in the factory, but still return 404 in the controller. Did some changes according to that [in the last commit](https://github.com/Altinn/app-lib-dotnet/pull/700/commits/69464a2b765dfa75b47dcf718890159ed2bd7b06) if you'd like to check it out 😊

src/Altinn.App.Core/Features/ExternalApi/ExternalApiService.cs
```diff
@@ -0,0 +1,51 @@
...
+        string externalApiId,
+        InstanceIdentifier instanceIdentifier,
+        Dictionary<string, string> queryParams
+    )
+    {
+        var externalApiClient = _externalApiFactory.GetExternalApiClient(externalApiId);
+        if (externalApiClient is null)
+        {
+            _logger.LogWarning("External api with id {ExternalApiId} not found", externalApiId);
+            throw new KeyNotFoundException($"External api with id {externalApiId} not found");
```
danielskovli: If I understood Martin correctly, he suggested returning null here if the client is not found, instead of throwing.

src/Altinn.App.Api/Controllers/ExternalApiController.cs
```diff
@@ -0,0 +1,70 @@
...
+        try
+        {
+            var externalApiData = await _externalApiService.GetExternalApiData(
+                externalApiId,
+                instanceIdentifier,
+                queryParams
+            );
+            return Ok(externalApiData);
+        }
+        catch (Exception e)
```
danielskovli: Coupled with my other comment, a null check here instead of a catch.

danielskovli (commented):
A couple of notes on exceptions vs. nullable, as per discussion above.

PR #708, opened 2024-07-09
Return 400 on wrong external api client id

cammiida:
## Description
Changing http status code to return 400 instead of 404 when supplied external api client id does not correspond to an existing implementation of `IExternalApiClient`.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


danielskovli (approved):
👍

PR #710, opened 2024-07-10
Intercepts unhandled exceptions, returns somewhat descriptive message to caller

danielskovli:
Minor adjustments to the error handling in `ExternalApiController`.

Catching unhandled errors and specifically returning a 500 status code to the caller, with what we know about the error; preventing exceptions from being automatically handled by the middleware.


cammiida (approved):
Gitt at alt kjører grønt og sånt så ser det veldig bra ut for meg!  🎸 ✨

PR #730, opened 2024-08-19
Support Extra data models and subforms in validation

ivarne:
Work on supporting multiple data models and subforms in backend validations with expressions.

## Related Issue(s)
- #600 
- replaces https://github.com/Altinn/app-lib-dotnet/pull/701
- replaces https://github.com/Altinn/app-lib-dotnet/pull/619

## Main changes
The main goal for this PR is to add 

### New single interface for validation `IValidator`
With changes in one data element affects the validity of another element, the current hierarchy of validators and only `IFormDataValidator` supporting incremental validation break down. Adaptors for the old interfaces has been provided so no changes to current apps are required, but they might want to update in order to support incremental validation and read instance data with the `IInstanceDataAccessor`.

### New interface `IInstanceDataAccessor`
When validation has dependencies across data models, we need a cached way to access the content of data elements that does not require us to download the same data element multiple times for the same request. The implementation is not available from DI, but always passed as an argument in order to eliminate risk of it being reused cross requests because of Singleton services or other unintentionally shared global state.

### New `DataElementId` struct to ensure type safety for a wrapped Guid
36 places in the updated code need to reference the Guid of a data element, and adding type safety is an important step to ensure correctness. An implicit cast is provided from `DataElement` to simplify usage

### New `ModelBinding` struct in expressions engine
To support the new syntax for referring to a non default data element in components and expressions
```json
{
    "dataModelBindings": {"field": "path.to.data", "dataType": "default"}
    "hidden": ["equals", null, ["dataModel", "path.to.other", "prefill"]]
}
```

When handling references to a specific fields (eg. for removal) we needed to also add `DataReference` to combine a Field (path) with a specific DataElementId

### New PATCH api endpoint that allows frontend to update multiple data models at the same time
Saving and triggering incremental validation is expensive and some apps might want to batch multiple changes to reduce cost. The code in frontend would also be simplified if it could do a single request, instead of multiple.

### `ValidationIssueWithSource` class for external apis
Previously we had a single class `ValidationIssue` that included a `Source` property. Service owners sometimes tries to set this in their validators (because it shows up in intellisense), but as we override it based on `IValidator.ValidationSource` in our `ValidationService` it would be good if it wasn't there to confuse. This PR creates a new type `ValidationIssueWithSource` for external apis (where the source is included) and marks it `[Obsolete]` in the type returned in the validator interfaces.

### Respect new app logic settings for data elements `DisallowUserCreate` and `DisallowUserDelete`
This means that users can not create or delete this types. Typically used for prefill, when you want only the org token to be able to modify the number of data elements on an instance.

## Minor changes
Used IL code to generate real C# classes based on the json data models in shared tests. Previously different code was used to access properties of the test code as in production, which is not really good for tests. 

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -322,7 +321,7 @@ private static void AddExternalApis(IServiceCollection services)
 
     private static void AddProcessServices(IServiceCollection services)
     {
-        services.TryAddTransient<IProcessEngine, ProcessEngine>();
+        services.TryAddScoped<IProcessEngine, ProcessEngine>();
```
martinothamar: Seems like this will break 1 app - digdir/oed

[DeceasedStateMiddleware](https://altinn.studio/repos/digdir/oed/src/commit/67bd185a198cbf380d72c173ae4fd73edc8ea953/App/Middlewares/DeceasedStateMiddleware.cs#L25) -> [StorageService](https://altinn.studio/repos/digdir/oed/src/commit/67bd185a198cbf380d72c173ae4fd73edc8ea953/App/Services/StorageService.cs#L38)

Since that type of middleware is a singleton, and they don't inject from the HttpContext. Should we just remember to notify them when we release? I don't think any of the other DI changes here affect anyone

ivarne: I reverted the `AddScoped` changes. They are no longer needed after I rejected the idea of caching data element downloads using a scoped service.

I opened https://github.com/Altinn/app-lib-dotnet/issues/751, because I think we should default more to using scoped services.

src/Altinn.App.Core/Internal/Validation/IValidationService.cs
```diff
@@ -20,52 +20,31 @@ public interface IValidationService
...
-        DataElement dataElement,
-        DataType dataType,
+        string taskId,
+        IInstanceDataAccessor dataAccessor,
         string? language
     );
 
     /// <summary>
-    /// Validates a single data element. Used by frontend to continuously validate form data as it changes.
+    ///
```
martinothamar: Would be nice with a doc here

src/Altinn.App.Core/Internal/Validation/ValidationService.cs
```diff
@@ -26,251 +19,136 @@ public class ValidationService : IValidationService
...
-            dataElementsToValidate.Select(dataElement =>
-                ValidateDataElement(
-                    instance,
-                    dataElement,
-                    dataTypesForTask.First(dt => dt.Id == dataElement.DataType),
-                    language
-                )
-            )
-        );
+        var validators = _validatorFactory.GetValidators(taskId);
```
martinothamar, line 46: The comment here says "run", but the select isn't materialized

src/Altinn.App.Core/Models/Validation/ValidationIssueWithSource.cs
```diff
@@ -0,0 +1,90 @@
...
+
+/// <summary>
+/// Represents a detailed message from validation.
+/// </summary>
+public class ValidationIssueWithSource
+{
+    /// <summary>
+    /// Converter function to create a <see cref="ValidationIssueWithSource"/> from a <see cref="ValidationIssue"/> and adding a source.
+    /// </summary>
+    public static ValidationIssueWithSource FromIssue(ValidationIssue issue, string source)
```
martinothamar: Really like this design! 🥂 
Could this (or even the whole class) be made internal?

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -460,48 +519,76 @@ public async Task<ActionResult<DataPatchResponse>> PatchFormData(
             if (!InstanceIsActive(instance))
             {
                 return Conflict(
-                    $"Cannot update data element of archived or deleted instance {instanceOwnerPartyId}/{instanceGuid}"
+                    new ProblemDetails()
```
martinothamar: I like that we use problem details more often, with the OTel setup this makes error responses a lot easier to debug 👏

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -164,78 +162,89 @@ [FromQuery] string dataType
...
             }
 
-            if (appLogic)
+            int existingElements = instance.Data.Count(d => d.DataType == dataTypeFromMetadata.Id);
+            if (dataTypeFromMetadata.MaxCount > 0 && existingElements >= dataTypeFromMetadata.MaxCount)
             {
-                return await CreateAppModelData(org, app, instance, dataType);
+                return Conflict(
+                    $"Element type `{dataType}` has reached its maximum allowed count ({dataTypeFromMetadata.MaxCount})"
+                );
```
martinothamar: @danielskovli was this the error condition that was missing before? Can we make the response type `ProblemDetails` in that case?

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -452,6 +464,53 @@ public async Task<ActionResult<DataPatchResponse>> PatchFormData(
...
     )
+    {
+        var request = new DataPatchRequestMultiple()
+        {
+            Patches = new() { [dataGuid] = dataPatchRequest.Patch },
+            IgnoredValidators = dataPatchRequest.IgnoredValidators
+        };
+        var response = await PatchFormDataMultiple(org, app, instanceOwnerPartyId, instanceGuid, request, language);
+
+        if (response.Result is OkObjectResult { Value: DataPatchResponseMultiple newResponse })
```
martinothamar, line 422: Is it possible to just check `response.Value` instead? The return from `PatchFormDataMultiple` might change to `StatusCode(200, ...)` for example which is also a valid `ActionResult<>` but then `.Result` is not this object. Feels kind of brittle

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -452,6 +464,53 @@ public async Task<ActionResult<DataPatchResponse>> PatchFormData(
...
+    /// <param name="instanceOwnerPartyId">unique id of the party that is the owner of the instance</param>
+    /// <param name="instanceGuid">unique id to identify the instance</param>
+    /// <param name="dataPatchRequest">Container object for the <see cref="JsonPatch" /> and list of ignored validators</param>
+    /// <param name="language">The language selected by the user.</param>
+    /// <returns>A response object with the new full model and validation issues from all the groups that run</returns>
+    [Authorize(Policy = AuthzConstants.POLICY_INSTANCE_WRITE)]
+    [HttpPatch("")]
+    [ProducesResponseType(typeof(DataPatchResponseMultiple), 200)]
+    [ProducesResponseType(typeof(ProblemDetails), 409)]
+    [ProducesResponseType(typeof(ProblemDetails), 422)]
```
martinothamar, line 452: Codepath below produces 404 and 400 as well I think

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -556,23 +643,27 @@ [FromRoute] Guid dataGuid
...
                 _logger.LogError(errorMsg);
                 return BadRequest(errorMsg);
             }
-            else if (dataType.AppLogic?.ClassRef is not null)
+
+            if (
+                dataType.AppLogic?.ClassRef is not null
+                && dataType.AppLogic.DisallowUserDelete
+                && !UserHasValidOrgClaim()
+            )
```
martinothamar: Is this a change in behavior, or does `DisallowUserDelete` default to `true`?

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -963,32 +1054,18 @@ await _instanceClient.UpdateDataValues(
...
 
     private static bool InstanceIsActive(Instance i)
     {
-        if (i?.Status?.Archived != null || i?.Status?.SoftDeleted != null || i?.Status?.HardDeleted != null)
-        {
-            return false;
-        }
-
-        return true;
+        return i?.Status?.Archived is null && i?.Status?.SoftDeleted is null && i?.Status?.HardDeleted is null;
```
martinothamar: With the `?` this reads a little funny. If `i` or `i.Status` are null then the instance is active?

src/Altinn.App.Core/Features/Telemetry.InterfaceFactory.cs
```diff
@@ -0,0 +1,11 @@
+using System.Diagnostics;
+
+namespace Altinn.App.Core.Features;
+
+public partial class Telemetry
```
martinothamar: Daniel recently created the `Telemetry` folder

ivarne: Oops, this file was from a different PR

src/Altinn.App.Api/Controllers/DataController.cs
ivarne: I'm open to suggestions, but I'm pretty sure my first instinct was to write `response.Value` and it did not work. Fortunately this code has tests, so if it breaks, we'll know about it. The correct way would probably be to add a helper method for both controller methods without the `ActionResult<>` mess.

src/Altinn.App.Core/Models/Validation/ValidationIssueWithSource.cs
ivarne: It is part of the public api in controllers, so I don't think so.

src/Altinn.App.Core/Internal/Validation/ValidationService.cs
ivarne: OK, comment reworded and moved

src/Altinn.App.Core/Internal/Validation/IValidationService.cs
ivarne: Fixed.

src/Altinn.App.Api/Controllers/ValidateController.cs
```diff
@@ -67,9 +78,13 @@ public async Task<IActionResult> ValidateInstance(
 
         try
         {
-            List<ValidationIssue> messages = await _validationService.ValidateInstanceAtTask(
+            var dataAccessor = new CachedInstanceDataAccessor(instance, _dataClient, _appMetadata, _appModel);
+            var ignoredSources = ignoredValidators?.Split(',').ToList();
```
martinothamar, line 90: Should we validate somewhere that these exist/are wellformed?

src/Altinn.App.Core/Features/Telemetry/Telemetry.Validation.cs
```diff
@@ -18,52 +18,32 @@ private void InitValidation(InitContext context) { }
...
+        var activity = ActivitySource.StartActivity($"{Prefix}.ValidateIncremental");
+        activity?.SetTaskId(taskId);
         activity?.SetInstanceId(instance);
-        activity?.SetDataElementId(dataElement);
+        // Log the IDs of the elements that have changed together with their data type
+        // default:123-678-8900-54,group:123-678-8900-55
+        activity?.SetTag(
+            InternalLabels.ValidatorChangedElementsIds,
+            string.Join(',', changes.Select(c => $"{c.DataElement.DataType}:{c.DataElement.Id}"))
+        );
```
martinothamar: Should we use span events instead here? My current understanding is that tags/attributes are mostly suited for short/low-cardinality and thus filterable/indexable data while span events are must "structured logs"-like data colocated with the span

src/Altinn.App.Core/Features/Validation/Default/ExpressionValidator.cs
```diff
@@ -67,82 +100,109 @@ public async Task<List<ValidationIssue>> ValidateFormData(
...
+            var baseField = new DataReference() { Field = validationObject.Key, DataElementId = dataElementId };
+            var resolvedFields = await evaluatorState.GetResolvedKeys(baseField);
             var validations = validationObject.Value;
             foreach (var resolvedField in resolvedFields)
             {
-                if (hiddenFields.Contains(resolvedField))
+                if (
+                    hiddenFields.Exists(d =>
+                        d.DataElementId == dataElementId
+                        && resolvedField.Field.StartsWith(d.Field, StringComparison.InvariantCulture)
```
martinothamar, line 133: Is `StartsWith` correct here?

src/Altinn.App.Api/Controllers/DataController.cs
danielskovli: Yeah this is the check we discussed a while back 👍 

I think we can make the response whatever you/we want. I only did a 409 because it sort-of-almost followed the established pattern from this controller.

The frontend is specifically catching this response to display something other than "Oooops", but that check can easily be rewritten. It's not perfect as-is anyway, because it it assumes all 409 responses on `Create` is because of maxcount: https://github.com/Altinn/app-frontend-react/blob/feature/subform/src/features/subformData/useSubformMutations.tsx#L33

src/Altinn.App.Api/Controllers/DataController.cs
danielskovli: Defaults to no value, eg `false`: https://github.com/Altinn/altinn-storage/blob/663f4b71747ab721b663d802c8506345be0eaa1f/src/Storage.Interface/Models/ApplicationLogic.cs#L48

src/Altinn.App.Core/Internal/Expressions/LayoutEvaluatorStateInitializer.cs
```diff
@@ -9,34 +12,101 @@ namespace Altinn.App.Core.Internal.Expressions;
...
+            return Task.FromResult<object?>(_data);
+        }
+    }
+
     /// <summary>
     /// Initialize LayoutEvaluatorState with given Instance, data object and layoutSetId
     /// </summary>
-    public virtual Task<LayoutEvaluatorState> Init(
+    [Obsolete("Use the overload with ILayoutEvaluatorStateInitializer instead")]
+    public async Task<LayoutEvaluatorState> Init(
```
martinothamar, line 135: So if I understand correctly this and the code above is kept for compatibility, should we still have some tests for this to make sure it still works? From what I can tell it's only SSB that are using this to acccess `RemoveHiddenData`, so maybe just a test for that?

src/Altinn.App.Core/Models/Layout/PageComponentConverter.cs
```diff
@@ -349,6 +374,19 @@ private BaseComponent ReadComponent(ref Utf8JsonReader reader, JsonSerializerOpt
...
+                    layoutSetId = reader.GetString();
+                    break;
+                // case "tablecolumns":
+                //     tableColumns = JsonSerializer.Deserialize<List<SubFormComponent.TableColumn>>(ref reader, options);
+                //     break;
+                // case "showaddbutton":
+                //     showAddButton = reader.TokenType != JsonTokenType.False;
+                //     break;
+                // case "showdeletebutton":
+                //     showDeleteButton = reader.TokenType != JsonTokenType.False;
```
martinothamar, line 389: What are these commented lines for?

src/Altinn.App.Api/Controllers/ValidateController.cs
ivarne: I don't think so. The requirement for these is that frontend can have a list of validators that it knows (thinks) it has a good client side implementation for. It would be lots of extra code to let frontend know if these validators are really used.

src/Altinn.App.Core/Features/Telemetry/Telemetry.Validation.cs
ivarne: Done. I did not know it was possible.

src/Altinn.App.Core/Features/Validation/Default/ExpressionValidator.cs
ivarne: I'm pretty confident it is, because the only situation when a field would be a prefix of another field is if it is a repeating group that has a `hidden` or a `hiddenRow` property.

src/Altinn.App.Core/Internal/Expressions/LayoutEvaluatorStateInitializer.cs
ivarne: Yes, we should.

src/Altinn.App.Core/Models/Layout/PageComponentConverter.cs
ivarne: For the properties that frontend uses, but I realised I did not need for any purpose in backend. If we need them in the future it would be easier to uncomment than to write them from scratch.

ivarne (commented):
Æsj, svar på kommentarer grupperte seg sammen til en review og ble ikke sendt

src/Altinn.App.Api/Controllers/DataController.cs
ivarne: Fixed

src/Altinn.App.Core/Internal/Data/CachedFormDataAccessor.cs
```diff
@@ -0,0 +1,159 @@
...
+using Altinn.Platform.Storage.Interface.Models;
+
+namespace Altinn.App.Core.Internal.Data;
+
+/// <summary>
+/// Class that caches form data to avoid multiple calls to the data service for a single validation
+///
+/// Do not add this to the DI container, as it should only be created explicitly because of data leak potential.
+/// </summary>
+internal sealed class CachedInstanceDataAccessor : IInstanceDataAccessor
```
martinothamar: Rename the file to match the class name?

src/Altinn.App.Core/Internal/Data/CachedFormDataAccessor.cs
```diff
@@ -0,0 +1,159 @@
...
+    private sealed class LazyCache<TKey, TValue>
+        where TKey : notnull
+        where TValue : notnull
+    {
+        private readonly ConcurrentDictionary<TKey, Lazy<Task<TValue>>> _cache = new();
+
+        public async Task<TValue> GetOrCreate(TKey key, Func<TKey, Task<TValue>> valueFactory)
+        {
+            Task<TValue> task;
+            lock (_cache)
```
martinothamar: What's the reasoning for using both a `ConcurrentDictionary` and a `lock`?

src/Altinn.App.Core/Internal/Process/ProcessTasks/Common/ProcessTaskFinalizer.cs
```diff
@@ -54,58 +47,107 @@ public async Task Finalize(string taskId, Instance instance)
...
+                    data,
+                    Guid.Parse(instance.Id.Split('/')[1]),
+                    data.GetType(),
+                    instance.Org,
+                    instance.AppId.Split('/')[1],
+                    int.Parse(instance.InstanceOwner.PartyId, CultureInfo.InvariantCulture),
+                    Guid.Parse(dataElement.Id)
+                );
+            })
+        );
```
martinothamar: I think the other places `CachedInstanceDataAccessor` is used, the cache is updated after updates/modifications, but not here, maybe because it isn't needed. If it is true that when using `CachedInstanceDataAccessor` and making data updates, we should always make sure the cache is in sync, can we write the code such that updating data always updates  the cache or vice versa?

src/Altinn.App.Core/Internal/Data/CachedFormDataAccessor.cs
ivarne: https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-8.0#system-collections-concurrent-concurrentdictionary-2-addorupdate(-0-system-func((-0-1))-system-func((-0-1-1)))
> If you call [AddOrUpdate](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-8.0) simultaneously on different threads, addValueFactory may be called multiple times, but its key/value pair might not be added to the dictionary for every call.
For modifications and write operations to the dictionary, [ConcurrentDictionary<TKey,TValue>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-8.0) uses fine-grained locking to ensure thread safety (read operations on the dictionary are performed in a lock-free manner). The addValueFactory and updateValueFactory delegates may be executed multiple times to verify the value was added or updated as expected. However, they are called outside the locks to avoid the problems that can arise from executing unknown code under a lock. Therefore, [AddOrUpdate](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-8.0) is not atomic with regards to all other operations on the [ConcurrentDictionary<TKey,TValue>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-8.0) class.

ivarne: I suggested this in the maskinporten cache, but you wrote some wizard code that actually trigged that race condition. As I (think i) know the Lazy constructor is reasonably fast, there is no reason not to lock to be extra sure.

src/Altinn.App.Core/Internal/Data/CachedFormDataAccessor.cs
ivarne: 😊

src/Altinn.App.Core/Internal/Process/ProcessTasks/Common/ProcessTaskFinalizer.cs
ivarne: If you just mutate the data models, the cache will be in sync. The issue is if you add or remove data elements from the instance, or if you want to replace the root data object in the cache.

We actually have such a modification when removing shadow fields (or making a copy of the data element without shadow fields)

https://github.com/Altinn/app-lib-dotnet/blob/450f53d978f52e65c275b8146d4c6ac6ebea1c55/src/Altinn.App.Core/Internal/Process/ProcessTasks/Common/ProcessTaskFinalizer.cs#L157-L185

Maybe `IInstanceDataProcessor` should have apis to mutate Instance.Data and sync with both local state and storage?

src/Altinn.App.Api/Controllers/DataController.cs
martinothamar: Yeah I think 409 is good, I was only thinking about the string value being converted to `ProblemDetails` as we have middleware that annotates the ASP.NET Core trace with additional information when the response type is `ProblemDetails` (which means easier debugging)

src/Altinn.App.Core/Internal/Data/CachedFormDataAccessor.cs
martinothamar: Right, I just meant that you could use a normal `Dictionary<,>` since there is a "global" mutex wrapping the operations anyway

ivarne: Yes I could, but ConcurrentDictionary sounds like exactly what I need.

Apart from needing Lazy<> and a lock.

src/Altinn.App.Api/Controllers/DataController.cs
danielskovli: @martinothamar Unsurprisingly I don't know exactly what this should look like. But I can confirm that the frontend does check the body payload at all, only the response status code.

So if you're feeling inspired, please make whatever changes you think are necessary :)

src/Altinn.App.Api/Controllers/DataController.cs
martinothamar: So in 
v8.3.5 - deleting form data is disallowed
v8.4.0 - deleting form data is allowed by default (so users who want same behavior as in v8.3.5 need to "opt in"?)

Has the change in default behavior been discussed? Not sure if this will affect anyone

danielskovli: @martinothamar You could argue that `8.3.5` breaks with previous versions, and `8.4.0` restores the established behavior. 

The change has been discussed in the context of a subform, and agreed that this new implementation is desired.

Do as you will with that information 🙂

danielskovli: Hmm I'm actually not sure about the versioning here, so don't read too much into that. But the essence of the situation is that I introduced a somewhat-breaking change (for API users), which Ivar has reverted by inverting the logic in DataController

src/Altinn.App.Core/Internal/Data/CachedFormDataAccessor.cs
ivarne: Changed to using regular Dictionary, as it is protected by a lock

src/Altinn.App.Core/Internal/Process/ProcessTasks/Common/ProcessTaskFinalizer.cs
ivarne: This has been resolved by splitting the interface into `IInstanceDataAccessor` and `IInstanceDataMutator`. One for validations, and the mutator for actions and data processing.
 CachedInstanceDataAccessor implements both and can be used in both

src/Altinn.App.Api/Controllers/DataController.cs
martinothamar: Pushed a commit for this

ivarne: Thanks!

src/Altinn.App.Core/Helpers/Serialization/ModelSerializationService.cs
```diff
@@ -0,0 +1,265 @@
...
+        }
+        try
+        {
+            using XmlTextReader xmlTextReader = new XmlTextReader(new StringReader(streamContent));
+            XmlSerializer serializer = _xmlSerializer.GetSerializer(modelType);
+
+            return serializer.Deserialize(xmlTextReader)
+                ?? throw new InvalidOperationException("Deserialization returned null");
+        }
+        catch (InvalidOperationException)
```
HauklandJ, line 185: I stedet for å bruke exceptions til å styre control flow, kunne vi gjort noe som:
```
    bool containsNamespace = CheckIfNamespaceExists(streamContent);
    XmlSerializer serializer = containsNamespace
        ? _xmlSerializer.GetSerializer(modelType)
        : _xmlSerializer.GetSerializerIgnoreNamespace(modelType);
```

src/Altinn.App.Core/Helpers/Serialization/ModelSerializationService.cs
```diff
@@ -0,0 +1,265 @@
...
+    /// </summary>
+    public object DeserializeXml(ReadOnlySpan<byte> data, Type modelType)
+    {
+        using var activity = _telemetry?.StartDeserializeFromXmlActivity(modelType);
+        // convert to UTF16 string as it seems to be preferred by the XmlTextReader
+        // and aligns with previous implementation
+        string streamContent = Encoding.UTF8.GetString(data.RemoveBom());
+        if (string.IsNullOrWhiteSpace(streamContent))
+        {
+            throw new Exception("No XML content read from stream");
```
HauklandJ: Bruke ArgumentException?

HauklandJ (approved):
LGTM, some minor things to consider

src/Altinn.App.Core/Helpers/Serialization/ModelSerializationService.cs
ivarne: Jeg skulle gjerne gjort det, men jeg aner ikke hvordan jeg implementerer `CheckIfNamespaceExists(stream)`, og dette er kode jeg har kopiert fra slik vi gjorde ting tidligere.

src/Altinn.App.Core/Helpers/Serialization/ModelSerializationService.cs
ivarne: Vet ikke helt om jeg ser poenget gitt at vi er enige om å ikke styre controlFlow med exceptions.
```suggestion
            throw new ArgumentException("No XML content read from stream");
```

src/Altinn.App.Core/Internal/Expressions/LayoutEvaluatorStateInitializer.cs
ivarne: Added in aee32775a664c5117c2fd9f9bb5639f8fcea1de9

PR #732, opened 2024-08-21
Feat/679 named pdf prereqs

HauklandJ:
Div. changes for the named pdf issue which are not related to the new feature, but rather general cleanup/changes in preparation.

Split out this pr so that it does not pollute the actual changes.

## Description
* LanguageConst from Altinn.App.Core.Internal.Language rather than hardcoded strings everywhere.
* Move httpcontext out of language helper methods. Would like to move it out of the service all together, but having it in random GetLanguage helper method is "too deep into the [onion](https://dzone.com/articles/onion-architecture-is-interesting) arch" for my taste.
* Underscore prefix for fields.
* Additional tests on current functionality.

## Related Issue(s)
- #{issue number}

## Verification
- [x] All tests run green

## Documentation


ivarne (approved):
> LanguageConst from Altinn.App.Core.Internal.Language rather than hardcoded strings everywhere.

Yes, you can use `Bokmål` as a variable, but why tempt faith that nobody messes up the encoding when `LanugageConsts.Nb` works just fine? In addition, most usages (outside tests) just refer to a convenient default for Norway, but one of the things that needs to be configurable for international usage (if that ever happens).

> Move httpcontext out of language helper methods. Would like to move it out of the service all together, but having it in random GetLanguage helper method is "too deep into the [onion](https://dzone.com/articles/onion-architecture-is-interesting) arch" for my taste.

Totally agree. @bjorntore needs to add `?language` parameter to the process interfaces, so we can just read it in the controller.

> Underscore prefix for fields.

😿 I did this too, Hope the merge conflict isn't too confusing

> Additional tests on current functionality.

Looks good, but these tests will fail when we pass the language properly through the process engine.

PR #739, opened 2024-08-29
Add external api ids to application metadata

cammiida:
## Description
Adds the ids of the registered external APIs to application metadata. This is because the expression engine in the frontend has to know all the data to resolve in advance. At least the way it is working right now.

## Related Issue(s)
- #736 

## Verification
- [x] All tests run green

## Documentation


danielskovli:
@cammiida  Just a quick question about DI: Why did you swap the direct injection of `IExternalApiFactory` in favor of `IServiceProvider`?

cammiida:
Good question @danielskovli ! As far as I've understood, this is the way to do it if an injected interface is less accessible than the class it is injected in. In this case `AppMetadata` is `public` and `IExternalApiFactory` is `internal`. But happy to hear if you have any thoughts or suggestions 😃

danielskovli:
@cammiida Ah, I see. I have no opinion on that 😂 If it works as you want it to work in your testing, that's good enough for me.

src/Altinn.App.Core/Models/ApplicationMetadata.cs
```diff
@@ -70,4 +70,10 @@ public ApplicationMetadata(string id)
     /// </summary>
     [System.Text.Json.Serialization.JsonExtensionData]
     public Dictionary<string, object>? UnmappedProperties { get; set; }
+
```
danielskovli, line 73: This comment is for line 66 (not your code), because GH static code analysis triggers on it:

I would argue it's not breaking to change this `Exception` to something more specific, since that will still trigger a catch on `Exception` for callers. So if you're motivated to fix  this, even though it's not your code, please do 🙂  🙏

danielskovli (approved):
👍

src/Altinn.App.Core/Features/ExternalApi/ExternalApiFactory.cs
```diff
@@ -3,12 +3,14 @@ namespace Altinn.App.Core.Features.ExternalApi;
 /// <summary>
 /// Interface of factory class for resolving <see cref="IExternalApiClient"/> implementations
 /// </summary>
-public interface IExternalApiFactory
+internal interface IExternalApiFactory
```
martinothamar, line 6: Has this been released as a public interface already? If so we should probably run a search to see if anyone references it, just to be sure

bjorntore: No hits using code search in Gitea. 
[Le search](https://altinn.studio/repos/explore/code?q=IExternalApiFactory&fuzzy=true)

src/Altinn.App.Core/Models/ApplicationMetadata.cs
danielskovli: Placeing this in the too-hard basket for now

PR #742, opened 2024-09-03
Make sure exception is thrown when an exclusive gateway fails to return a valid path in the process engine

martinothamar:
## Description
A hard to diagnose bug occurred where due to a logic error in bpmn `conditionExpressions` there were some `process/next` invocations that returned the same task (which happens any time `GetNextTask` returns null).

* Exclusive gateways should only return 1 path AFAICT
* When `GetNextTask` resolves to null, we need to handle it in some way (is this always a logic error on the part of the app bpmn?)

## Related Issue(s)
- Lots

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Process/ProcessEngine.cs
+++ b/src/Altinn.App.Core/Internal/Process/ProcessEngine.cs
@@ -306,7 +306,11 @@ public class ProcessEngine : IProcessEngine
         }
 
         // ending process if next element is end event
-        var nextElementId = nextElement?.Id;
+        if (nextElement is null)
+        {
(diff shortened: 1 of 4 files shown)


bjorntore (approved):

ivarne:
There seems to be some logic to throw if there isn't any return, but as far as I can tell, it won't trigger 
https://github.com/Altinn/app-lib-dotnet/blob/eaf0af863a653aac2acceb7f13fd6c13c976f96e/src/Altinn.App.Core/Internal/Process/ProcessEngine.cs#L327-L329

PR #744, opened 2024-09-04
Unlock dataelements for a task if eFormidling service task fails

martinothamar:
## Description
Just another failure condition the process engine doesn't handle at all. This isn't great, but it fixes 1 of the issues I've seen resulting in locked dataelements that users can't proceed from

## Related Issue(s)
- Support

## Verification
- [ ] All tests run green

## Documentation


bjorntore (approved):
If PDF succedes but eformidling fails, will another PDF be generated the next time process next is run? Will old PDFs be cleaned up first?

bjorntore:
@RonnyB71 This won't cause any trouble that you know of? Can't recall why we didn't do this for eFormidling when we did it for PDF.

ivarne (approved):

martinothamar:
Tested the PDF-case manually - it is just recreated

PR #746, opened 2024-09-04
Add ProblemDetails information to trace for improved telemetry

martinothamar:
## Description
Adds some information to the root request trace when an action returns `ProblemDetails`

Considerations:
* Use span attributes/tags or span events here?

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Features/Telemetry/TelemetryActivityExtensions.cs
+++ b/src/Altinn.App.Core/Features/Telemetry/TelemetryActivityExtensions.cs
@@ -1,6 +1,7 @@
 using System.Diagnostics;
 using Altinn.App.Core.Models.Process;
 using Altinn.Platform.Storage.Interface.Models;
+using Microsoft.AspNetCore.Mvc;
 using OpenTelemetry.Trace;
 using InternalLabels = Altinn.App.Core.Features.Telemetry.InternalLabels;
 using Labels = Altinn.App.Core.Features.Telemetry.Labels;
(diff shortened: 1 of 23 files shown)


src/Altinn.App.Core/Features/Telemetry/TelemetryActivityExtensions.cs
```diff
@@ -265,6 +266,21 @@ public static Activity SetOrganisationNumber(this Activity activity, string? org
...
+    {
+        // Leave activity status to ASP.NET Core, as it will be set depending on status code?
+        // activity.SetStatus(ActivityStatusCode.Error, problemDetails.Title);
+
+        activity.SetTag(InternalLabels.ProblemType, problemDetails.Type);
+        activity.SetTag(InternalLabels.ProblemTitle, problemDetails.Title);
+        activity.SetTag(InternalLabels.ProblemStatus, problemDetails.Status);
+
+        // Is it safe to use detail here? In some cases the detail field is set from exception message
+        // activity.SetTag(InternalLabels.ProblemDetail, problemDetails.Detail);
```
ivarne, line 279: Vi logger jo vanligvis exceptions, så usikker på når dette skulle være et problem.

ivarne (approved):
Dette blir jo litt dobbelt i forhold til at vi ofte logger feil før vi returnerer `ProblemDetails`. I prinsipp burde vi vel bare logge det en gang.

martinothamar:
In the majority of cases we don't log, and span attributes/events has the benefit of living closer to the trace. Sampling mechanics are also different between traces and logs I think, at least I've found it a bit inconsistent sometimes when debugging (expecting there to be a related log message but then there isn't one). Span events are more comparable to log messages, but they are more structured, aren't indexed (cheaper to store in the infra I'm aware of). Let's discuss this with the team today

PR #756, opened 2024-09-10
Make MaskinportenSettings configuration path configurable

### LINKED ISSUE #755: Make `MaskinportenSettings`configuration path configurable
## Description
The configuration name for the `MaskinportenSettings` object is currently hardcoded at "MaskinportenSettings" in [AddMaskinportenClient](https://github.com/Altinn/app-lib-dotnet/blob/3e8be7c093c2596f43defcb7e0b4510eb1bf7848/src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs#L141). Vi have the ability to override this via the [ConfigureMaskinportenClient](https://github.com/Altinn/app-lib-dotnet/blob/3e8be7c093c2596f43defcb7e0b4510eb1bf7848/src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs#L119C38-L119C65) method, but this is perhaps a bit heavy handed in some situations.

Some service owners have many apps running in the same cluster, with the same KeyVault, but require each app to have their own Maskinporten client setup. Which means that the KeyVault entries and appsettings.json configurations are typically prefixed with a unique string (often the app name).

Eg:

`appsettings.json`
```
"appname-MaskinportenSettings": {
    "authority": "https://test.maskinporten.no/",
    "clientId": "",
    "jwkBase64": ""
}
```

`keyvault`
```
appname-MaskinportenSettings--ClientId = xxx
appname-MaskinportenSettings--EncodedJwk = xxx
```

It would be optimal if we could support an app binding to `appname-MaskinportenSettings` instead of the default hardcoded value.

The manual config done through `ConfigureMaskinportenClient` can possibly service this purpose, but moves a lot of the configuration out of appsettings.json and in to the codebase itself. Unless there are clever patterns available there to inject values as needed.

## Proposed implementation
Create a public method `ServiceCollectionExtensions.ConfigureMaskinportenClient` that takes a `string` parameter to pass to `IServiceCollection.BindConfiguration`.

Calling this method before `services.AddAltinnAppServices` in Program.cs will result in `ServiceCollectionExtensions.AddMaskinportenClient` using the newly provided configuration binding instead of the default `MaskinportenSettings`.

danielskovli:
## Description
closes #755

## Related Issue(s)
- #755

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
@@ -127,19 +127,36 @@ public static class ServiceCollectionExtensions
     }
 
     /// <summary>
+    /// <para>
+    /// Binds a <see cref="MaskinportenClient"/> configuration to the supplied config section path.
+    /// </para>
+    /// <para>
+    /// If you have already provided a <see cref="MaskinportenSettings"/> configuration, either manually or
+    /// implicitly via <see cref="WebHostBuilderExtensions.ConfigureAppWebHost"/>, this will be overridden.
+    /// </para>
+    /// </summary>
+    /// <param name="services">The service collection</param>
+    /// <param name="configSectionPath">The configuration section path (Eg. "MaskinportenSettings")</param>
+    public static IServiceCollection ConfigureMaskinportenClient(
+        this IServiceCollection services,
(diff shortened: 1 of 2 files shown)


martinothamar (approved):
🥇

PR #773, opened 2024-09-12
Chore/ensure enum serialization

### LINKED ISSUE #729: Enforce enum serialization policy
As discovered in [2309](https://github.com/Altinn/app-frontend-react/pull/2309). If an application chooses to specify `JsonStringEnumConverter` for `JsonSerializerOptions` then enums which are implemented with the [default](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum#conversions) serialization behaviour will result in unexpected behaviour. 

We should consider enforcing serialization rules for these enums.

HauklandJ:
## Description
Find all enums in Altinn.App.Core which do not have JsonConverterAttribute.
Assume that they use the default serialization policy (number)
Add JsonNumberEnumConverter

## Related Issue(s)
- #729

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Features/Telemetry/Telemetry.Data.cs
+++ b/src/Altinn.App.Core/Features/Telemetry/Telemetry.Data.cs
@@ -1,5 +1,6 @@
 using System.ComponentModel.DataAnnotations;
 using System.Diagnostics;
+using System.Text.Json.Serialization;
 using Altinn.Platform.Storage.Interface.Models;
 using NetEscapades.EnumGenerators;
 using static Altinn.App.Core.Features.Telemetry.Data;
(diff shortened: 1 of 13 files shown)


ivarne (approved):
This is an improvement over what we had, but there are a few imperfections.

* It fixes enums that we never serialise in http responses, making it seem like they must be serialised as number, when it is actually still an open question.
* It only looks at `Altinn.App.Core`, but we have a few enums in `Altinn.Platform.Storage.Interface`, and there is even a possibility that we expose some BCL enums.

An improved approach would integrate into swagger, to look at enums that we actually use in apis.

HauklandJ:
This PR has been replaced by [assembly scope solution pr](https://github.com/Altinn/app-lib-dotnet/pull/784)

Issue #717, opened 2024-07-17
Input validation before creating an eFormidlings "shipment"


SandGrainOne:
### Description

I would like to suggest a small improvement to the eFormidlings client and business logic during the steps right before creating a shipment. The logic would need to look at the instance and its data elements to determine if a shipment can be successfully generated.

The first and most critical validation step would be to check for more than one attachment with the same filename. The integration point for eFormidling gives off an error message and crashes the process if the app is attempting to upload a file with a name that match an earlier upload.

The application owner can add logic like this themselves, but this one example should be built in I think.

### Additional Information

I created this issue here in this repo even though I'm not sure exactly who it is that owns the eFormidlings client code. The code is hosted in the [altinn-studio repository](https://github.com/Altinn/altinn-studio/tree/main/src/Altinn.Common/Altinn.EFormidlingClient) for the time being. I do assume that this would require some changes in the app lib code as well.

ivarne:
I don't know much about eFormidling, but as far as I can tell this can be stopped at multiple stages.

* file analyzer/validator that (barring race conditions) prevents a user from uploading duplicate file names.
* normal task validation that prevents submission of attachments with duplicate file names. (User need to delete and reupload with unique names)
* logic in the preparation of the message that changes file names to be unique (eg add `-1` postfix when required)
* use the element id (guid) instead of the file name from the upload request.

* (also suggested) provide a way to make the file names on the instance unique when uploading. This is race condition prone.

martinothamar:
> logic in the preparation of the message that changes file names to be unique (eg add -1 postfix when required)

I think this is the suggestion we're considering. Will investigate

PR #774, opened 2024-09-12
Feat/717 Ensure unique filenames in eFormidling shipment

bjorntore:
Avoid duplicated filenames in eFormidling shipment.

## Related Issue(s)
- #717 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/EFormidling/Implementation/DefaultEFormidlingService.cs
+++ b/src/Altinn.App.Core/EFormidling/Implementation/DefaultEFormidlingService.cs
@@ -210,7 +210,7 @@ public class DefaultEFormidlingService : IEFormidlingService
             );
 
             string fileName = appLogic ? $"{dataElement.DataType}.xml" : dataElement.Filename;
-            using Stream stream = await _dataClient.GetBinaryData(
+            await using Stream stream = await _dataClient.GetBinaryData(
                 applicationMetadata.Org,
                 applicationMetadata.AppIdentifier.App,
                 instanceOwnerPartyId,
--- a/src/Altinn.App.Core/EFormidling/Implementation/DefaultEFormidlingService.cs
+++ b/src/Altinn.App.Core/EFormidling/Implementation/DefaultEFormidlingService.cs
@@ -260,7 +260,7 @@ public class DefaultEFormidlingService : IEFormidlingService
         }
     }
 
-    private static string GetUniqueFileName(
+    internal static string GetUniqueFileName(
         string? fileName,
         string dataTypeId,
         bool hasAppLogic,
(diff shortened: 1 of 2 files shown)


ivarne (approved):
Looks good, and great tests.

I added a few tests for the unique function, and had to make a smal change to ensure everything passes.
Screenshot 2024-09-12 at 21 54 48

After writing the test I realised that it might be good to prefix attachments with filename `arkivmelding` with their data type.

src/Altinn.App.Core/EFormidling/Implementation/DefaultEFormidlingService.cs
```diff
@@ -192,25 +192,47 @@ Instance instance
...
 
         Guid instanceGuid = Guid.Parse(instance.Id.Split("/")[1]);
         int instanceOwnerPartyId = int.Parse(instance.InstanceOwner.PartyId, CultureInfo.InvariantCulture);
+
+        // Keep track of already used file names to ensure they are unique. eFormidling does not allow duplicate filenames.
+        HashSet<string> usedFileNames = [eformidlingMetadataFilename];
+
+        List<string> dataTypeIds = applicationMetadata.DataTypes.Select(x => x.Id).ToList();
+
         foreach (DataElement dataElement in instance.Data)
```
martinothamar: Should we make this loop deterministic (sort or something)? AFAICT the `dataElement.Id` => `uniqueFilename` mapping can vary depending on the order of elements in `instance.Data`. `IServiceTask` may be retried in the future, so idempotency and determinism will be important properties to have.

Otherwise LGTM 👍

bjorntore: Simply order by id or something?
`foreach (DataElement dataElement in instance.Data.OrderBy(x => x.Id))`

ivarne: I think they get sorted by created time in storage.

I’m not sure when this order would matter, but it’s better to just sort.

ivarne:
> After writing the test I realised that it might be good to prefix attachments with filename `arkivmelding` with their data type.

I was thinking something like this to ensure that no files starts with `arkivmelding...` but gets renamed to `{dataType}-arkivmelding...`
```
        List<string> dataTypeIds = applicationMetadata
            .DataTypes.Select(x => x.Id)
            .Concat([eformidlingMetadataFilename.Substring(eformidlingMetadataFilename.IndexOf('.'))])
            .ToList();
```

src/Altinn.App.Core/EFormidling/Implementation/DefaultEFormidlingService.cs
martinothamar: I _think_ that should work, I assume the sorting algo in .NET for strings is stable, so that the order is the same everytime

martinothamar: Or is there a timestamp that could be used?

bjorntore: Created datetime?

martinothamar: Yeah that makes sense to me 👍

martinothamar (approved):

Issue #757, opened 2024-09-10
Tests that make sure user provided implementations of interfaces have expected `ServiceLifetime`


martinothamar:
### Description

If an app developer registers an `IInstanceProcessor` as `Transient`, we should always instantiate it transiently (not accidentally make it a singleton by rooting it in some other service). Ideally we would not have to, but this is the least surprising behavior as app devs typically aren't concerned with what they put in constructors (ref the `IHttpContextAccessor` leak incident).

We should develop tests that ensure that all (commonly implemented) public interfaces live for the expected duration. 

### Tasks

* Get a list of common user-implementable interfaces which we should develop lifetime tests for
* Implement tests

### Acceptance Criterias

* When `IInstantiationProcessor` services are registrered as `Transient`, they are instantiated with each invocation of `IProcessEventHandlerDelegator` (or downstream services/deps) methods
* The above criteria should be true for any other user implemented interface

martinothamar:
A list of interfaces


eFormidling:
* IEFormidlingMetadata
* IEFormidlingReceivers

validation:
* IDataElementValidator
* IFormDataValidator
* IInstanceValidator
* IInstantiationValidator
* ITaskValidator
* IFileValidator
* IFileAnalyser

process:
* IProcessExclusiveGateway
* IProcessTaskAbandon
* IProcessTaskEnd
* IProcessTaskStart
* IProcessTask
* IServiceTask
* IUserAction
* IUserActionAuthorizer

events:
* IEventHandler
* IEventSecretCodeProvider

data:
* IInstantiationProcessor
* IDataProcessor
* IExternalApiClient

payment:
* IOrderDetailsCalculator
* IPaymentProcessor

options (?):
* IInstanceAppOptionsProvider
* IAppOptionsProvider
* IAppOptionsFileHandler

datalist (?):
* IDataListProvider
* IInstanceDataListProvider

misc:
* IPageOrder
* IPdfFormatter
* IAppEvents
* IX509CertificateProvider

martinothamar:
2 more things I'd like to do:

* Analyzer for constructor use of `IHttpContextAccessor`: https://github.com/Altinn/app-lib-dotnet/pull/651
* Internal analyzer for usage of `AppImplementationFactory` in our constructors (https://github.com/Altinn/app-lib-dotnet/pull/1234)

PR #779, opened 2024-09-17
AppImplementationFactory to decouple app service lifetimes from library service lifetimes 

martinothamar:
## Description

Been a while, but describing the design and some review comments here

Src changes:
- `internal ImplementableByAppsAttribute` - marker attribute to tag interfaces that are meant to be implemented by apps
- `internal AppImplementationFactory` - common factory for services tagged by the above attribute
  - A recent change here - the `IServiceProvider` use here prefers the one through `IHttpContextAccessor` but will always fall back to the root/current scope provider. A case that necessitates this is `IHostedService`-based subscription through `IEventsSubscription` use. I still think we achieve what we want, that we always inject through request scope if there is one
- Analyzer for diagnostics when constructor injection or `IServiceProvider` injection is used on `ImplementableByAppsAttribute` services
- Generally just moving from constructor injection -> `AppImplementationFactory` injection
  -  Some exceptions for `ValidatorFactory`, so there is support for suppression through `/* altinn:injection:ignore */` comments. There might be a better design here, but in the interest of time this is what I landed on

For all public members changed I have searched through deployed apps and didn't discover any apps that we break.

Test changes:
- Lots of fixtures to
  - Reduce some boilerplate. There's been a ton of copy pasta
  - Do construction based on `IServiceProvider` since a lot of stuff needs that now to get access to `AppImplementationFactory`
  - Benefits: no deps on ambient state (xunit magic), generally less boilerplate
  - Drawbacks: missing services is now runtime/test exception and not build error
- `BuildStrictServiceProvider` extension method to default to container validation

## Tasks

eFormidling:

- [x] EFormidlingMetadata
- [x] EFormidlingReceivers

validation:

- [x] IDataElementValidator
- [x] IFormDataValidator
- [x] IInstanceValidator
- [x] IInstantiationValidator
- [x] ITaskValidator
- [x] IFileValidator
- [x] IFileAnalyser
- [x] IValidator

process:

- [x] IProcessExclusiveGateway
- [x] IProcessTaskAbandon
- [x] IProcessTaskEnd
- [x] IProcessTaskStart
- [x] IProcessTask
- [x] IServiceTask
- [x] IUserAction
- [x] IUserActionAuthorizer

events:

- [x] IEventHandler
- [x] IEventSecretCodeProvider

data:

- [x] IInstantiationProcessor
- [x] IDataProcessor
- [x] IDataWriteProcessor
- [x] IExternalApiClient

payment:

- [x] IOrderDetailsCalculator
- [x] IPaymentProcessor

options (?):

- [x] IInstanceAppOptionsProvider
- [x] IAppOptionsProvider
- [x] IAppOptionsFileHandler

datalist (?):

- [x] IDataListProvider
- [x] IInstanceDataListProvider

misc:

- [ ] IPageOrder
- [ ] IPdfFormatter
- [ ] IAppEvents
- [ ] IX509CertificateProvider

## Related Issue(s)
- #757

## Verification
- [x] All tests run green

## Documentation


martinothamar:
Nevermind, this doesn't actually work... Working on a new suggestion

ivarne:
I think the correct approach would be to have a shared factory for all extension point interfaces, that are registered as a `Scoped` service and retrieves services from a `IServiceProvider` (that inherits the scope). That way we can also add telemetry span around the `GetRequiredService` call to catch potentially slow constructors written by users.

src/Altinn.App.Core/AppImplementationFactory.cs
```diff
@@ -0,0 +1,57 @@
...
+        ?? throw new InvalidOperationException("Couldn't resolve DI container from current context");
+
+    public T GetRequired<T>()
+        where T : class => _sp.GetRequiredService<T>();
+
+    public T? Get<T>()
+        where T : class => _sp.GetService<T>();
+
+    public IEnumerable<T> GetAll<T>()
+        where T : class => _sp.GetServices<T>();
```
ivarne: What is the point of this wrapper around the ServiceProvider giving the methods new names? 

It would make more sense if it contained a function per interface or something like that.
```C#
    public IEnumerable<IDataProcessor> GetDataProcessors() => 
        _sp.GetServices<IDataProcessor>()
```

src/Altinn.App.Core/AppImplementationFactory.cs
```diff
@@ -0,0 +1,57 @@
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Altinn.App.Core.Features;
+
+/// <summary>
+/// Marker attribute for interfaces that are meant to be implemented by apps.
+/// </summary>
+[AttributeUsage(AttributeTargets.Interface, AllowMultiple = false)]
+internal sealed class ImplementableByAppsAttribute : Attribute { }
```
ivarne, line 10: What benefits will marking interfaces in this way give? Can it make them show up as suggestions in app developers editors?

src/Altinn.App.Core/AppImplementationFactory.cs
```diff
@@ -0,0 +1,57 @@
...
+        services.AddSingleton<AppImplementationFactory>();
+        return services;
+    }
+}
+
+internal sealed class AppImplementationFactory
+{
+    internal sealed record DIScopeHolder(IServiceProvider? ScopedServiceProvider);
+
+    private readonly DIScopeHolder _diScopeHolder;
```
ivarne: I find this concept and the additional indirection a little hard to understand.

The straight forward solution would be to just inject a scoped ServiceProvider and use reflection to access the `IsRootScope` to verify when running in debug mode (as we do in tests, but not when apps are running)

```C#
#if DEBUG
        var serviceProviderType = serviceProvider.GetType();
        if (
            serviceProviderType.FullName
            != "Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope"
        )
        {
            throw new InvalidOperationException(
                "InterfaceFactory expects ServiceProviderEngineScope to run debug scope validation"
            );
        }
        var rootProperty = serviceProviderType.GetProperty(
            "IsRootScope",
            System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Instance
        );
        if (rootProperty == null)
        {
            throw new InvalidOperationException(
                "InterfaceFactory expects ServiceProviderEngineScope to have IsRootScope property"
            );
        }
        var rootPropertyValue = rootProperty.GetValue(serviceProvider);
        switch (rootPropertyValue)
        {
            case true:
                throw new InvalidOperationException("InterfaceFactory should not be used in root scope");
            case false:
                break;
            default:
                throw new InvalidOperationException("InterfaceFactory expects IsRootScope to be a boolean");
        }
#endif
```

src/Altinn.App.Core/AppImplementationFactory.cs
martinothamar: In this PR just as a mechanism to categorize. I could make a list of interfaces, but then it is hard to know that the list should be updated when an interface is added. There is future potential as well. Main point is having some mechanism to categorize the interfaces (since it is public). Could also be the inverse - public interfaces that are _not_ meant to be public are attributed with `Pubternal` or something

src/Altinn.App.Core/AppImplementationFactory.cs
martinothamar: No specific intention behind the names, I went with short and simple, open to alternatives. 

Specific overalods per "type" feels like a lot of code for little reason to me, what is the added benefit specifically?

src/Altinn.App.Core/AppImplementationFactory.cs
martinothamar: Hmmm so my only intention with this thing was to enable HTTP-less test fixtures, and that's it. Might have given it a poor name 😄 So ideally what we go with only changes behavior for those specific cases (simple tests using only a DI container). Since for all other cases (app runtime, integration tests) we want the `IHttpContextAccessor` request scope, I'm not sure if the code you posted is justified. To me it seems complicating rather than simplifying

src/Altinn.App.Core/AppImplementationFactory.cs
ivarne: I think it makes more sense to use the same names as IServiceProivder if we expose the same functionality.

1. This list of methods will work as a list of interfaces that we intend users to implement to override functionality.
2. More obvious to use the special method, instead of just "use this different service provider".
3. It documents and enforces that an interface should be retrieved as `IEnumerable<T>` instead of just fetching the first `T, so that we do that consistently.

src/Altinn.App.Core/AppImplementationFactory.cs
ivarne: I think having a constructor `(IServiceProivder sp)` is simpler (especially for tests) than having to wrap it in an `internal DIScopeHolder`. The code I posted is just part of an internal verification, and not the how the rest of the code base uses the interface factory.

src/Altinn.App.Internal.Analyzers/Altinn.App.Internal.Analyzers.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <Using Include="Microsoft.CodeAnalysis.CSharp.Syntax" />
+
+    <AdditionalFiles Include="AnalyzerReleases.Shipped.md" />
+    <AdditionalFiles Include="AnalyzerReleases.Unshipped.md" />
+  </ItemGroup>
+
+  <!-- 
+    NOTE:
+    InternalsVisibleTo messes with PolySharp, which we use to support newer C# features in analyzer code (such as records)
+    So it is disabled currently and relevant members sin analyzer code is public instead
```
HauklandJ: members sin analyzer -> miks av norsk og engelsk?

src/Altinn.App.Internal.Analyzers/AppImplementationInjectionAnalyzer.cs
```diff
@@ -0,0 +1,210 @@
...
+        "ALTINNINT9999",
+        Category.General,
+        "Unknown analyzer error",
+        "Unknown error occurred during analysis: '{0}'"
+    );
+
+    public static readonly DiagnosticDescriptor DangerousConstructorInjection = Error(
+        "ALTINNINT0001",
+        Category.General,
+        "Dangerious constructor injection",
```
HauklandJ: Dangerious typo

src/Altinn.App.Internal.Analyzers/AppImplementationInjectionAnalyzer.cs
```diff
@@ -0,0 +1,210 @@
...
+        "Service interface '{0}' is injected into the constructor of '{1}'."
+            + " This is dangerous as we couple the lifetime of '{1}' to the lifetime of the implementation of '{0}'."
+            + " Use the 'AppImplementationFactory' to instantiate app implementations lazily instead."
+    );
+
+    public static readonly DiagnosticDescriptor DangerousServiceProviderServiceResolution = Error(
+        "ALTINNINT0002",
+        Category.General,
+        "Dangerious 'IServiceProvider' service resolution",
```
HauklandJ: typo copy pasta

src/Altinn.App.Core/AppImplementationFactory.cs
```diff
@@ -0,0 +1,57 @@
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Altinn.App.Core.Features;
+
+/// <summary>
+/// Marker attribute for interfaces that are meant to be implemented by apps.
+/// </summary>
+[AttributeUsage(AttributeTargets.Interface, AllowMultiple = false)]
```
HauklandJ, line 9: Makes sense to me that we mark the ones we want to expose to apps rather than the reverse

src/Altinn.App.Core/AppImplementationFactory.cs
HauklandJ: I like the idea of going `factory.GetServiceIWant` rather than` factory.Get<ServiceIWant`> because of the doc and autocomplete it gives.

HauklandJ (commented):
I like the approach, some comments

danielskovli:
@martinothamar I don't a useful opinion or contribution to this discussion, unfortunately. It would be great to get a summary of the solution at some point though, after the details have been worked out.

src/Altinn.App.Internal.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,10 @@
+; Unshipped analyzer release
+; https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
+
+### New Rules
+
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+ALTINNINT0001 | General | Error | Dangerous constructor injection
+ALTINNINT0002 | General | Error | Dangerious 'IServiceProvider' service resolution
```
bjorntore: Dangerious typo

src/Altinn.App.Core/AppImplementationFactory.cs
martinothamar: > This list of methods will work as a list of interfaces that we intend users to implement to override functionality.

We can get the same effect from searching for attribute use though
image

> More obvious to use the special method, instead of just "use this different service provider".

In this case there are build errors if one tries to inject `IEFormidlingMetadata` using constructors or `IServiceProvider` directly though (and so you _have_ to use the the different service provider/factory regardless of how we implement it). In any case, if we are developing in the context of some feature, these methods living somewhere else probably won't matter in terms of "obviousness"? In that scenario you are thinking about how to implement the feature, right?

> It documents and enforces that an interface should be retrieved as IEnumerable<T> instead of just fetching the first `T, so that we do that consistently.

Isn't that obvious depending on the `T`? Presumably the person developing the code requiring the `T` also knows how many `T`s there are/could be?

> I like the idea of going factory.GetServiceIWant rather than factory.Get<ServiceIWant> because of the doc and autocomplete it gives.

Again, at the time one is reaching for `ServiceIWant`, you already know what you need right? I'm struggling a bit with these scenarios that are raised 😄

src/Altinn.App.Core/AppImplementationFactory.cs
martinothamar: The tests don't need to care about that though. There are the 2 DI methods for registering the factory as a service, so in that sense I would say this code is internal to the factory. So the two "modes" of execution for the current code is

1. `AddAppImplementationFactory` -> `IHttpContextAccessor.HttpContext.RequestServices` (for real/app and integration test use)
2. `AddTestAppImplementationFactory` -> Root container (for simple tests where we don't want to spin up ASP.NET Core test host and whatnot)

I added a commit where I get rid of the so called "scope holder" and just use a bool in the constructor instead. So in the first mode we will get an exception if there is no request scope, and in the test mode I don't think we really care

src/Altinn.App.Core/AppImplementationFactory.cs
bjorntore: I'm fine with both ways, with a slight preference for what @martinothamar already did, but with the same method names as IServiceProvider.

bjorntore:
Really like the attribute plus DiagnosticAnalyzer approach. No objections.

bjorntore (approved):

martinothamar:
Finally got some time to work through the rest and have tested this in an app. Since this is all internal stuff I'll just merge this now so I can spend some time rebasing other branches from main, in case I cause a lot of conflicts. Feel free to still review and/or create PRs for stuff if you want changes. I've added `backport-ignore` since this is a pretty broad change. If there is something I haven't thought of I'd rather find out during prerelease 😄

PR #784, opened 2024-09-23
Use ApiConventions to solve enum serialization on assembly scope

HauklandJ:
## Description
This is a new solution, compare to https://github.com/Altinn/app-lib-dotnet/pull/773. 

Here we annotate controllers automatically using apiconventions. This means we do not have to update all enums with an annotation and remember it for future enums, and do the same for enums from storage/platform.

The tradeoff is the scaffolding to make this happen.

## Related Issue(s)
- #{issue number}

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Api/Controllers/Conventions/EnumAsNumberFormatter.cs
```diff
@@ -0,0 +1,59 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Formatters;
+
+namespace Altinn.App.Api.Controllers.Conventions;
+
+internal class EnumAsNumberFormatter : SystemTextJsonOutputFormatter
```
martinothamar: Do we need to implement `CanWriteResult` here? I think this formatter should only do its thing if `GetJsonSettingsName` matches the settings name?

src/Altinn.App.Api/Controllers/Conventions/EnumAsNumberFormatter.cs
```diff
@@ -0,0 +1,71 @@
...
+    public override JsonConverter CreateConverter(Type typeToConvert, JsonSerializerOptions options)
+    {
+        var instance =
+            Activator.CreateInstance(typeof(EnumToNumberJsonConverter<>).MakeGenericType(typeToConvert))
+            ?? throw new InvalidOperationException($"Failed to create converter for type {typeToConvert.FullName}");
+        return (JsonConverter)instance;
+    }
+}
+
+internal class EnumToNumberJsonConverter<TEnum> : JsonConverter<TEnum>
```
martinothamar: Do we need a custom converter? Isn't number encoding the default?

src/Altinn.App.Api/Controllers/PartiesController.cs
```diff
@@ -21,6 +22,7 @@ namespace Altinn.App.Api.Controllers;
 /// </summary>
 [Authorize]
 [ApiController]
+[JsonSettingsName("EnumAsNumber")]
```
martinothamar: Maybe the setting name should just be `AltinnApis` or something? Denoting that whatever settings are injected are what is required to make our own APIs stable?

Now maybe we should consider API conventions as well so that we don't have to remember this attribute everywhere?

src/Altinn.App.Api/Controllers/Conventions/EnumAsNumberFormatter.cs
ivarne: If I understand correctly, the issue is that service owners might change the default, affecting our apis.

ivarne: Oh! you are thinking about using [`JsonNumberEnumConverter<TEnum>`](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonnumberenumconverter-1?view=net-8.0) instead?

HauklandJ: Sure. Its generic clause is struct instead of struct, Enum.

src/Altinn.App.Api/Controllers/Attributes/JsonSettingsNameAttribute.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        Name = name;
+    }
+
+    internal string Name { get; }
+}
+
+internal static class JsonSettingNames
+{
+    internal const string ALTINN_API = "AltinnApi";
```
martinothamar: I thought the convention for consts was just PascalCase now, do you get any IDE hints/warnings?

src/Altinn.App.Api/Controllers/Conventions/ConfigureMvcJsonOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+                .OutputFormatters.Select((formatter, index) => (formatter, index))
+                .Where(f => f.formatter is SystemTextJsonOutputFormatter)
+                .Select(f => f.index + 1)
+                .FirstOrDefault() - 1;
+        if (indexOfDefaultJsonFormatter < 0)
+            throw new InvalidOperationException("Could not find the default JSON output formatter");
+
+        options.OutputFormatters.Insert(
+            indexOfDefaultJsonFormatter,
+            AltinnApiJsonFormatter.CreateFormatter(_jsonSettingsName, jsonOptions)
```
martinothamar, line 42: Should we create the `JsonSerializerOptions` from the default/builtin JSON formatter instead?

martinothamar (commented):
Could we have a test that does the inverse as well? Like have some random controller in the test project that doesn't have the same attribute?

src/Altinn.App.Api/Controllers/Attributes/JsonSettingsNameAttribute.cs
HauklandJ: VsCode with C#, C# devkit, and C# Utilities extensions seems to think this is OK. But I see that Microsoft disagrees: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions..

martinothamar: I thought there was a rule touching on this in .editorconfig, maybe the EditorConfig ext is needed as well

HauklandJ:
> Could we have a test that does the inverse as well? Like have some random controller in the test project that doesn't have the same attribute?

Is this still relevant after the additional tests?

src/Altinn.App.Api/Controllers/Conventions/ConfigureMvcJsonOptions.cs
```diff
@@ -10,39 +13,42 @@ namespace Altinn.App.Api.Controllers.Conventions;
...
+        var indexOfDefaultJsonFormatter = options.OutputFormatters.IndexOf(defaultJsonFormatter);
 
-        options.OutputFormatters.Insert(
-            indexOfDefaultJsonFormatter,
-            AltinnApiJsonFormatter.CreateFormatter(_jsonSettingsName, jsonOptions)
-        );
+        var defaultSerializerOptions = defaultJsonFormatter.SerializerOptions;
+        var serializerOptions = new JsonSerializerOptions(defaultSerializerOptions)
+        {
+            Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping
```
martinothamar: I thought the intention was to remove this? So that whatever is in `serializerOptions` always just reflects the defaults of ASP.NET Core

HauklandJ: good spot!

src/Altinn.App.Api/Controllers/Conventions/AltinnApiJsonFormatter.cs
```diff
@@ -0,0 +1,41 @@
...
+    internal static AltinnApiJsonFormatter CreateFormatter(string settingsName, JsonSerializerOptions serializerOptions)
+    {
+        if (serializerOptions.Encoder is null)
+        {
+            // If the user hasn't explicitly configured the encoder, use the less strict encoder that does not encode all non-ASCII characters.
+            serializerOptions = new JsonSerializerOptions(serializerOptions)
+            {
+                Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
+            };
+        }
```
martinothamar, line 40: So now we shouldn't need this part right? Since this part is coming from the options of the initialized formatter?

HauklandJ: You are correct :face_with_peeking_eye:

test/Altinn.App.Api.Tests/Controllers/Conventions/AltinnApiJsonFormatterTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            (stream, encoding) => new StreamWriter(stream, encoding),
+            typeof(object),
+            new object()
+        );
+
+        // Act
+        bool canWrite = formatter.CanWriteResult(context);
+
+        // Assert
+        Assert.True(canWrite);
```
martinothamar, line 66: Would be nice to have a test for the inverse of this? An endpoint without `JsonSettingsNameAttribute` has `CanWriteResult` return false

martinothamar (approved):

PR #792, opened 2024-09-24
Add an optional footer for the generated PDF

### LINKED ISSUE #787: Option to add a watermark on the generated PDF receipts
### Description

There is a need to add a watermark stamp on the generated PDF receipts to ensure that the PDF is actually generated after submitting the form. "Clever" users have used the build in print preview to fake a form being submitted to avoid payment steps later in the process. The case manager in the municipalities use the watermark as part of their routine and we should continue to support this feature from Altinn 2.

The watermark itself is basically a text value including date/time for when the PDF was generated as well as a reference id provided by the service owner.

The solution should be dynamic in terms of what values goes into the watermark allowing the application developers to construct the value based on date/time, values from the data model and/or instance metadata.

### Additional Information
This issue should be considered i relation to https://github.com/Altinn/app-frontend-react/issues/1805, and maybe allow for the inclusion of a "not valid" watermark if the PDF is generated before the final data submission.

**Example image showing watermarks in red**
image

### Design
Component : https://www.figma.com/design/aECTVzeAa3sJDucjYWQuUQ/Oppsummering-og-PDF?node-id=966-190399&t=1JdATXK1BkP2v3qF-4

Example: https://www.figma.com/design/aECTVzeAa3sJDucjYWQuUQ/Oppsummering-og-PDF?node-id=1081-140542&t=1JdATXK1BkP2v3qF-4

### Tasks

_No response_

### Acceptance Criterias
- [x] Given that you enable watermark on the PDF, when you **don't** implement a custom watermark prover then we should return date, time and Altinn archive reference. 
- [ ] Given that you need a custom watermark text, when you implement a custom watermark provider then you should get the custom text returned from the watermark provider.
- [x] Given that you are on any page, when you show the PDF then the watermark should be printed in the footer area.
- [x] Given that you enable watermark, when you see show the PDF, then the watermark should stand out (ref. the design)
- [ ] Given that the PDF is generated before the data is complete, when you preview the PDF, then the watermark should show a language specific text for example in Norwegian "Forhåndsvisning" by default. 
- [ ] Given that the service owner would like their own text instead of the default, it should be an option to override the text with a specified text resource.

Magnusrm:
Adds config for showing a footer at the bottom of the generated PDF pages.

image

## Description
Adds a new property to `"PdfGeneratorSettings"` in `appsettings.json`: `"DisplayFooter"`.
The property is a boolean that, when true, adds a footer to each page of your generated PDF. The footer contains information about the title of the document, the date and time the pdf was generated, the altinn-reference-id, and the current and total page number.

## Related Issue(s)
- #787 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Infrastructure/Clients/Pdf/PdfGeneratorClient.cs
+++ b/src/Altinn.App.Core/Infrastructure/Clients/Pdf/PdfGeneratorClient.cs
@@ -4,6 +4,7 @@ using Altinn.App.Core.Configuration;
 using Altinn.App.Core.Internal.Auth;
 using Altinn.App.Core.Internal.Pdf;
 using Altinn.App.Core.Models.Pdf;
+using Altinn.Common.EFormidlingClient.Models;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Options;
 
@@ -52,12 +53,27 @@ public class PdfGeneratorClient : IPdfGeneratorClient
     /// <inheritdoc/>
     public async Task<Stream> GeneratePdf(Uri uri, CancellationToken ct)
     {
+        return await GeneratePdf(uri, null, ct);
+    }
+
+    /// <inheritdoc/>
+    public async Task<Stream> GeneratePdf(Uri uri, string? FooterContent, CancellationToken ct)
+    {
         bool hasWaitForSelector = !string.IsNullOrWhiteSpace(_pdfGeneratorSettings.WaitForSelector);
+        bool footer = _pdfGeneratorSettings.Footer;
         PdfGeneratorRequest generatorRequest =
             new()
             {
                 Url = uri.AbsoluteUri,
-                WaitFor = hasWaitForSelector ? _pdfGeneratorSettings.WaitForSelector : _pdfGeneratorSettings.WaitForTime
+                WaitFor = hasWaitForSelector
+                    ? _pdfGeneratorSettings.WaitForSelector
+                    : _pdfGeneratorSettings.WaitForTime,
+                Options =
+                {
+                    HeaderTemplate = "<span/>",
+                    FooterTemplate = FooterContent ?? "<span/>",
+                    DisplayHeaderFooter = footer,
+                },
             };
 
         generatorRequest.Cookies.Add(
--- a/src/Altinn.App.Core/Infrastructure/Clients/Pdf/PdfGeneratorClient.cs
+++ b/src/Altinn.App.Core/Infrastructure/Clients/Pdf/PdfGeneratorClient.cs
@@ -60,7 +60,6 @@ public class PdfGeneratorClient : IPdfGeneratorClient
     public async Task<Stream> GeneratePdf(Uri uri, string? FooterContent, CancellationToken ct)
     {
         bool hasWaitForSelector = !string.IsNullOrWhiteSpace(_pdfGeneratorSettings.WaitForSelector);
-        bool displayFooter = _pdfGeneratorSettings.Footer;
         PdfGeneratorRequest generatorRequest =
             new()
             {
@@ -72,7 +71,7 @@ public class PdfGeneratorClient : IPdfGeneratorClient
                 {
                     HeaderTemplate = "<div/>",
                     FooterTemplate = FooterContent ?? "<div/>",
-                    DisplayHeaderFooter = displayFooter,
+                    DisplayHeaderFooter = FooterContent != null,
                 },
             };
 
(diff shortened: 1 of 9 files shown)


src/Altinn.App.Core/Infrastructure/Clients/Pdf/PdfGeneratorClient.cs
```diff
@@ -51,13 +51,27 @@ IHttpContextAccessor httpContextAccessor
 
     /// <inheritdoc/>
     public async Task<Stream> GeneratePdf(Uri uri, CancellationToken ct)
+    {
+        return await GeneratePdf(uri, null, ct);
+    }
+
+    /// <inheritdoc/>
+    public async Task<Stream> GeneratePdf(Uri uri, string? FooterContent, CancellationToken ct)
```
martinothamar: `FooterContent` should be camelCase

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
```diff
@@ -225,4 +229,34 @@ private static string GetValidFileName(string fileName)
         fileName = Uri.EscapeDataString(fileName.AsFileName(false));
         return fileName;
     }
+
+    private static string GetFooterContent(Instance instance)
+    {
+        TimeZoneInfo norwegianTimeZone = TimeZoneInfo.FindSystemTimeZoneById("Europe/Oslo");
```
martinothamar: Have you tested this in a container also? I hope base images that apps use have the tz db/`tzdata` installed, that's not always the case and depends on the base image used. We use alpine and in some versions `tzdata` had to be manually installed. So it would be nice to get this tested with a fully built app

martinothamar (approved):
A couple of comments, code LGTM 👍

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
martinothamar: I managed to test this, so it looks fine
image
image

martinothamar: image

:(

martinothamar: Looks like we would need `RUN apk add --no-cache icu-libs tzdata` to make this work

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
```diff
@@ -225,4 +229,34 @@ private static string GetValidFileName(string fileName)
...
+        DateTimeOffset norwegianNow = TimeZoneInfo.ConvertTime(DateTimeOffset.UtcNow, norwegianTimeZone);
+
+        string dateGenerated = norwegianNow.ToString("G", new CultureInfo("nb-NO"));
+
+        string altinnReferenceId = instance.Id.Split("/")[1].Split("-")[4];
+
+        string footerTemplate =
+            $@"<div style='font-family: Inter; font-size: 12px; width: 100%; display: flex; flex-direction: column; gap: 12px; padding: 0 70px 0 70px;'>
+                <span class='title'></span>
```
bjosttveit: Since you are adding the page title in the PDF footer, maybe we should have unit/cypress tests in app-frontend verifying that this always shows the `appName - serviceOwner` in PDF mode? This could be changed when we start working on navigation 🤔

bjosttveit (approved):

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
Magnusrm: I dont think the title here is related to pages. It is just the document title, which seems to get created based on the filename. The filename gets created using the `"appName"` and `"serviceOwner"` text resources. If the text resources does not exist, the title will just be the "raw" appname (the repo name i think?).

bjosttveit: I am pretty sure it is the page title, as in the `<title>` element. Chrome knows nothing about file-names as it simply returns the PDF as a base64 string, and then we determine the filename afterwards

Magnusrm: You are correct, it seems to be the `title` tag in the apps `index.cshtml` file:
image

Edit: Seems to not be this one after all, changing it does nothing 😅

bjosttveit: That is because we control it in app-frontend 😄 (`WindowTitleProvider.tsx`) which is why I think we should add tests for it there. Because if I remember correctly, one wish was that the title reflects the current page you are on in the form, and we could inadvertantly break the PDF title if we don't handle that case specifically

martinothamar (approved):

PR #807, opened 2024-09-27
Make backend expression "equals" support float and long data types

ivarne:
@tina-ahm reported that `RemoveHiddenData` fails for expressions like `["equals",["dataModel", "field"],3]` if `field` is of type `long` in the C# class.

This patch adds support for `long` and `float`, and adds a message to the exception to make the issue easier to debug if more types are needed for `"equals"`.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
+++ b/src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
@@ -485,39 +485,19 @@ public static class ExpressionEvaluator
             return bvalue ? "true" : "false";
         }
 
-        if (value is string svalue)
+        return value switch
         {
             // Special case for "TruE" to be equal to true
-            if ("true".Equals(svalue, StringComparison.OrdinalIgnoreCase))
-            {
-                return "true";
-            }
-            else if ("false".Equals(svalue, StringComparison.OrdinalIgnoreCase))
-            {
-                return "false";
-            }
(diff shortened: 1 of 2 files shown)


martinothamar (approved):
nice 👍  are `byte`/`sbyte` and unsigned integer variants also usable in datamodels?

src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
```diff
@@ -485,39 +485,19 @@ bool ab
...
-            else if ("false".Equals(svalue, StringComparison.OrdinalIgnoreCase))
-            {
-                return "false";
-            }
-            else if ("null".Equals(svalue, StringComparison.OrdinalIgnoreCase))
-            {
-                return null;
-            }
-
-            return svalue;
```
ivarne, line 504: I used Riders refactoring to turn this into a switch expression, and it lost this default case for string. Good I had tests for that.

ivarne:
> nice 👍 are `byte`/`sbyte` and unsigned integer variants also usable in datamodels?

I don't really know, but I see no reason why they should not work here. Only question would be how to translate them to string for comparison, but given that `System.Text.Json` has answered that question, I think it would be safe to follow their example (we do trust JsonSerializer to do a good job before sending anything to frontend anyway).

Issue #776, opened 2024-09-16
New IDataProcessing interface that supports multiple DataElements and adding elements in response to changes


ivarne:
### Description

With support for subforms and multiple data models and the multiple PATCH endpoint, app developers want to read and write multiple data elements in a single request, and also to be able to ADD subform instances based on information changes in the main data element.

Deleting is different, because the current PATCH interface does not allow to signal deletion, and in addition it is likely better to have some sort of hidden mechanism so that the data isn't actually deleted until we trigger `/process/next`

### Additional Information

Frontend will not yet support receiving new data elements in a PATCH response, and this needs to be added. If we want to support deletion of whole data elements, an appropriate interface needs to be negotiated. Likely solution would be to send `instance.Data` (or the full instance object)

kentare:
Ref https://altinn.slack.com/archives/C02EJ9HKQA3/p1726137517047409 

Gjentar behov vi har fra denne tråden:
Det er viktig at vi kan prefille med forskjellige verdier i datamodellen til subformen, vi gjør dette i dag i InstantiationProcessor. Vi gjør dette i dag med hjelp av gateways+InstantiationProcessor.

Casen er slik: Når instansen opprettes opprettes det et subform for Selskapsregnskap, denne prefilles med readonly data som er relevant til selskapsregnskapet ( i dag gjøres dette i InstantiationProcessor).  Om en bruker svarer "JA"  på en radiobutton (Skal konsernregnskap legges ved?) i skjemaet skal en subform (konsernregnskap) legges til med data som er readonly  som omhandler konsernet. Dette er data som brukeren ikke kan legge inn selv eller endre på

ivarne:
Fixed and available in 8.5.0-preview.4
- #730 

The new interface is named `IDataWriteProcessor`

PR #809, opened 2024-09-30
Add draft support for adding/removing/modifying other data elements in dataProcessWrite

ivarne:
The main goal is to add a method to `IDataProcessor` that takes an `IInstanceDataAccessor` and a list of changes and to add methods to `IInstanceDataAccessor` to add and remove data elements.

## Major refactorings
- New service `ModelSerializationService` to handle (de)serialisation of json and xml with convenience methods `*Storage` that looks in applicationmetadata 
- Extend `IInstanceDataAccessor` with methods to add and remove data elements from the instance
    - These are currently lazy, which means that they return void, and updating storage is deferred till after data processing is complete
- Extend `CachedInstanceDataAccessor` with functionality to ensure that nothing has been changed, after a certain point. This is used so that after data processing we can save to storage in parallel with validation without having to worry about validators mutating the models undetected.
- `DataElementId` was previously intended to be a thin wrapper around `System.Guid`, to communicate that this isn't just any guid, but a Guid from a specific database column. Every time I created one it was from a `DataElement`, and it was often required to get the data type, so I added fields to avoid lookups. Now it is more of an Immutable DataElement. Not sure everybody loves that development.
- `DataClient` uses the new `ModelSerializationService` and `IDataClient` gets a new `GetDataBytes` to argument the old `GetBinaryData` for those who want a clean `byte[]` that isn't wrapped in a `Stream`.

## Open questions
- New `IDataProcessing` interface, or just extend the existing with a new method. What to call the new one in that case?
- I got some help from ChatGPT to write `MemoryAsStream` in order to be able to pass `ReadOnlyMemory<byte>` to functions that expects a stream. I also found an alternative in [Windows Developer Toolkit](https://learn.microsoft.com/en-us/dotnet/api/microsoft.toolkit.highperformance.extensions.readonlymemoryextensions.asstream?view=win-comm-toolkit-dotnet-6.1). I'm not sure it is worth including a full toolkit for 14 effective lines of code.

## Remaining updates to make old code conform to new patterns
- Deprecate `ModelDeserializer` and use `ModelSerializationService` everywhere.
- Use `IInstanceDataAccessor` instead of `IDataClient` in more places where it makes sense to use the new data-processing interface
    - Instance creation
    - Data POST
    - Data PUT
    - Simpler detection for action implementations
    - Run new data-processing method on file/attachment upload
    

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/776

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
+++ b/src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
@@ -10,7 +10,7 @@ namespace Altinn.App.Core.Helpers.Serialization;
 /// Represents logic to deserialize a stream of data to an instance of the given type
 /// </summary>
 // [Obsolete(
-//     "This class is deprecated and will be removed in a v9. Use Altinn.App.PlatformServices.Helpers.Serialization.ModelSerializationSerivce from dependency injection instead."
+//     "This class is deprecated and will be removed in a v9. Use Altinn.App.Core.Helpers.Serialization.ModelSerializationSerivce from dependency injection instead."
 // )]
 public class ModelDeserializer
 {
--- a/src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
+++ b/src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
@@ -9,6 +9,9 @@ namespace Altinn.App.Core.Helpers.Serialization;
 /// <summary>
 /// Represents logic to deserialize a stream of data to an instance of the given type
 /// </summary>
+// [Obsolete(
+//     "This class is deprecated and will be removed in a v9. Use Altinn.App.PlatformServices.Helpers.Serialization.ModelSerializationSerivce from dependency injection instead."
+// )]
 public class ModelDeserializer
 {
     private static readonly JsonSerializerOptions _jsonSerializerOptions = new(JsonSerializerDefaults.Web);
--- a/src/Altinn.App.Core/Helpers/Serialization/ModelSerializationService.cs
+++ b/src/Altinn.App.Core/Helpers/Serialization/ModelSerializationService.cs
@@ -88,12 +88,7 @@ public class ModelSerializationService
 
         XmlSerializer serializer = _xmlSerializer.GetSerializer(modelType);
         serializer.Serialize(xmlWriter, model);
-        if (!memoryStream.TryGetBuffer(out var segment))
-        {
-            throw new InvalidOperationException("Failed to get buffer from memory stream");
-        }
-
-        return segment.AsMemory().RemoveBom();
+        return memoryStream.ToArray().AsMemory().RemoveBom();
     }
 
     /// <summary>
@@ -161,11 +156,8 @@ public class ModelSerializationService
     }
 
     /// <summary>
-    ///
+    /// Deserialize utf8 encoded xml data to a model of the specified type
     /// </summary>
-    /// <param name="data"></param>
-    /// <param name="modelType"></param>
-    /// <returns></returns>
     public object DeserializeXml(ReadOnlySpan<byte> data, Type modelType)
     {
         using var activity = _telemetry?.StartDeserializeFromXmlActivity(modelType);
--- a/src/Altinn.App.Core/Helpers/Serialization/ModelSerializationService.cs
+++ b/src/Altinn.App.Core/Helpers/Serialization/ModelSerializationService.cs
@@ -1,0 +1,267 @@
+using System.Collections.Concurrent;
+using System.Text;
+using System.Text.Json;
+using System.Xml;
+using System.Xml.Serialization;
+using Altinn.App.Core.Features;
+using Altinn.App.Core.Internal.AppModel;
+using Altinn.Platform.Storage.Interface.Models;
+
+namespace Altinn.App.Core.Helpers.Serialization;
+
+/// <summary>
+/// DI registered service for centralizing (de)serialization logic for data models
+/// </summary>
+public class ModelSerializationService
+{
+    private static readonly JsonSerializerOptions _jsonSerializerOptions = new(JsonSerializerDefaults.Web);
+    private static readonly XmlSerializerCache _xmlSerializer = new();
+
+    private readonly Telemetry? _telemetry;
+    private readonly IAppModel _appModel;
+
+    /// <summary>
+    /// Constructor for the ModelDeserializerService
+    /// </summary>
+    public ModelSerializationService(IAppModel appModel, Telemetry? telemetry = null)
+    {
+        _appModel = appModel;
+        _telemetry = telemetry;
+    }
+
+    /// <summary>
+    /// Deserialize binary data from storage to a model of the classRef specified in the dataType
+    /// </summary>
+    /// <param name="data">The binary data</param>
+    /// <param name="dataType">The data type used to get content type and the classRef for the object to be returned</param>
+    /// <returns>The model specified in </returns>
+    public object DeserializeFromStorage(ReadOnlySpan<byte> data, DataType dataType)
+    {
+        var type = GetModelTypeForDataType(dataType);
+
+        // TODO: support sending json to storage based on dataType.ContentTypes
+        return DeserializeXml(data, type);
+    }
+
+    /// <summary>
+    /// Serialize an object to binary data for storage, respecting classRef and content type in dataType
+    /// </summary>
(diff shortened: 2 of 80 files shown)


src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
```diff
@@ -9,6 +9,9 @@ namespace Altinn.App.Core.Helpers.Serialization;
 /// <summary>
 /// Represents logic to deserialize a stream of data to an instance of the given type
 /// </summary>
+// [Obsolete(
+//     "This class is deprecated and will be removed in a v9. Use Altinn.App.PlatformServices.Helpers.Serialization.ModelSerializationSerivce from dependency injection instead."
```
martinothamar: Namespace in this comment seems to be slightly off. Users of this class (I can only see brg/starteogdrive), which overload would they switch to? `DeserializeAsync` -> ?

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
```diff
@@ -70,33 +77,42 @@ public async Task<DataElement> InsertFormData<T>(
         int instanceOwnerPartyId,
         string dataType
     )
+        where T : notnull
```
martinothamar, line 80: Need to notify these guys? https://altinn.studio/repos/brg/starteogdrive/src/commit/e3b62d445a112e2db65004b90c1f0575ee3c52ab/App/Services/BRGProfileService.cs#L618-L631

src/Altinn.App.Core/Internal/Data/CachedInstanceDataAccessor.cs
```diff
@@ -40,120 +49,314 @@ IAppModel appModel
...
         }
-        if (dataTypeObj.MaxCount != 1)
+
+        return dataType;
+    }
+
+    /// <inheritdoc />
+    public void AddFormDataElement(string dataTypeString, object data)
+    {
+        var dataType = GetDataTypeByString(dataTypeString).Result;
```
martinothamar: Any way to avoid `.Result`? 😅

src/Altinn.App.Core/Internal/Data/CachedInstanceDataAccessor.cs
```diff
@@ -40,120 +49,314 @@ IAppModel appModel
...
+        var idAsString = dataElementId.ToString();
+        var dataElement = Instance.Data.Find(d => d.Id == idAsString);
+        if (dataElement is null)
+        {
+            throw new InvalidOperationException($"Data element with id {idAsString} not found in instance");
+        }
+        //TODO: Add to list of data elements to delete
+        // await _dataClient.DeleteData(_org, _app, _instanceOwnerPartyId, _instanceGuid, dataElementId.Guid, true);
+
+        Instance.Data.Remove(dataElement);
```
martinothamar: Is this meant to use `_dataElementsToDelete` instead?

src/Altinn.App.Core/Internal/Data/CachedInstanceDataAccessor.cs
ivarne: Yes, this is somewhat bad 😊.
I'm thinking that reads to applicationmetadata should not need to be asynchronous (nor should any other file that we can cache in memory), but they currently are asynchronous for the first read of the file, which is why I think the use of `.Result` is appropriate as a temporary workaround. Would you be OK, if I make `GetDataTypeByString` synchronous with a `.Result` on the applicationmetadata read? or do you think everything should be async until we actually have a way to read the file into memory build time?

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
ivarne: I'm not sure. It's only a warning and Microsoft adds nullable annotations between versions. Devs should check for null because I'm pretty sure the app instance is broken beyond repair if anyone tries to actually store `null` in a form data element. Brg also implicitly check for null by checking the Error attribute, although I'm not sure we can fix that with annotations.

It is also an old version of app-lib:
https://altinn.studio/repos/brg/starteogdrive/src/commit/e3b62d445a112e2db65004b90c1f0575ee3c52ab/App/App.csproj#L12

> `<PackageReference Include="Altinn.App.Api" Version="7.13.1">`

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
ivarne: It is not a super clean upgrade, so I'm not sure I'll actually deprecated this (thus the comment). I'm very open to suggestions on how to organise the code in `CachedInstanceDataAccessor`, `DataClient` and the serialisation logic.

src/Altinn.App.Core/Internal/Data/CachedInstanceDataAccessor.cs
ivarne: Thanks!

src/Altinn.App.Core/Helpers/Serialization/ModelDeserializer.cs
ivarne: What method to use depends on how you want to select between json and xml.

If it is from storage, you need to use DeserialiseFromStorage, and pass the DataElement, so that we (in future) can look at AllowedContentTypes to select json or xml.

If the data comes from another source, you can use `DeserializeXml` or `DeserializeJson` or maybe use the (currently commented out) `DeserializeSingleFromRequest` that looks at a httpReqest to negotiate content type.

src/Altinn.App.Core/Helpers/Serialization/ModelSerializationService.cs
```diff
@@ -0,0 +1,267 @@
...
+        XmlWriter xmlWriter = XmlWriter.Create(memoryStream, xmlWriterSettings);
+
+        XmlSerializer serializer = _xmlSerializer.GetSerializer(modelType);
+        serializer.Serialize(xmlWriter, model);
+        if (!memoryStream.TryGetBuffer(out var segment))
+        {
+            throw new InvalidOperationException("Failed to get buffer from memory stream");
+        }
+
+        return segment.AsMemory().RemoveBom();
```
ivarne: I kind of want to avoid the need for copies, and it does not look like `MemoryStream` is renting buffers that gets recycled in dispose, but this should probably be simplified to
```suggestion
        return memoryStream.ToArray().AsMemory().RemoveBom();
```

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
martinothamar: Ahh ok nice

src/Altinn.App.Core/Helpers/Serialization/ModelSerializationService.cs
```diff
@@ -0,0 +1,267 @@
...
+using Altinn.App.Core.Features;
+using Altinn.App.Core.Internal.AppModel;
+using Altinn.Platform.Storage.Interface.Models;
+
+namespace Altinn.App.Core.Helpers.Serialization;
+
+/// <summary>
+/// DI registered service for centralizing (de)serialization logic for data models
+/// </summary>
+public class ModelSerializationService
```
martinothamar, line 15: Since this is public, is it worth hiding it behind an interface? Makes it a little more obvious that it should be retrieved from DI. Should we also make the namespace `Altinn.App.Core.Features.Serialization` or something? I think we talked about centralizing new stuff there in the future

ivarne: Yes, With the addition of `(de)SerialiseFromStorage` and possibly `(de)SerializeFromRequest` it kind of grew from a helper to a service.

martinothamar (approved):
`IInstanceDataAccessor` feels a lot more important now then the name suggest, maybe we should consider a new name before we merge to main?

ivarne:
>IInstanceDataAccessor feels a lot more important now then the name suggest, maybe we should consider a new name before we merge to main?

Do you have any suggestions?

**edit:** We decided to split the interface into IInstanceDataAccessor and IInstanceDataMutator which inherits from the former

ivarne:
Merging this now, so that we get back to having only one PR

PR #814, opened 2024-10-06
Run analyzers that sonar cloud runs

ivarne:
This enables analysers that Sonar Cloud runs so that they can be caught in local development instead of the slow cycle of getting an issue in sonar cloud.

This is part of #730, because I changed these to warning in `.editorconfig` because it made the sonar cloud issues easier to find. I think this could be reviewed separately.

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -111,7 +111,7 @@ csharp_style_namespace_declarations = file_scoped:error
 dotnet_diagnostic.IDE1006.severity = error
 
 # Unused usings
-dotnet_diagnostic.IDE0005.severity = suggestion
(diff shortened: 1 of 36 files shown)


martinothamar:
nice! are you adding `SonarAnalyzer.CSharp` as well or is this just about catching builtin warnings that sonar also complains about?

ivarne:
No, I just took the ones that seemed to be available with a code.

I’m not sure about the static if you don’t access class variables thing. It doesn’t always seem like an improvement to differentiate methods that write telemetry with a static modifier.

ivarne:
This is part of #730, and I seem to have not fixed all warnings here.

PR #815, opened 2024-10-07
Expose endpoint for searching for person using ssn and last name.

bjorntore:
The idea behind this new endpoint is that it should be used by the component in frontend used for looking up people that should be signees.

1. Can we expose person search to the frontend like this? ⚠️ Or should we just confirm the SSNs and last names on the backend before sending correspondence and delegating access? @RonnyB71 
2. Do we just return the full Person object as a response, or have I taken the correct approach using a DTO with a subset of data?


HauklandJ:
Returning a minimized personDTO is imo a better approach than returning the full object. This way we keep the response to only that which is needed to populate the field.

martinothamar (commented):
+1 for using a dedicated type for the API contract

* When do we use `class` and when do we use `struct`? 
* `Dto`-suffix vs `Request`/`Response`?

bjorntore:
> +1 for using a dedicated type for the API contract
> 
>     * When do we use `class` and when do we use `struct`?
> 
>     * `Dto`-suffix vs `Request`/`Response`?

Yep, want answers to these questions. :P

src/Altinn.App.Api/Controllers/PersonSearchController.cs
```diff
@@ -0,0 +1,54 @@
...
+        CancellationToken cancellationToken
+    )
+    {
+        Person? person = await _personClient.GetPerson(
+            personSearchRequest.SocialSecurityNumber,
+            personSearchRequest.LastName,
+            cancellationToken
+        );
+
+        return person is null ? NotFound() : PersonSearchResponse.MapFromPerson(person);
```
ivarne: Is 404 the correct status code when a search returns no results? Isn't it better to have 200 and modify `PersonSearchResponse` to have a `Success` property. The code is mostly for looking at logs, and a miss is a valid result and no reason to investigate.

Somehow Swagger seems to think that you return `ProblemDetails` for the 404, so if we keep the code, it need to actually return what we promised.

martinothamar: alternatively 204 No Content

bjorntore: @ivarne Take another look.

HauklandJ (approved):

Issue #833, opened 2024-10-10
Application crashes when ASCII control characters are input into any field


trymval:
### Description of the bug

When a user inputs certain ASCII control characters into any input field, the entire application crashes (backend throws an exception). Instead of crashing, the application should handle such inputs gracefully by sanitizing the input, preventing the entry of control characters, or displaying an appropriate error message to the user.

### Steps To Reproduce
1. Open any altinn3 application.
2. Navigate to any input field (e.g., text box, form field)
3. Input any of the following ASCII control characters:
    - 0x01 (SOH - Start of Heading)
    - 0x02 (STX - Start of Text)
    - 0x03 (ETX - End of Text)
    - 0x04 (EOT - End of Transmission)
    - 0x05 (ENQ - Enquiry)
    - 0x06 (ACK - Acknowledge)
    - 0x07 (BEL - Bell)
    - 0x08 (BS - Backspace)
    - 0x0B (VT - Vertical Tab)
    - 0x0C (FF - Form Feed)
    - 0x0E (SO - Shift Out)
    - 0x0F (SI - Shift In)
    - 0x10 (DLE - Data Link Escape)
    - 0x11 (DC1 - Device Control 1)
    - 0x12 (DC2 - Device Control 2)
    - 0x13 (DC3 - Device Control 3)
    - 0x14 (DC4 - Device Control 4)
    - 0x15 (NAK - Negative Acknowledge)
    - 0x16 (SYN - Synchronous Idle)
    - 0x17 (ETB - End of Transmission Block)
    - 0x18 (CAN - Cancel)
    - 0x19 (EM - End of Medium)
    - 0x1A (SUB - Substitute)
    - 0x1B (ESC - Escape)
    - 0x1C (FS - File Separator)
    - 0x1D (GS - Group Separator)
    - 0x1E (RS - Record Separator)
    - 0x1F (US - Unit Separator)

4. Observe that the application crashes upon processing the input.

### Additional Information
**Tested on following versions:**
- Altinn.App.Core: 7.15.2, 8.3.5, 8.5.0-preview.7
- Altinn.App.Api: 7.15.2, 8.3.5, 8.5.0-preview.7
- Frontend: 3.68.6, 4.8.2, 4.9.0-rc.1

**Error Message from API/Backend:**
```
There was an error generating the XML document. '', hexadecimal value 0x02, is an invalid character.
```
**Relevant Stack Trace:**
```
System.InvalidOperationException:
[...]

PR #838, opened 2024-10-14
Handle invalid XML chars

martinothamar:
## Description
There are some characters, e.g. ASCII control characters, that aren't supported in XML strings. Options:

* Just remove unsupported characters
* Replace invalid XML characters with the `�` so that we still output a symbol (user can then deal with it)
* Validate (return validation issues)

## Related Issue(s)
- #833

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Helpers/ObjectUtils.cs
+++ b/src/Altinn.App.Core/Helpers/ObjectUtils.cs
@@ -170,6 +170,8 @@ public static partial class ObjectUtils
         }
     }
 
+    // Regex copied from: https://stackoverflow.com/a/961504
+    // Which is based on spec: https://www.w3.org/TR/xml/#charsets
     [GeneratedRegex(
         @"(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]|[\uD800-\uDBFF](?![\uDC00-\uDFFF])|[\x00-\x08\x0B\x0C\x0E-\x1F\x7F-\x9F\uFEFF\uFFFE\uFFFF]"
     )]
--- a/src/Altinn.App.Core/Helpers/ObjectUtils.cs
+++ b/src/Altinn.App.Core/Helpers/ObjectUtils.cs
@@ -154,7 +154,7 @@ public static partial class ObjectUtils
                         if (prop.SetMethod is not null)
                         {
                             // Remove invalid xml characters
-                            prop.SetValue(model, xmlRegex.Replace(s, "�"));
+                            prop.SetValue(model, xmlRegex.Replace(s, "\uFFFD")); // \uFFFD is the unicode replacement character �
                         }
                         else if (xmlRegex.IsMatch(s))
                         {
--- a/src/Altinn.App.Core/Helpers/ObjectUtils.cs
+++ b/src/Altinn.App.Core/Helpers/ObjectUtils.cs
@@ -147,17 +147,14 @@ public static partial class ObjectUtils
                     }
                     else
                     {
-                        var xmlRegex = XmlInvalidCharsRegex();
                         if (prop.SetMethod is not null)
                         {
+                            // If a property doesn't have a setter, it hopefully doesn't have user input,
+                            // and therefore it far less likely to have invalid XML chars. If that were the case
+                            // we will still just error out when serializing to XML
+
                             // Remove invalid xml characters
-                            prop.SetValue(model, xmlRegex.Replace(s, "\uFFFD")); // \uFFFD is the unicode replacement character �
-                        }
-                        else if (xmlRegex.IsMatch(s))
-                        {
-                            throw new InvalidOperationException(
-                                $"Property {prop.Name} of datamodel contains invalid xml characters"
-                            );
+                            prop.SetValue(model, XmlInvalidCharsRegex().Replace(s, "\uFFFD")); // \uFFFD is the unicode replacement character �
                         }
                     }
                 }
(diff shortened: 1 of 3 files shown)


src/Altinn.App.Core/Helpers/ObjectUtils.cs
```diff
@@ -137,14 +138,31 @@ public static void PrepareModelForXmlStorage(object model, int depth = 64)
...
+                            prop.SetValue(model, null);
+                        }
+                    }
+                    else
+                    {
+                        var xmlRegex = XmlInvalidCharsRegex();
+                        if (prop.SetMethod is not null)
+                        {
+                            // Remove invalid xml characters
+                            prop.SetValue(model, xmlRegex.Replace(s, "�"));
```
ivarne: I've had a few instances where editors mess up unicode, so I tend to prefere to use the unicode escapes in string literals. Especially here when it is the error character.

```suggestion
                            prop.SetValue(model, xmlRegex.Replace(s, "\uFFFD")); // \uFFFD is the unicode replacement character �
```

src/Altinn.App.Core/Helpers/ObjectUtils.cs
```diff
@@ -137,14 +138,31 @@ public static void PrepareModelForXmlStorage(object model, int depth = 64)
...
+                        }
+                        else if (xmlRegex.IsMatch(s))
+                        {
+                            throw new InvalidOperationException(
+                                $"Property {prop.Name} of datamodel contains invalid xml characters"
```
ivarne: This looks wrong. This error shows up if
1. value is string
2. not string.IsNullOrWhitespace
3. prop.SetMethod is null
4. xmlRegex.IsMatch(s)

Especially the 3 seems unlikely to be true, and the message does not make sense in that case.

src/Altinn.App.Core/Helpers/ObjectUtils.cs
```diff
@@ -137,14 +138,31 @@ public static void PrepareModelForXmlStorage(object model, int depth = 64)
...
-                    && string.IsNullOrWhiteSpace(s)
-                    && prop.GetCustomAttribute<XmlTextAttribute>() is not null
-                )
+                if (value is string s)
                 {
-                    // Ensure empty strings are set to null
-                    prop.SetValue(model, null);
+                    if (string.IsNullOrWhiteSpace(s))
+                    {
+                        if (prop.GetCustomAttribute<XmlTextAttribute>() is not null)
```
ivarne: I don't think the nesting here makes sense. We should continue matching regex to remove FORM-FEED ++ when the string is whitespace and not `[XmlText]`.

src/Altinn.App.Core/Helpers/ObjectUtils.cs
```diff
@@ -155,6 +173,11 @@ value is string s
         }
     }
 
+    [GeneratedRegex(
+        @"(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]|[\uD800-\uDBFF](?![\uDC00-\uDFFF])|[\x00-\x08\x0B\x0C\x0E-\x1F\x7F-\x9F\uFEFF\uFFFE\uFFFF]"
+    )]
```
ivarne, line 174: Can you explain this regex?

ivarne:
I introduced `ObjectUtils`, for very specific tasks, but don't really like it. How would you do this without ObjectUtils? Why would it be worse?

src/Altinn.App.Core/Helpers/ObjectUtils.cs
martinothamar: I didn't have this branch orignally, but ran into it because some of the models have `public string Prop => ""` kind of properties

ivarne: I don't understand what branch?

I'm trying to understand when you want to throw exception and when you want to just replace the offending characters,

src/Altinn.App.Core/Helpers/ObjectUtils.cs
martinothamar: Added comment referring to the origin. I think regexr is a better tool for explanation of the actual regex: https://regexr.com/870t1

src/Altinn.App.Core/Helpers/ObjectUtils.cs
martinothamar: It replaces when it can (the property _isnt_ readonly), and throws error when it can't (the property is readonly)

ivarne: Now it makes sense.

I would just assume that the readonly fields doesn't take user updates (they can't be set via json deserialisation), so we don't need to inspect them. If we do conclude that it makes sense to validate readonly this is a sensible approach.

martinothamar: yeah it's not strictly necessary, we will just get a similar error from XmlSerializer if it ends up getting serialized (and has invalid chars)

martinothamar:
> How would you do this without ObjectUtils? Why would it be worse?

I'm not `ObjectUtils` biggest fan either, but I don't think we have time for larger changes (this was the simplest change I could think of)

src/Altinn.App.Core/Helpers/ObjectUtils.cs
ivarne: To be strictly correct you would need to check for `[XmlIgnore]`. But I should probably not suggest more branches in order suggest removing a branch.

bjorntore (approved):

Issue #840, opened 2024-10-16
Enhetsregister integration endpoint


HauklandJ:
Similar to folkeregister endpoint: https://github.com/Altinn/app-lib-dotnet/issues/799, PR: https://github.com/Altinn/app-lib-dotnet/pull/815

Vi har eksisterende client, RegisterERClient, trenger dto og controller

PR #847, opened 2024-10-17
API Endpoint for getting org from Enhetsregisteret

cammiida:
## Description
GET endpoint for getting organisation from Enhetsregisteret (ER) by organisation number.

## Related Issue(s)
- #840

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Api/Controllers/OrganisationSearchController.cs
```diff
@@ -0,0 +1,53 @@
...
+/// <summary>
+/// This controller class provides Enhetsregisteret (ER) organisation search functionality.
+/// </summary>
+[AutoValidateAntiforgeryTokenIfAuthCookie]
+[ApiController]
+[Produces(MediaTypeNames.Application.Json)]
+[Consumes(MediaTypeNames.Application.Json)]
+[ResponseCache(Duration = 0, Location = ResponseCacheLocation.None, NoStore = true)]
+[Route("{org}/{app}/api/v1/organisations")]
+public class OrganisationSearchController : ControllerBase
```
bjorntore: I just renamed the controller for looking up a person to "LookupPersonController". I felt like this is a lookup of one specific item, not necessarily a search, as we only accept a complete orgNr, not a partial one. 

If you agree, maybe rename this one as well to "LookupOrganisationController". And same with response object.

src/Altinn.App.Api/Controllers/OrganisationSearchController.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// <summary>
+    /// Allows searching for an organisation in ER
+    /// </summary>
+    /// <param name="orgNr">Payload that contains params for executing a search for an organisation.</param>
+    /// <returns>A <see cref="OrganisationSearchResponse"/> object.</returns>
+    [HttpGet]
+    [Route("{orgNr}")]
+    [ProducesResponseType(typeof(OrganisationSearchResponse), StatusCodes.Status200OK)]
+    public async Task<ActionResult<OrganisationSearchResponse>> SearchForOrganisation([FromRoute] string orgNr)
```
bjorntore: I'm 100% comfortable with this being a GET with orgNr as route param, but I still want to ask:
Consider following same pattern as the one for Person, just to keep these two "lookups" consistent with each other?
Meaning POST with a JSON request object that contains the parameters. Then the frontend components can be more similar as well.

The reason it was done that way for person was to avoid having the SSN in the URL, so I get that the same reason is not valid here.

bjorntore (approved):

src/Altinn.App.Api/Controllers/OrganisationSearchController.cs
cammiida: Agreed!

src/Altinn.App.Api/Controllers/OrganisationSearchController.cs
cammiida: I see what you mean, and I did consider it. But I think I'd rather adhere to REST guidelines as a base case, and make exceptions when it makes sense, like for the person lookup. For the frontend, this needs to separate queries anyways, so there won't be much difference either way.

Feel free to disagree 😅

src/Altinn.App.Api/Controllers/LookupOrganisationController.cs
```diff
@@ -39,15 +39,15 @@ ILogger<OrganisationSearchController> logger
     /// Allows searching for an organisation in ER
     /// </summary>
     /// <param name="orgNr">Payload that contains params for executing a search for an organisation.</param>
```
bjorntore: Nitpick: Description for orgNr is outdated. It references a payload with params inside, which we don't do here.

src/Altinn.App.Api/Controllers/LookupOrganisationController.cs
```diff
@@ -16,19 +16,19 @@ namespace Altinn.App.Api.Controllers;
...
 {
     private readonly IOrganizationClient _organisationClient;
-    private readonly ILogger<OrganisationSearchController> _logger;
+    private readonly ILogger<LookupOrganisationController> _logger;
 
     /// <summary>
-    /// Initialize a new instance of <see cref="OrganisationSearchController"/> with the given services.
+    /// Initialize a new instance of <see cref="LookupOrganisationController"/> with the given services.
     /// </summary>
     /// <param name="organisationClient">A client for searching for an organisation in ER.</param>
```
bjorntore: Nitpick: Replace the term "searching" with lookup in documentation comments.

cammiida: Haha du er grundig. Veldig nice 👏

src/Altinn.App.Api/Controllers/LookupOrganisationController.cs
```diff
@@ -16,19 +16,19 @@ namespace Altinn.App.Api.Controllers;
 [Consumes(MediaTypeNames.Application.Json)]
 [ResponseCache(Duration = 0, Location = ResponseCacheLocation.None, NoStore = true)]
 [Route("{org}/{app}/api/v1/organisations")]
```
bjorntore, line 18: When I renamed the controller for person I also changed the route to be:
[Route("{org}/{app}/api/v1/lookup/person")]

That would mean this for organisation, if you agree on that URL. 
`[Route("{org}/{app}/api/v1/lookup/organisation")]`
We can also flip it around and have person/oranisation first, but I grouping them in the swagger doc under "Lookup" felt kind of nice.

src/Altinn.App.Api/Controllers/LookupOrganisationController.cs
bjorntore: Frivillig å bry seg om slike nitpicks altså. :P

src/Altinn.App.Api/Controllers/LookupOrganisationController.cs
cammiida: Hm, I'm not sure if I agree with this approach. According to REST a `{noun}/{id}` _is_ a lookup. Won't it be redundant then to specify `/lookup` in the URL?

src/Altinn.App.Api/Controllers/OrganisationSearchController.cs
cammiida: Question: To adhere to grammatical correctness, should it be `LookUpOrganisation/Person` if it is the action of looking up, or should it be `Organisation/PersonLookup` if it is the noun?

src/Altinn.App.Api/Controllers/LookupOrganisationController.cs
bjorntore: Yea, made more sense when it was in RPC form (aka a POST with a payload).
For person the URL isn't `{noun}/{id}`.

bjorntore: It's also not super clear in our URLs what is normal REST resources and what is RPC calls.

bjorntore: The register API has the sensitive params as headers, which would allow us to not have a body on the person request, but figured that was more hassle for the frontend.

src/Altinn.App.Api/Controllers/OrganisationSearchController.cs
bjorntore: I don't have a super strong opinion on naming/urls, but we should just keep in mind that some TE are probably gonna want to use these two endpoints directly through the app API, so they just need be logical and easy to understand.

src/Altinn.App.Api/Controllers/LookupOrganisationController.cs
cammiida: Ah, that makes more sense. Although, I have believed it is more best practice to structure the urls like `{resource}/{action}`, like `person/lookup`, if we're following a REST paradigm (are we, or is it more RPC?). Although I can see that it's nice to group them in swagger 😄 But I might be wrong!

bjorntore: Ye, we can change it to that. I was 51% in favor of what I did, and now I'm 50/50. :P Feel free to change it both places.

cammiida: Hahahah

HauklandJ (approved):
LGTM

PR #848, opened 2024-10-17
Fix NullReferenceException when using a gateway before instance is created

chrrust:
A fix that allows instantiation of `CachedInstanceDataAccessor` when the instance hasn't been created yet. The purpose of this is to allow a `IProcessExclusiveGateway` implementation to run directly after the start event.

## Description

The problem initially was that the app and org was derived from `instance.AppId`, but because the instance was not created yet, a null reference exception was thrown when trying to Split on it. 

Because `CachedInstanceDataAccessor` already had a dependency on `IAppMetadata`, I changed to using that instead.

I was thinking about changing the signature of `CachedInstanceDataAccessor.ctor` to accept the `ApplicationMetadata` directly instead of the `IAppMetadata`. This is because `IAppMetadata` is used only to retrieve the `ApplicationMetadata` through `IAppMetadata.GetApplicationMetadata` anyways. If this seems like a reasonable change, I can make a new commit with this change. One of the reasons I was thinking about this is because of how the verification in `ProcessTaskFinalizerTests.cs` is done. That test probably shouldn't need to know that `IAppMetadata.GetApplicationMetadata` is invoked inside `CachedInstanceDataAccessor`(?).

Another issue that arised was that the `instance.Id` will be null in this case. A null check sufficed to allow the creation of `CachedInstanceDataAccessor` regardless.

Finally, I couldn't find any test class where the `CachedInstanceDataAccessor` is  tested itself. Do you want me to write a test for this fix, and if so, where should I put it?

## Related Issue(s)
N/A

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Data/CachedInstanceDataAccessor.cs
+++ b/src/Altinn.App.Core/Internal/Data/CachedInstanceDataAccessor.cs
@@ -18,8 +18,6 @@ namespace Altinn.App.Core.Internal.Data;
 internal sealed class CachedInstanceDataAccessor : IInstanceDataMutator
 {
     // DataClient needs a few arguments to fetch data
-    private readonly string _org;
-    private readonly string _app;
     private readonly Guid _instanceGuid;
     private readonly int _instanceOwnerPartyId;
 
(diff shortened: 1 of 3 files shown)


ivarne (approved):
Looks good.

martinothamar (approved):

Issue #850, opened 2024-10-18
New Post endpoint for creating data elements with PATCH like response


ivarne:
### Description

 Service owners want to make changes to the dataElements when attachments gets uploaded, but the current POST endpoint on DataController, does not return the updated instance.Data (with new and deleted data elements), and changed validation issues.

The proposal is for a new URL to work in addition to the previous url where the dataType is a url parameter instead of a GET parameter.
```
// Previous url
/{org}/{app}/instances/{instanceOwnerPartyId}/{instanceGuid}/data?dataType=xx

// New url
/{org}/{app}/instances/{instanceOwnerPartyId}/{instanceGuid}/data/{dataType}
```
This means that we keep backwards compatibility for current API users, but new versions of frontend can use the new url to get a richer response back.

### Additional Information

Another benefit is that we can (in the future) hack into the openApi generation process and registrer special urls for all DataTypes in applicationmetadata.json so that the actual data models is exposed.

PR #853, opened 2024-10-20
Call IDataWriteProcessor on all data element mutations and add new POST endpoint with better response type

ivarne:
This should be simple and (mostly)backwards compatible changes before adding new hooks for data processing and file validation.

The commits are fairly well separated, so they might be reviewed individually if you want smaller chunks.

## Related Issue(s)
- #850 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Internal/Patch/PatchService.cs
+++ b/src/Altinn.App.Core/Internal/Patch/PatchService.cs
@@ -26,7 +26,7 @@ internal class PatchService : IPatchService
     private readonly IDataClient _dataClient;
     private readonly IInstanceClient _instanceClient;
     private readonly ModelSerializationService _modelSerializationService;
-    private readonly IWebHostEnvironment _hostingEnvironment;
+    private readonly IHostEnvironment _hostingEnvironment;
     private readonly Telemetry? _telemetry;
     private readonly IValidationService _validationService;
     private readonly IEnumerable<IDataProcessor> _dataProcessors;
@@ -46,7 +46,7 @@ internal class PatchService : IPatchService
         IEnumerable<IDataProcessor> dataProcessors,
         IEnumerable<IDataWriteProcessor> dataWriteProcessors,
         ModelSerializationService modelSerializationService,
-        IWebHostEnvironment hostingEnvironment,
+        IHostEnvironment hostingEnvironment,
         Telemetry? telemetry = null
     )
     {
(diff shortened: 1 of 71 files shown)


src/Altinn.App.Core/Internal/Patch/PatchService.cs
```diff
@@ -202,6 +203,24 @@ await RunDataProcessors(
         };
     }
 
+    public async Task<List<ValidationSourcePair>> RunIncrementalValidation(
```
martinothamar, line 223: What's the rationale for adding this?

martinothamar (approved):

HauklandJ (approved):
:raised_hands:

src/Altinn.App.Core/Internal/Patch/PatchService.cs
ivarne: Just that I did not want to inject the full ValidationService twice for PatchService and directly in the controller.

martinothamar: Would prefer to just inject another service tbh 😄

ivarne: Yes, when we get your patch to dynamically initialise the validators, so that it don't happen twice for every request (service owners does strange things in their constructors), it would be my preference too.

Actually I would prefer to not have PatchService in the first place, because it doesn't have natural boundaries, but it was added in code review to make DataController shorter 🥲.

src/Altinn.App.Core/Models/DataElementChanges.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// <summary>
+    /// The data element was deleted and will have <see cref="DataElementChange.DataElement"/> set
+    /// </summary>
+    Deleted
+}
+
+/// <summary>
+/// Representation of a change to a binary data element
+/// </summary>
+public sealed class BinaryChange : DataElementChange
```
martinothamar: `BinaryDataChange` ?

martinothamar (approved):

src/Altinn.App.Core/Models/DataElementChanges.cs
ivarne: 🎉 👍

PR #866, opened 2024-10-28
Use IUserTokenProvider instead of IHttpContextAccessor in InstanceClient

ivarne:
Writing tests for InstanceDataUnitOfWork I found a curious difference between DataClient and InstanceClient, and figured out that there was no reason for the difference in service dependencies, so I just updated InstanceClient to use the same abstraction as DataClient.

The replaced code does just the same job, using a common and easily mocked service

https://github.com/Altinn/app-lib-dotnet/blob/3bb474ec93e4b55c88e3dfda65f0785001f30f96/src/Altinn.App.Core/Implementation/UserTokenProvider.cs#L41

A funny difference is that the old InstanceClient used `IOptionsMonitor` to get the cookie name, but as it is rather impossible to change, and not monitored anywhere else, I think it makes sense to remove that feature.

## Verification
- [ ] All tests run green

## Documentation


martinothamar (approved):
nice!

PR #867, opened 2024-10-29
Ensure release label on PR

ivarne:
Copied and adapted from https://github.com/Altinn/app-frontend-react/blob/main/.github/workflows/check-label-added.yml

Can be tested by removing and adding labels


martinothamar (approved):
nice 🥇

PR #870, opened 2024-10-30
Implement IInstanceDataMutator.AbandonAllChanges

ivarne:
This enables IDataWriteProcessor to signal that the current change set is bad, and should be ignored, with a list of errors that can be shown to the user.

```C#
public class DataWriteProcessor : IDataWriteProcessor
{
    public Task ProcessDataWrite(
        IInstanceDataMutator instanceDataMutator,
        string taskId,
        DataElementChanges changes,
        string? language
    )
    {
        var fileDataChanges = changes.BinaryDataChanges.Where(d => d.DataType.Id == "file").ToArray();
        if (fileDataChanges is [var fileDataChange])
        {
            if (fileDataChange.FileName == "newFileName")
            {
                instanceDataMutator.AbandonAllChanges(
                    [new() { Severity = ValidationIssueSeverity.Error, Description = "hello, this is strange" }]
                );
            }
        }

        return Task.CompletedTask;
    }
}
```

This also fixes a bug where POST data element (used to create subform instances) created duplicates.

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #875, opened 2024-10-31
Add tests for data element Post and fix bug with AbandonChanges

ivarne:
Fix a bug where you had to add more than 1 issues (instead of at least 1 issue) in `AbandonAllChanges`. 

## Related Issue(s)
- This includes tests for  #870 

## Verification
- [ ] All tests run green

## Documentation


martinothamar (approved):
Woops, missed this. Maybe a good candidate for making a `HasAbandonIssues` prop to keep the conditional only one place?

ivarne:
> Maybe a good candidate for making a HasAbandonIssues prop to keep the conditional only one place?

Or just a `ProblemDetails GetAbandonResponse()` to internalise the mapping to a `ProblemDetails` instance as well?

martinothamar:
I would not use the ASP.NET Core `ProblemDetails` model directly in there, I think that responsibility (deciding what to return in the API) should lie in the API/outer layer. Maybe an extension method or something?

The `IInstanceMutator` property could be something like `HasAbandonIssues(out IEnumerable<AbandonIssue> issues)` to access the actual data?

src/Altinn.App.Core/Internal/Patch/IPatchService.cs
```diff
@@ -1,45 +0,0 @@
...
-using Altinn.App.Core.Models.Validation;
-using Altinn.Platform.Storage.Interface.Models;
-using Json.Patch;
-
-namespace Altinn.App.Core.Internal.Patch;
-
-/// <summary>
-/// Service for handling JsonPatches to data elements.
-/// </summary>
-public interface IPatchService
```
martinothamar, line 13: This along with `InternalPatchService` doesn't feel great. Have you run a search to make sure noone is referencing this?

martinothamar (approved):

src/Altinn.App.Core/Internal/Patch/IPatchService.cs
ivarne: It's been pretty heavily changed in 8.5.0-x anyway, so using it won't be problem free.

Issue #763, opened 2024-09-11
Authorization scenarios and mechanisms analysis


martinothamar:
### Description

We need to get an overview of common authorization scenarios and mechanisms that exist today and will be implemented into the platform. Here are some movements happening where there are questions related to authorization.

* Simplified integration for system owners: https://github.com/digdir/roadmap/issues/84
  * Client systems could be "fagsystemer", "sluttbrukersystemer", "leverandører" etc 
* "System users": https://github.com/Altinn/altinn-authentication/issues/200
* App authorization
  * Required due to various internal operations and integrations, read more at  https://github.com/Altinn/altinn-decision-log/issues/28 and  https://github.com/Altinn/app-lib-dotnet/issues/813
* Various upcoming features and integrations
  * User-controlled and parallel signing 
  * BPMN service tasks, background work (Correspondence, Dialogporten)
* Simplify and improve current integrations
  * eFormidling, PDF
* Avoid token exchange in the future?
  * Has been discussed, will probably be handled through PEP  

### In scope

* Map out the inbound flow and possible contexts
* Map out the outbound flow (which tokens and mechanisms are used)
* Start a design to better abstract authenticated state and context

#### Mechanisms for authorization

These methods are supported through Altinn Authentication and Authorization through PEP and PDP.

* User token (logged in through Altinn)
  * Authorized as: person
  * Access control: roles, instance delegations, dedicated XACML rules
* Org token
  * Authorized as: org
  * Access control: dedicated XACML rules in app (i.e. `[ORG]`)
  * Always exchanged to an Altinn token, we don't support Maskinporten directly (that would be done in PEP).
  * To get the right permissions, [we have to use serviceowner scopes here, as that  gives us the `urn:altinn:org` claim in the resulting Altinn token](https://github.com/Altinn/altinn-authentication/blob/231ee42530df35e793c2cabe1093f6204bf35c3d/src/Authentication/Controllers/AuthenticationController.cs#L435-L447)
* Systemuser
  * Authorized as: system user
  * Access control: roles, system user delegations (what actions the system user owner approves that the system can do on behalf of the system user owner)
  * A system user is a special Maskinporten client (in principle) using Rich Authorization Request (RAR) protocol to enrich Maskinporten tokens with Altinn system user details (`systemuser_id`, `system_id`)
* App (WIP)
  *  Authorized as: app within an org
  * Access control: XACML rule either manually written or injected (TBD)
[...]

PR #880, opened 2024-11-04
`IAuthenticationContext` for current auth information, systemuser support

martinothamar:
## Description
Initial proposal for https://github.com/Altinn/app-lib-dotnet/issues/763

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/763
- #652 
- #908 
- https://github.com/Altinn/app-frontend-react/pull/2889

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Api/Controllers/AuthenticationController.cs
```diff
@@ -15,16 +18,137 @@ public class AuthenticationController : ControllerBase
...
-    public AuthenticationController(IAuthenticationClient authenticationClient, IOptions<GeneralSettings> settings)
+    public AuthenticationController(
+        IAuthenticationClient authenticationClient,
+        IOptions<GeneralSettings> settings,
+        IServiceProvider serviceProvider
+    )
     {
         _authenticationClient = authenticationClient;
         _settings = settings.Value;
+        _authenticationContext = serviceProvider.GetRequiredService<IAuthenticationContext>();
```
ivarne: Could you add a comment about why you inject `IServiceProvider` instead of `IAuthenticationContext`?

src/Altinn.App.Api/Controllers/AuthenticationController.cs
```diff
@@ -15,16 +18,137 @@ public class AuthenticationController : ControllerBase
...
     }
 
+    /// <summary>
+    /// Gets current party by reading cookie value and validating.
+    /// </summary>
+    /// <returns>Party id for selected party. If invalid, partyId for logged in user is returned.</returns>
+    [Authorize]
+    [ResponseCache(Duration = 0, Location = ResponseCacheLocation.None, NoStore = true)]
+    [HttpGet("{org}/{app}/api/[controller]/current")]
```
ivarne: Why `/[controller]/` instead of `authentication` (with a lower case `a`)

src/Altinn.App.Api/Controllers/AuthenticationController.cs
```diff
@@ -15,16 +18,137 @@ public class AuthenticationController : ControllerBase
...
+                throw new Exception($"Unexpected authentication context: {current.GetType().Name}");
+        }
+    }
+
+    [JsonDerivedType(typeof(UnauthenticatedResponse), typeDiscriminator: "Unauthenticated")]
+    [JsonDerivedType(typeof(UserResponse), typeDiscriminator: "User")]
+    [JsonDerivedType(typeof(OrgResponse), typeDiscriminator: "Org")]
+    [JsonDerivedType(typeof(ServiceOwnerResponse), typeDiscriminator: "ServiceOwner")]
+    [JsonDerivedType(typeof(SystemUserResponse), typeDiscriminator: "SystemUser")]
+    private abstract record CurrentAuthenticationBaseResponse { }
```
ivarne: I'm not sure I like the idea of exposing a discriminated union in the api. Does OpenAPI generation support this nicely?

At least we need some docs on the properties

src/Altinn.App.Api/Controllers/AuthenticationController.cs
```diff
@@ -15,16 +18,137 @@ public class AuthenticationController : ControllerBase
...
+
+    private sealed record UnauthenticatedResponse : CurrentAuthenticationBaseResponse { }
+
+    private sealed record UserResponse : CurrentAuthenticationBaseResponse
+    {
+        public required UserProfile Profile { get; init; }
+
+        public required Party Party { get; init; }
```
ivarne: Is this the party of the authenticated user or the party he currently intends to represent as specified by the `AltinnPartyId` cookie (which is kind of useless because it will be wrong for one of the tabs if you have multiple instances for multiple reportees open at the same time )?

src/Altinn.App.Core/Internal/Auth/IAuthenticationContext.cs
```diff
@@ -0,0 +1,670 @@
...
+            return _extra;
+        }
+    }
+
+    /// <summary>
+    /// The logged in client is a user (e.g. Altinn portal/IDporten) with auth level 0.
+    /// This means that the user has authenticated with a username/password, which can happen using
+    /// * Altinn "self registered users"
+    /// * IDporten through Ansattporten ("low"), MinID self registered eID
+    /// These have limited access to Altinn and can only represent themselves.
```
ivarne: Is this correct? At least if you come through Ansattporten, you likely represent the org you're employed at?

src/Altinn.App.Api/Infrastructure/Middleware/AuthenticationContextMiddleware.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    /// <summary>
+    /// Invokes the middleware to process the HTTP context.
+    /// </summary>
+    /// <param name="context">The HTTP context.</param>
+    public Task InvokeAsync(HttpContext context)
+    {
+        var authenticationContext = (AuthenticationContext)
+            context.RequestServices.GetRequiredService<IAuthenticationContext>();
+        authenticationContext.ResolveCurrent(context);
```
ivarne: Why do you need a middleware here? Don't we already have a middleware that sets `IHttpContextAccessor.Current`?

src/Altinn.App.Core/Internal/Auth/IAuthenticationContext.cs
ivarne, line 1: This file has too many classes, records and interfaces. Would be nice to split it up.

ivarne (commented):
Fin opprydding i mange av testene og autorisasjon :).

Har kommentert litt ved gjennomlesning

src/Altinn.App.Api/Controllers/AuthenticationController.cs
ivarne: You also need to somehow annotate for swagger to document the possible response(s).

src/Altinn.App.Api/Controllers/AuthenticationController.cs
martinothamar: in code or just here? It's because the interface is `internal` for now

src/Altinn.App.Core/Internal/Auth/IAuthenticationContext.cs
martinothamar: I think you end up as a user, and not with `authentication_level = 0`? I'll check...

src/Altinn.App.Core/Internal/Auth/IAuthenticationContext.cs
martinothamar: Will do

src/Altinn.App.Api/Infrastructure/Middleware/AuthenticationContextMiddleware.cs
martinothamar: I don't remember if there is a good reason anymore, I'll try to retrieve it lazily instead, without middleware

src/Altinn.App.Api/Controllers/AuthenticationController.cs
martinothamar: I might remove or comment this out. I was orginally intending to replace 3 separate frontend requests with this endpoint, but put that on the backburner as it seemed difficult to refactor the frontend at the time. Wether a union is a good idea I'm not sure. I did have the start of a TS implementation that used the discriminator to figure which type it was, and a .NET client would be fine with it, but yeah it's not the most common thing

HauklandJ (approved):
lgtm

src/Altinn.App.Core/Features/Auth/IAuthenticationContext.cs
```diff
@@ -0,0 +1,108 @@
...
+    internal static void AddAuthenticationContext(this IServiceCollection services)
+    {
+        services.TryAddSingleton<IAuthenticationContext, AuthenticationContext>();
+    }
+}
+
+/// <summary>
+/// Provides access to the current authentication context.
+/// </summary>
+public interface IAuthenticationContext
```
danielskovli, line 6: Separate file for the interface and extension? Not sure if we actually have a guideline on this 🤷

src/Altinn.App.Core/Features/Auth/TokenIssuer.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <summary>
+    /// Token is from Altinn Studio
+    /// </summary>
+    AltinnStudio,
+
+    /// <summary>
+    /// Token is from external IDporten, e.g. SBS
+    /// </summary>
+    IDporten,
```
danielskovli, line 31: Should we add Ansattporten? It's not technically the same as ID-porten, although it's very close. Whether that's reflected on our end or not, I don't know.

src/Altinn.App.Core/Features/Auth/Authenticated.cs
```diff
@@ -0,0 +1,795 @@
...
+            Func<string, Task<Party>> lookupParty
+        )
+            : base(tokenIssuer, tokenIsExchanged, token)
+        {
+            SystemUserId = systemUserId;
+            SystemUserOrgNr = systemUserOrgNr;
+            SystemId = systemId;
+            // System user tokens can either be raw Maskinporten or exchanged atm.
+            // If the token is not exchanged, we don't have these claims and so we default to what altinn-authentication currently does.
+            AuthenticationLevel = authentictaionLevel ?? 3;
```
danielskovli: Typo for `authentictaionLevel` 🙂

src/Altinn.App.Core/Features/Auth/IAuthenticationContext.cs
```diff
@@ -0,0 +1,108 @@
...
+                    _authorizationClient.GetUserRoles,
+                    _appMetadata.GetApplicationMetadata
+                );
+                httpContext.Items[ItemsKey] = authInfo;
+            }
+            else
+            {
+                authInfo =
+                    authInfoObj as Authenticated
+                    ?? throw new Exception("Unexpected type for authentication info in HTTP context");
```
danielskovli: Another one that we don't have a good guideline on, but should this be a more narrow type than `Exception`?

src/Altinn.App.Core/Features/Auth/Scopes.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    public override bool Equals(object? obj) => obj is Scopes other ? Equals(other) : false;
+
+    public override int GetHashCode() => _scope?.GetHashCode() ?? 0;
+
+    public override string ToString() => _scope ?? "";
+
+    private static readonly SearchValues<char> _whitespace = SearchValues.Create(" \t\n");
+    private static readonly SearchValues<char> _alphaNumeric = SearchValues.Create(
+        "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZÆØÅabcdefghijklmnopqrstuvwxyzæøå"
```
danielskovli: Is this exchaustive? Like, do we know which symbols are allowed in Maskinporten/Altinn Auth? Perhaps the most important question is: does that even matter with the logic in this type? 🤷

src/Altinn.App.Core/Infrastructure/Clients/Authentication/AuthenticationClient.cs
```diff
@@ -16,36 +15,36 @@ namespace Altinn.App.Core.Infrastructure.Clients.Authentication;
...
         _client = httpClient;
+        _authenticationContext = authenticationContext;
     }
 
     /// <inheritdoc />
     public async Task<string> RefreshToken()
     {
         string endpointUrl = $"refresh";
-        string token = JwtTokenUtil.GetTokenFromContext(_httpContextAccessor.HttpContext, General.RuntimeCookieName);
+        string token = _authenticationContext.Current.Token; // TODO: check if authenticated?
```
danielskovli, line 47: Would verifying authenticated state here be a change of behaviour, or would `endpointUrl` deal with this eventually anyway? 

If the latter, it seems like an easy win to catch expired token here before passing it on.

danielskovli (approved):
Love it 🥇

src/Altinn.App.Core/Features/Auth/Scopes.cs
danielskovli: Never mind, this is handled differently in https://github.com/Altinn/app-lib-dotnet/pull/1058/files#diff-785dcda9ae4ec7d7311c26e2cc4e6237f1cf2eb5fa6a1faffa80f2104c60603bR104

src/Altinn.App.Core/Features/Auth/TokenIssuer.cs
martinothamar: I don't think it's possible to get an Altinn token through Ansattporten? Altinn portal = IDporten, and there is no exchange support in altinn-authentication. AFAICT Ansattporten is a separate endpoint with different metadata etc right?

danielskovli: Let me do some digging

danielskovli: Agreed, I also can't find a way to trigger an Ansattporten login for an Altinn form. I guess the "Altinn Portal login" is a pseudo-Ansattporten type system, where you get an identity through ID-porten (or a loopback to Altinn self-registered users), after which you're presented with the selector for who you want to represent.

Straight up Ansattporten looks virtually the same, so its pretty confusing. Test a login here if you're interested: https://onboarding.test.maskinporten.no

(possibly worth noting that the org selector for Ansattporten is called "altinn orgvelger", so I guess that really seals the space-time loop)

src/Altinn.App.Core/Infrastructure/Clients/Authentication/AuthenticationClient.cs
martinothamar: Wasn't confident I could match the current behavior exactly

src/Altinn.App.Core/Internal/Auth/IAuthenticationContext.cs
martinothamar: There is no way currently to login using Ansattporten

PR #887, opened 2024-11-06
Use presence of org in user claims for conditionally setting read status on instance create

chrrust:
## Description
Uses the presence of org in user claim to conditionally set created instance to read or unread. The problem this tries to solve is when the instance is created by a different org than owns the instance. This is the case for Digital Gravferdsmelding. This project uses Digdir as Maskinporten-tenant, but the apps are owned by Statsforvalteren i Vestfold og Telemark (sfvt).

## Related Issue(s)
N/A

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -905,9 +905,9 @@ public class InstancesController : ControllerBase
 
     private void ConditionallySetReadStatus(Instance instance)
     {
-        string? orgClaimValue = User.GetOrg();
+        var isInstantiatedByOrg = User.GetOrg() is not null;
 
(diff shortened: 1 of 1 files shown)


ivarne (approved):
Looks good to me.

This is just about the read status, and the access for cross service owner writes is still regulated by policy.xml

martinothamar (commented):
Hmmm I'm not sure I'm comfortable with a behavior change like this. I don't know that no one else relies on the current behavior (or will have their internal workflow impacted in some way)

ivarne:
This is just about the read status of the instance (as shown in Altinn inbox). As long as it is an api call with a service owner token, it should never be marked as read.

chrrust:
Any updates on the review of this?

PR #888, opened 2024-11-06
POST instance: don't attempt to parse instance from body when it's empty or not json

ivarne:
Reported by frontend, causing issues in tests.

## Verification
- [ ] All tests run green

## Documentation


martinothamar (approved):
We should probably have a test for this as well, maybe create an issue for it

ivarne:
> We should probably have a test for this as well, maybe create an issue for it

Very good idea. Content type `.Contains("application/json")` works much better when the actual content type is `application/json;charset=utf8`

martinothamar (approved):
Double approve 😄

Issue #851, opened 2024-10-18
Implement Maskinporten -> Altinn token exchange in built-in Maskinporten abstractions


martinothamar:
We need to implement token exchange

## Acceptance criteria

- [x] It is clear which kind of token you receive
- [x] It is clear what the expiry of the token is

Issue #849, opened 2024-10-18
Implement internal variant of `IMaskinportenClient`


martinothamar:
We are going to end up with running 2 variants of the `IMaskinportenClient` - 1 using the custom client/scopes defined by the app owner in Studio, and 1 internal client with some built in scopes for operations that the app needs to do. Related issue: https://github.com/Altinn/app-lib-dotnet/issues/813

## Acceptance criteria

- [x] It is possible to request both the public and internal variant from DI
- [x] Same test coverage for both clients
- [x] Internal logic should be mostly the same (apart from configuration, since configuration should be different)
- [x] Telemetry makes it clear which client is making the requests

RonnyB71:
Closing as this will be fixed with an internal authentiaction

danielskovli:
Just for posterity: We ended up keeping this internal variant of the MaskinportenClient after all. The specific use-case @RonnyB71  was referring to will most likely be solved with an internal token from the authentication api.

However, we felt that the app itself may very well need its own way to authenticate with Maskinporten in the (near) future, so this internal client implementation lives on.

Issue #832, opened 2024-10-10
Implement metrics for Correspondence use


martinothamar:
* Number of messages
* Latency to open message by receiver?
* Error rates (creation, delivery)

danielskovli:
We're currently tracking number of correspondence orders and their immediate status, but nothing else. We may need to implement an event subscription if we want more details than that. I'm unsure of the available events, but at least `CorrespondencePublished` and `CorrespondencePublishFailed` are available.

danielskovli:
Correction, all available events are listed here: https://docs.altinn.studio/correspondence/getting-started/developer-guides/events/#event-subcription

martinothamar:
Speaking of docs, we should update this: https://docs.altinn.studio/altinn-studio/reference/monitoring/registry/metrics/

danielskovli:
We're way behind on docs atm, at least regarding Maskinporten and Correspondence. Though we're only currently in RC, so I guess there's no cause for panic 😬 

I'll make an issue to update the monitoring docs

ErlingHauan:
Metrics implemented

Issue #826, opened 2024-10-10
Implement the designed `ICorrespondenceClient`


martinothamar:
Implement `ICorrespondenceClient` as defined

## Acceptance criteria

- [x] The client implementation is a sealed internal class
~~It uses an internal variant of the `IMaskinportenClient`~~ 
- [x] It has robust error handling and telemetry
- [x] It can be used from within an Altinn app (`Altinn.App.Core`)

danielskovli:
The acceptance criteria for using an _internal_ variant of `IMaskinportenClient` I believe is no longer desired. In either case, that's not how the client has currently been implemented.

martinothamar:
agreed!

PR #897, opened 2024-11-08
Correspondence client

danielskovli:
## Description
This PR has perhaps become a bit large and unruly, but there were many interleaved concepts involved in this feature, which turned out to be difficult to separate.

### Maskinporten
The MaskinportenClient has been extended to allow for a second (internal use) instance, and the possiblity of exchanging Maskinporten tokens for Altinn tokens. Some updates have been made to the data models which *may* be considered breaking changes.

I have also renamed `HttpClientBuilderExtensions.UseMaskinportenAuthorization`, but forwarded the legacy method to the new one. It's not clear to me if this is necessary or not -- it's extremely unlikely that anyone has started using this product yet.

### Correspondence
The CorrespondenceClient and all associated logic/models/functionality is brand new. The implementation of this ecosystem is the main bulk of the PR.

The naming conventions for all things Correspondence could use a second pair of eyes. Perhaps especially the CorrespondenceRequestBuilder pattern naming.

### Telemetry
The Telemetry class has been extended with a partial class for Correspondence, and CorrespondenceClient implements `StartActivity` and request counting. This does **not** cover all objectives in https://github.com/Altinn/app-lib-dotnet/issues/832, so more work is likely needed in this area.

### Misc
Certain data that has traditionally been "stringly typed" now has their own struct types: `AccessToken`, `OrganisationNumber` and `LanguageCode`. Implicit operators and JSON converters have been added where applicable for current use.

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/832
- https://github.com/Altinn/app-lib-dotnet/issues/825
- https://github.com/Altinn/app-lib-dotnet/issues/826
- https://github.com/Altinn/app-lib-dotnet/issues/849
- https://github.com/Altinn/app-lib-dotnet/issues/851

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceAttachmentBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+/// <summary>
+/// Builder factory for creating <see cref="CorrespondenceAttachment"/> objects
+/// </summary>
+public class CorrespondenceAttachmentBuilder : CorrespondenceBuilderBase, ICorrespondenceAttachmentBuilder
+{
+    private string? _filename;
+    private string? _name;
+    private string? _sendersReference;
+    private string? _dataType;
+    private Stream? _data;
```
martinothamar: Curious to hear @ivarne thoughts on the type being `Stream` here. This field is likely to be populated from binary data elements in the app right? In `IDataClient` we return `Stream` while in `IInstanceDataAccessor` we return `ReadOnlyMemory<byte>`

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceAttachmentBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+    public ICorrespondenceAttachmentBuilder WithDataLocationType(CorrespondenceDataLocationType dataLocationType)
+    {
+        _dataLocationType = dataLocationType;
+        return this;
+    }
+
+    /// <inheritdoc/>
+    public CorrespondenceAttachment Build()
+    {
+        NotNullOrEmpty(_filename, "Filename is required");
```
martinothamar: I wonder if some of these fields should also be validated in the setters? Like `WithFilename("")` will not throw, but it will throw during build

src/Altinn.App.Core/Features/Correspondence/Models/CorrespondenceRequest.cs
```diff
@@ -0,0 +1,238 @@
...
+            var hasDuplicateFilenames = attachments
+                .GroupBy(x => x.Filename.ToLowerInvariant())
+                .Where(x => x.Count() > 1)
+                .Select(x => x.ToList());
+
+            foreach (var duplicates in hasDuplicateFilenames)
+            {
+                for (int i = 0; i < duplicates.Count; i++)
+                {
+                    duplicates[i].FilenameClashUniqueId = i + 1;
```
martinothamar: Is it possible to construct a usage pattern where instances within `duplicates` are shared across threads? Like building a request and then creating copies of it with different recipients but the same attachments or something and then sending them in parallel? Or even just sequentially that would have a strange effect I think

I couldn't find a test for the unique file name logic

src/Altinn.App.Core/Features/Correspondence/Models/SendCorrespondencePayload.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    // TODO: `WithRequest` overloads for built-in authz schemes here
+
+    /// <summary>
+    /// The payload contains an <see cref="ICorrespondenceRequestBuilder"/> builder instance
+    /// </summary>
+    /// <param name="correspondenceRequestBuilder">The correspondence request builder to send</param>
+    /// <param name="accessTokenFactory">Access token factory delegate (e.g. <see cref="MaskinportenClient.GetAltinnExchangedToken"/>)</param>
+    public static SendCorrespondencePayload WithBuilder(
```
martinothamar: `WithRequest` and `WithBuilder` in a simple class like this feels kind of verbose to me. Maybe it could be just `static SendCorrespondencePayload Create(CorrespondenceRequest correspondenceRequest)` (with an additional overload for the builder, same name) and `SendCorrespondencePayload WithCustomAuth(...)` or something?

src/Altinn.App.Core/Features/Maskinporten/IMaskinportenClient.cs
```diff
@@ -12,21 +12,46 @@ public interface IMaskinportenClient
...
     /// <exception cref="Maskinporten.Exceptions.MaskinportenAuthenticationException">
-    /// Authentication failed. This could be caused by an authentication/authorization issue or a myriad of other circumstances.
+    /// Authentication failed. This could be caused by an authentication/authorisation issue or a myriad of other circumstances.
     /// </exception>
     /// <exception cref="Maskinporten.Exceptions.MaskinportenConfigurationException">
     /// The Maskinporten configuration is incomplete or invalid. Very possibly because of a missing or corrupt maskinporten-settings.json file.
     /// </exception>
     /// <exception cref="Maskinporten.Exceptions.MaskinportenTokenExpiredException">The token received from Maskinporten has already expired.</exception>
-    public Task<MaskinportenTokenResponse> GetAccessToken(
+    public Task<JwtBearerToken> GetAccessToken(
```
martinothamar: It looks strange to me to encode JWT Bearer in the name of this type, since this is based on convention. There's nothing stopping someone from implementing an API accepting this token in some other random header, right? And I'm not sure if it's useful information for the consumer either. The developer still has to check the API documentation on where they expect the access token to be

src/Altinn.App.Core/Features/Maskinporten/Models/JwtBearerToken.cs
```diff
@@ -0,0 +1,60 @@
+using System.ComponentModel;
+using Altinn.App.Core.Models;
+
+namespace Altinn.App.Core.Features.Maskinporten.Models;
+
+/// <summary>
+/// An object that contains a JWT <see cref="AccessToken"/> member
+/// </summary>
+public interface IAccessToken
```
martinothamar: Do we need an interface for this? Seems kind of complicated to me to have 3 levels of types to really represent 1 object, which is the JWT token

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceAttachmentBuilder.cs
danielskovli: I have no preference on the interim storage for the data stream here. The builder and the record it targets both use `Stream` internally and serialize to `StreamContent`. 

Happy to modify `CorrespondenceAttachment`as required and add whatever builder overloads seem reasonable.

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceAttachmentBuilder.cs
danielskovli: Hmm, good point. We're in a similar-ish situation if people instantiate `CorrespondenceAttachment` directly: `Filename = string.Empty` is fine in the model itself, but the downstream API will throw a 400-BadRequest.

I made the `NotNull...` methods purely to tell the compiler that things were under control, but at some point I must have felt the need to check for empty strings as well.

So I'm throwing it back to you here. Preference on approach based on my comments above?

src/Altinn.App.Core/Features/Correspondence/Models/CorrespondenceRequest.cs
danielskovli: Good point on the testing of this, afaik there is zero coverage on that. Like we discussed briefly the other day, this particular implementation still doesn't make me completely at ease.

I remember doing it this way to avoid pushing too much data from the top-level `.Serialize()` method down to the attachment object, but I think perhaps this would be the correct approach. The serializer deals out new filenames directly to nested serialize models as required, but that state never gets stored in the records themselves.

src/Altinn.App.Core/Features/Correspondence/Models/SendCorrespondencePayload.cs
danielskovli: `Create()` with overloads make sense, I like it! 

I originally had that union-like pattern with nested classes, so I probably got a bit blind on the naming after abandoning that approach.

As for how to most elegantly deal with the myriad of auth sources, perhaps we should discuss that a bit further?

Using a delegate like I'm doing now makes sense in a general way, but perhaps turns this into an `Create(...)` overload-hell if we want to start auto-magically supporting lots more options.

And I'm not sure if we actually need to support anything besides a Bearer scheme, but we've toyed with the idea of allowing callers to configure the `HttpRequestMessage` directly.

I like the strongly typed `AccessToken` implementation, but I'm not entirely sure it covers us for all the foreseeable scenarios?

src/Altinn.App.Core/Features/Maskinporten/IMaskinportenClient.cs
danielskovli: 👍

src/Altinn.App.Core/Features/Maskinporten/Models/JwtBearerToken.cs
danielskovli: So the dilemma here is that Maskinporten has its own domain model, which is returned when we call their API. That makes sense.

Then we need an internal domain model for the Altinn Authentication call, which only returns a JWT string. 

I was reluctant to make these dependent on eachother, since they are not really related besides both offering a JWT at the end of the day. Hence, the `IAccessToken` interface.

That said, as I'm reading the code now it doesn't seem like that interface is required anymore. It's only used by `MaskinportenDelegatingHandler` and was orignally needed because `MaskinportenClient.GetAccessToken()` and `MaskinportenClient.GetAltinnExchangedToken()` had no overlap. But now they are returning the same type, which makes the interface redundant.

Just perhaps needs a better name, as per another comment 🙂

src/Altinn.App.Api/Extensions/HttpClientBuilderExtensions.cs
```diff
@@ -10,25 +11,72 @@ public static class HttpClientBuilderExtensions
...
+        var factory = ActivatorUtilities.CreateFactory<MaskinportenDelegatingHandler>(
+            [typeof(TokenAuthorities), typeof(IEnumerable<string>),]
+        );
+        return builder.AddHttpMessageHandler(provider => factory(provider, [authorities, scopes]));
+    }
+
+    // TODO: Do we need this? Has anyone actually used this method?
+    /// <inheritdoc cref="UseMaskinportenAuthorisation"/>
+    [Obsolete("UseMaskinportenAuthorization is deprecated, use UseMaskinportenAuthorisation instead.")]
+    public static IHttpClientBuilder UseMaskinportenAuthorization(
```
danielskovli: Any thoughts on this legacy hook @martinothamar? I renamed the original extension method from `...Authorization` to `...Authorisation`, and thought originally that it was nice to offer this fallback for compatibility.

src/Altinn.App.Core/Features/Correspondence/Models/CorrespondenceContent.cs
```diff
@@ -0,0 +1,45 @@
...
+    public required string Body { get; init; }
+
+    /// <summary>
+    /// File attachments to associate with this correspondence
+    /// </summary>
+    public IReadOnlyList<CorrespondenceAttachment>? Attachments { get; init; }
+
+    // TODO: Should this be internal?
+    /// <inheritdoc />
+    public void Serialize(MultipartFormDataContent content)
```
danielskovli: Two questions on this and the other serializers @martinothamar:

- As per our naming convention, this should probably be named `Serialise`. But for whatever reason I liked the americanized spelling way more, so I stuck with it. But its not really consistent with the change I did on `UseMaskinportenAuthorisation`
- These methods are currently public, which I think makes sense. But at the same time it would be extremely strange if the correspondence models got used in a way that CorrespondenceClient wasn't involved and the caller had to do their own serialization

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceAttachmentBuilder.cs
ivarne: In my current understanding of C# I strongly prefer `byte[]` or `ReadOnlyMemory<byte>` to `Stream`.

`Stream` is really awkward to work with, and most likely you end up needing multiple copies in memory to read muliple times. If we really have a case where it makes sense to process bytes in a streaming fashion we should use IPipeReader.

src/Altinn.App.Core/Features/Correspondence/Models/CorrespondenceContent.cs
ivarne: Picking a different dialect of English than C# for our C# code is a great way to get discussions like this.

ivarne:
I'm reading and trying to understand, but I have to ask. What does the builder pattern provide that we can't get by just having

```C#
public interface ICorrespondanceClient
{
    Task SendCorrespondance(CorrespondanceRequest request, CancellationToken ct = default);
}
```
that can be used as follows
```
await _correspondanceClient. SendCorrespondance(new()
{
    Sender = ...
    SenderReference = ...
    RequestedPublishTime = ...
    Recipients = [
    new(){...}
    ],
    ...
});
```
It is a lot of code and docstrings that we have to review for typos to get a builder pattern that we don't use for anything else.

Also 4 new exception types! Should I go through my `InstanceDataUnitOfWork` to add a different type for each `InvalidOperationException`?

ivarne:
And if we first are using a builder pattern. Would't it be expected that it makes immutable updates, instead of mutating the root?

```
var builder = new Builder().WithSender(---).WithSenderReference(...)

var message1 = builder.WithRecipient(...);
var message2 = builder.WithRecipient(...);
await client.Send(message1.Build());
await client.Send(message2.Build());
```

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceAttachmentBuilder.cs
martinothamar: `ReadOnlyMemory<byte>` sounds good to me considering `IInstanceDataAccessor` uses it and it seems like that's likely where app devs will be getting their input from

src/Altinn.App.Core/Features/Correspondence/Models/CorrespondenceContent.cs
martinothamar: Regarding your second point I also think it's strange for people to use our serialisation code direclty. Making those internal sounds good to me 👍

martinothamar:
Builder pattern lets us have more/earlier validation code, and let's the caller split up initialization and have logic inbetween building up the request. So we've discussed this being beneficial when

* The resulting model is large
* Some props being required, some not

I remember we also discussed immutability or not, looking at some of the core abstractions in .NET like DI it seemed like the mutable approach (returning `this`) was the most common

Re: exceptions, I can only think of benefits (callers being able to catch exceptions by concrete type or the whole category of `CorrespondenceException` to do custom error handling).

In general I would say the cost of maining code liks this (models and exceptions) is cheap, so I think it's worth it given the benefits

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceAttachmentBuilder.cs
danielskovli: Just FYI, I ended up going with `ReadOnlyMemory<byte>` as the internal storage for this data, which has a nice implicit conversion from `byte[]`. 

The serialisation now uses `ReadOnlyMemoryContent` instead of `StreamContent`.

danielskovli:
@ivarne I don't have much to add in addition to what @martinothamar wrote above. 

I personally think that both the builder pattern and the exception strategy make sense. At the risk of opening up another can of worms, I feel obliged to mention that I implemented exactly the same exception concept with `MaskinportenClient`.

The main reason for deciding on a builder, like Martin mentioned, is that the CorrespondenceRequest model is quite complex. A builder lets you split the logic into chunks, while also holding your hand a bit regarding required and optional properties.

It's well worth mentioning that sending through a `CorrespondenceRequest` instance directly, the way you outlined, also works well. So in effect I'm trying to offer the callers both options.

The concern about maintaining lots of boilerplate and XML comments isn't necessarily something I think we should be too worried about.

Builder mutation or not isn't an area I have a strong preference about, but the way it's currently implemented here gels pretty well with the rest of the .NET ecosystem from my experience 🤷‍♂️

src/Altinn.App.Core/Features/Correspondence/Models/CorrespondenceRequest.cs
```diff
@@ -0,0 +1,272 @@
...
+    internal static void SerializeAttachmentItems(
+        MultipartFormDataContent content,
+        IReadOnlyList<CorrespondenceAttachment>? attachments
+    )
+    {
+        if (IsEmptyCollection(attachments))
+            return;
+
+        // Ensure unique filenames
+        var overrides = new Dictionary<CorrespondenceAttachment, string>();
```
martinothamar: Since `CorrespondenceAttachment` is a record and it is being used as a key here, it is doing value equality across all the fields within. I wonder if this is actually a good case for reference equality since it is a class?

danielskovli: Good catch!

src/Altinn.App.Api/Extensions/HttpClientBuilderExtensions.cs
```diff
@@ -10,25 +11,60 @@ public static class HttpClientBuilderExtensions
...
-    /// <see cref="MaskinportenClient.GetAccessToken"/> directly and handling authorization details manually.
+    /// If your target API does <em>not</em> use this authorisation scheme, you should consider implementing
+    /// <see cref="MaskinportenClient.GetAccessToken"/> directly and handling the specifics manually.
     /// </para>
     /// </summary>
     /// <param name="builder">The Http client builder</param>
     /// <param name="scope">The scope to claim authorization for with Maskinporten</param>
     /// <param name="additionalScopes">Additional scopes as required</param>
-    public static IHttpClientBuilder UseMaskinportenAuthorization(
+    public static IHttpClientBuilder UseMaskinportenAuthorisation(
```
martinothamar, line 25: Just leaving as a paper trail, we decided a breaking change is OK here as we could't find usages of this in apps yet

src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -160,22 +165,61 @@ string configSectionPath
...
+        );
 
         return services;
     }
 
+    /// <summary>
+    /// Adds a <see cref="ICorrespondenceClient"/> service to the service collection.
+    /// </summary>
+    /// <param name="services">The service collection</param>
+    private static IServiceCollection AddCorrespondenceClient(this IServiceCollection services)
```
martinothamar: Should we start putting these types of extension methods closer to the implementation? Usually it's the implementation that has the most say on the lifetime of the service

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceAttachmentBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+    private bool? _isEncrypted;
+    private CorrespondenceDataLocationType _dataLocationType =
+        CorrespondenceDataLocationType.ExistingCorrespondenceAttachment;
+
+    private CorrespondenceAttachmentBuilder() { }
+
+    /// <summary>
+    /// Creates a new <see cref="CorrespondenceAttachmentBuilder"/> instance
+    /// </summary>
+    public static ICorrespondenceAttachmentBuilderFilename Create() => new CorrespondenceAttachmentBuilder();
```
martinothamar, line 24: Sanity-checking, why do we have these create methods when they are empty?

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceRequestBuilder.cs
```diff
@@ -0,0 +1,299 @@
...
+    /// <summary>
+    /// Because of the interface-chaining in this builder, some properties are guaranteed to be non-null.
+    /// But the compiler doesn't trust that, so we add this check where needed.
+    ///
+    /// Additionally this method checks for empty strings and empty data allocations.
+    /// </summary>
+    /// <param name="value">The value to assert</param>
+    /// <param name="errorMessage">The error message to throw, if the value was null</param>
+    /// <exception cref="CorrespondenceValueException"></exception>
+    internal static void NotNullOrEmpty([NotNull] object? value, string? errorMessage = null)
```
martinothamar: Can we share this function without having `CorrespondenceBuilderBase` as a base class? Is it used for something else? Reason I'm asking is because the symbol is public and so would be part of the surface we have to maintain/can't change

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceRequestBuilder.cs
```diff
@@ -0,0 +1,299 @@
...
+    {
+        return WithRecipients([recipient]);
+    }
+
+    /// <inheritdoc/>
+    public ICorrespondenceRequestBuilderAllowSystemDeleteAfter WithRecipients(
+        IReadOnlyList<OrganisationOrPersonIdentifier> recipients
+    )
+    {
+        _recipients = [.. _recipients ?? [], .. recipients];
```
martinothamar: Sanity-checking again, since we use a mutable model here (for the builder in general), why do we not just add to the existing list?

src/Altinn.App.Core/Features/Correspondence/Builder/ICorrespondenceNotificationBuilder.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// <param name="notificationTemplate">The notification template</param>
+    ICorrespondenceNotificationBuilder WithNotificationTemplate(
+        CorrespondenceNotificationTemplate notificationTemplate
+    );
+}
+
+/// <summary>
+/// Indicates that the <see cref="CorrespondenceNotificationBuilder"/> instance has completed all required steps and can proceed to <see cref="CorrespondenceNotificationBuilder.Build"/>
+/// </summary>
+public interface ICorrespondenceNotificationBuilder : ICorrespondenceNotificationBuilderTemplate
```
martinothamar, line 22: One thing I came to think of now, a bit late in the game, but are we using the interfaces for something? I.e. could the builder all just be records that don't implement interfaces? Reason I ask is because usually we have interfaces because we want to hide implementation details and have the expected behavior defined independently. In this case the consumers are using the classes directly

src/Altinn.App.Core/Features/Correspondence/CorrespondenceClient.cs
```diff
@@ -0,0 +1,241 @@
...
+        return $"{baseUri}/{relativePath.TrimStart('/')}";
+    }
+
+    private async Task<TContent> HandleServerCommunication<TContent>(
+        HttpRequestMessage request,
+        CancellationToken cancellationToken
+    )
+    {
+        using HttpClient client = _httpClientFactory.CreateClient();
+        HttpResponseMessage response = await client.SendAsync(request, cancellationToken);
```
martinothamar: dispose the response?

src/Altinn.App.Core/Features/Correspondence/CorrespondenceClient.cs
```diff
@@ -0,0 +1,241 @@
...
+            var problemDetails = GetProblemDetails(responseBody);
+            throw new CorrespondenceRequestException(
+                $"Correspondence request failed with status {response.StatusCode}: {problemDetails?.Detail}",
+                problemDetails,
+                response.StatusCode,
+                responseBody
+            );
+        }
+
+        _logger.LogDebug("Correspondence request succeeded: {Response}", responseBody);
```
martinothamar, line 221: Use of debug logging should also be something we should discuss in the team, as in how often and how we do it (there are more perfomant APIs for this as well)

src/Altinn.App.Core/Features/Correspondence/Models/CorrespondenceNotificationOrderResponse.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// The senders reference of the notification
+    /// </summary>
+    [JsonPropertyName("sendersReference")]
+    public string? SendersReference { get; set; }
+
+    /// <summary>
+    /// The requested send time of the notification
+    /// </summary>
+    [JsonPropertyName("requestedSendTime")]
+    public DateTime RequestedSendTime { get; set; }
```
martinothamar: Can we be consistent in the use of datetime types? I believe when serialized we end up with the same format regardless

src/Altinn.App.Core/Features/Correspondence/Models/SendCorrespondenceResponse.cs
```diff
@@ -0,0 +1,21 @@
...
+/// <summary>
+/// Response after a successful <see cref="CorrespondenceClient.Send"/> request
+/// </summary>
+public sealed record SendCorrespondenceResponse
+{
+    /// <summary>
+    /// The correspondences that were processed
+    /// </summary>
+    [JsonPropertyName("correspondences")]
+    public required List<CorrespondenceDetailsResponse> Correspondences { get; init; }
```
martinothamar: For responses, which generally are immutable, maybe we should use `IReadOnlyList` or something instead?

src/Altinn.App.Core/Features/Maskinporten/Models/TokenWrapper.cs
```diff
@@ -0,0 +1,49 @@
+using System.ComponentModel;
+using Altinn.App.Core.Models;
+
+namespace Altinn.App.Core.Features.Maskinporten.Models;
+
+/// <summary>
+/// JWT wrapper which includes the <see cref="AccessToken"/> and relevant metadata
+/// </summary>
+[ImmutableObject(true)] // `ImmutableObject` prevents serialization with HybridCache
+public sealed record TokenWrapper
```
martinothamar: Hmm I'm a bit unsure still about this structure. Returning a wrapper from a public interface. What if `AccessToken` itself had these properties? They are all encoded inside the JWT anyway right?

src/Altinn.App.Core/Models/AccessToken.cs
```diff
@@ -0,0 +1,109 @@
...
+/// Represents an OAuth 2.0 access token in JWT format
+/// </summary>
+public readonly partial struct AccessToken : IEquatable<AccessToken>
+{
+    private readonly string _accessToken;
+
+    /// <summary>
+    /// Gets the access token as a string
+    /// </summary>
+    public string Get()
```
martinothamar: Could this be a `Value` property instead? Feels like that better fits the convention

test/Altinn.App.Api.Tests/Altinn.App.Api.Tests.csproj
```diff
@@ -30,6 +30,10 @@
     <ProjectReference Include="..\Altinn.App.Common.Tests\Altinn.App.Common.Tests.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Altinn.App.Core.Tests" />
```
martinothamar, line 34: Should Core tests depend on Api tests? If we have code that needs to be shared between both, maybe that's an argument for putting it in Altinn.App.Common.Tests

martinothamar (commented):
Made a first pass

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceAttachmentBuilder.cs
danielskovli: If I understand the question correctly, it's because of the returned interface. We need to new-up a builder instance, but only return a specific interface that says "I now need the filename, please and thank you".

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceRequestBuilder.cs
danielskovli: Absolutely! I was just hoping to avoid a helper class situation, but that can easily be done 🙂

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceRequestBuilder.cs
danielskovli: No reason that I can think of. This was just the pattern that stuck from the first POC we did a while back.

You prefer to init the list of required, then just add the new items?

src/Altinn.App.Core/Features/Correspondence/Models/CorrespondenceNotificationOrderResponse.cs
danielskovli: These types come from the Correspondence repo itself. It's possible that we can change them all to DateTimeOffset types, I haven't actually attempted that.

src/Altinn.App.Core/Features/Correspondence/Models/SendCorrespondenceResponse.cs
danielskovli: Agreed! This is the way it is just because of copy from altinn-correspondence

src/Altinn.App.Core/Features/Maskinporten/Models/TokenWrapper.cs
danielskovli: Feel like having a crack at something else here? I'm open to suggestions, but seeing as this was my second attempt, perhaps its time to hand over the reins 😄

src/Altinn.App.Core/Models/AccessToken.cs
danielskovli: If so, I think we should do the same for NationalIdentityNumber. I had them all as Get() since we used that for OrganisationNumber, but there's no reason why we can't change to Value for the types that don't need a formatting argument.

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceRequestBuilder.cs
martinothamar: Right, if we get the same behavior but better performance then it is tempting

test/Altinn.App.Api.Tests/Altinn.App.Api.Tests.csproj
danielskovli: I have no optinion on this. I just wanted access to some stuff across that boundary :)

src/Altinn.App.Core/Features/Correspondence/Models/CorrespondenceNotificationOrderResponse.cs
martinothamar: I hope that's possible, since in JSON this will be ISO8601 regardless of type

src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
danielskovli: Good question. I've already started putting the models, delegates, and other relevant things in the same namespace as the feature, so perhaps extensions also belong there 🤷

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceRequestBuilder.cs
martinothamar: maybe you've already seen it, but in this BCL example it's part of the exception type: https://github.com/dotnet/runtime/blob/6be24fd37e7d9f04c7fa903b8b6912c3eafe7198/src/libraries/System.Private.CoreLib/src/System/ArgumentException.cs#L104
apart from that `ThrowHelper` is also pretty common

src/Altinn.App.Core/Features/Correspondence/Builder/ICorrespondenceNotificationBuilder.cs
martinothamar: Discussed offline, reason being we have some guardrails in the form of interfaces to ensure required fields are input

src/Altinn.App.Core/Features/Maskinporten/Models/TokenWrapper.cs
martinothamar: heres a proposal: https://github.com/Altinn/app-lib-dotnet/pull/915

danielskovli: Makes sense, I like it! The name is awkward, because the "t" in "jwt" stands for "token, but I don't have any other good suggestions.

And people walk around saying stuff like "pin number" and "atm machine" all the time, so I doubt anyone besides me will care 😅

src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs
danielskovli: I moved the internal Maskinporten extensions as well, seeing as I've already got my hands dirty over in that domain. These are only non-public methods, so changing our minds later wouldn't be too dramatic.

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceRequestBuilder.cs
danielskovli: Ended up with a `BuilderUtils` class. It`s not exposed publicly, so we are free to change that to whatever we want down the line.

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceRequestBuilder.cs
danielskovli: The exception to this rewrite was `CorrespondenceContent.Attachments`, which is stored as a `IReadOnlyList` in the model. Due to the nature of _not_ building this from the `CorrespondenceContentAttachmentBuilder`, but instead from the main `CorrespondenceRequestBuilder`, this particular list had to be instantiated on each modification (unless I tracked it locally in the builder, which felt messy): https://github.com/Altinn/app-lib-dotnet/pull/897/files#diff-115e9f775e8a2658fc9cd7c0afd11678d96d4496ccae4ed2b5373131b1cd33dbR291

test/Altinn.App.Api.Tests/Altinn.App.Api.Tests.csproj
danielskovli: I'm putting this in the too-hard basket for now: https://github.com/Altinn/app-lib-dotnet/issues/916

src/Altinn.App.Core/Features/Maskinporten/Models/TokenWrapper.cs
danielskovli: Merged

src/Altinn.App.Core/Features/Correspondence/CorrespondenceClient.cs
danielskovli: I've got no useful opinions on this. What's here now is what made sense to me during testing 🤷

src/Altinn.App.Core/Features/Correspondence/CorrespondenceClient.cs
danielskovli: Good catch!

src/Altinn.App.Core/Models/NationalIdentityNumberConverter.cs
```diff
@@ -0,0 +1,32 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Altinn.App.Core.Models;
+
+/// <summary>
+/// Json converter to transform between <see cref="string"/> and <see cref="NationalIdentityNumber"/>
+/// </summary>
+public class NationalIdentityNumberConverter : JsonConverter<NationalIdentityNumber>
```
martinothamar: I see that some of these are internal and some are public. We should probably standardize, and also naming should either include `Json` or not

martinothamar (approved):

src/Altinn.App.Core/Models/NationalIdentityNumberConverter.cs
danielskovli: I assume they have to be public to be used in a public class for de/serialization?

danielskovli: As for naming, I had intended for them all to contain `json` in the name. Thoughts?

martinothamar: Json in the name sounds good to me. Not sure if they need to be public, some of them are internal and I assume they work?

PR #904, opened 2024-11-11
Added support for displaying a text if the user is previewing the pdf

adamhaeger:
Added support for displaying a text if the user is previewing the pdf before the form is submitted.

Frontend PR is dependent on this PR: https://github.com/Altinn/app-frontend-react/pull/2710

## Description
Please see acceptance critereas here: https://github.com/Altinn/app-frontend-react/issues/1804

- Added isPreview parameter to `PdfService.GeneratePdf`
- Enabled PDF preview endpoint in production
- Added logic to display the preview text in the footer instead of the footer when previewing
- Added default preview text
- Added support for custom preview text

## Related Issue(s)
- #{1804}

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Core/Internal/Pdf/IPdfService.cs
```diff
@@ -22,5 +22,6 @@ public interface IPdfService
     /// <param name="instance">The instance details.</param>
     /// <param name="taskId">The task id for witch the pdf is generated</param>
     /// <param name="ct">Cancellation Token for when a request should be stopped before it's completed.</param>
-    Task<Stream> GeneratePdf(Instance instance, string taskId, CancellationToken ct);
+    /// <param name="isPreview">If set to true, this will show a text saying that this is a preview PDF.</param>
+    Task<Stream> GeneratePdf(Instance instance, string taskId, CancellationToken ct, Boolean? isPreview = false);
```
ivarne: CancellationToken conventionally comes last. Not sure what is most important. Keeping existing calls working, or following conventions.

The middle ground would be to have two methods (something like this)
```suggestion
    [Obsolete("Use the GeneratePdf method with the isPreview parameter")]
    Task<Stream> GeneratePdf(Instance instance, string taskId, CancellationToken ct);
    
    Task<Stream> GeneratePdf(Instance instance, string taskId, bool isPreview, CancellationToken ct);
```

At least use `bool` instead of `Boolean?` when `null` doesn't mean anything.

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
```diff
@@ -103,15 +108,15 @@ public async Task<Stream> GeneratePdf(Instance instance, string taskId, Cancella
...
 
     private async Task<Stream> GeneratePdfContent(
         Instance instance,
-        string taskId,
         string language,
-        TextResource? textResource,
-        CancellationToken ct
+        CancellationToken ct,
+        TextResource? textResource = null,
+        Boolean? isPreview = false
```
ivarne: CancellationToken last. No need for default values for compatibility with `private` methods

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
```diff
@@ -249,12 +264,39 @@ private static string GetFileName(Instance instance, TextResource? textResource)
...
+                textResourceElement.Id.Equals("pdfPreviewText", StringComparison.Ordinal)
+            );
+
+            if (titleText is not null)
+            {
+                return titleText.Value;
+            }
+        }
+
+        return null;
```
ivarne: Should the default `"Dette er en forhåndsvisning"` be set here instead of in the `GetPreviewFooter`?

test/Altinn.App.Api.Tests/Data/apps/tdd/task-action/config/texts/resource.nb.json
```diff
@@ -28,6 +28,10 @@
     {
       "id": "Side1.Button-Lj0cYG.title",
       "value": "Avvis"
+    },
+    {
+        "id": "PdfPreviewText",
+        "value": "Dette er en forhåndsvisning"
```
ivarne: This is an app we use for testing, not the default texts used everywhere. Also, the capitalisation is wrong. It should be `"pdfPreviewText"`

src/Altinn.App.Core/Internal/Pdf/IPdfService.cs
ivarne: For the implementation, just use a stub implementation for the obsolete method

```c#
    Task<Stream> GeneratePdf(Instance instance, string taskId, CancellationToken ct)
    {
        await GeneratePdf(instance, taskId, isPreview:false, ct);
    }
```

ivarne (commented):
A few code style issues. Looks like you also need to remove the test that ensures that this don't run in production.

src/Altinn.App.Core/Internal/Pdf/IPdfService.cs
adamhaeger: Well according to the spec its fine when you have optional parameters:

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068#special-cases

I would just go with the optional parameter, but if you guys prefer two methods I dont mind :)

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
adamhaeger: The default value is required to have the parameter optional isnt it?

test/Altinn.App.Api.Tests/Data/apps/tdd/task-action/config/texts/resource.nb.json
adamhaeger: RIght, then Im not sure how we set default texts?

ivarne: We don't have that functionality in the backend (yet).

ivarne: Actually we don't translate anything in the backend, so there was no need. Pdf footer kind of throws a wrench in that assumption.

adamhaeger: Right, guess I could submit the translated text in the request from FE?

adamhaeger: Though that would allow a user to set whatever text they want ofc. Like "Signed by Bjørn Dælie" 😆

adamhaeger: I think we should just set the text in bokmål for now right?

ivarne: No, just store the defaults (for all languages) inline in C# and let the app developer override through the use of `resource_{lang}.json`

A set of default translations can be implemented later if we need it.

ivarne: Note the race condition in comments here. You commented twice before my reply

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
```diff
@@ -122,7 +128,17 @@ CancellationToken ct
         Uri uri = BuildUri(baseUrl, pagePath, language);
 
         bool displayFooter = _pdfGeneratorSettings.DisplayFooter;
-        string? footerContent = displayFooter ? GetFooterContent(instance, textResource) : null;
+
+        string? footerContent = null;
+
+        if (isPreview == true)
```
martinothamar, line 134: Should we not check `displayFooter` when generating preview? I'd imagine that might pop us as unexpected behavior from users

martinothamar (approved):
LGTM

src/Altinn.App.Core/Internal/Pdf/IPdfService.cs
martinothamar: We decided to have 2 methods like you propose, but not obsoleting the original method as it is fine as is

src/Altinn.App.Core/Internal/Pdf/PdfService.cs
adamhaeger: This is according to spec: https://github.com/Altinn/app-frontend-react/issues/1804 :)

PR #905, opened 2024-11-12
Fix VerifyDataElementsUnchangedSincePreviousChanges

ivarne:
FormDataChange has CurrentBinaryData as nullable, (because for data processing we don't have the serialized value when it is still possible to change)

I opted for setting delete to an empty ReadOnlyMemory<byte> so that the null value are reserved for data processing.

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Core/Models/DataElementChanges.cs
```diff
@@ -122,8 +122,10 @@ public sealed class FormDataChange : DataElementChange
     /// The binary representation (for storage) of the data element after changes
     /// </summary>
     /// <remarks>
-    /// Not available for form data in data processing phase, because it can't reflect
+    /// null during the form data in data processing phase, because it can't reflect
```
martinothamar: What happened to this part of the comment?

martinothamar (approved):

src/Altinn.App.Core/Models/DataElementChanges.cs
ivarne: I will fix (and include the tests) soon.

I ran into an issue overriding environment so that this code runs under webapplicationfactory

PR #909, opened 2024-11-18
update builder to standard set in correspondance

HauklandJ:
## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Core/Features/Signing/SigningDelegationService.cs
```diff
@@ -26,25 +27,48 @@ CancellationToken ct
...
+                        .WithInstanceId(instance.Id)
+                        .WithDelegator(new Delegator { IdType = DelegationConst.Party, Id = "" })
+                        .WithRecipient(
+                            new Delegatee { IdType = DelegationConst.Party, Id = signeeContext.PartyId.ToString() }
+                        )
+                        .WithRights(
+                            [
+                                AccessRightBuilder
+                                    .Create()
+                                    .WithAction(DelegationConst.ActionId, ActionType.Read)
```
bjorntore: Do we need the first parameter on `WithAction` where you pass in `DelegationConst.ActionId`? Or can that we hardcoded within the method? Will it not always be the same for any action?

Same for Resource probably. Isn't the type for a Resource always `DelegationConst.Resource`?

HauklandJ: Resource may have Type = DelegationConst.Task, but it makes sense to set Resource as default with the possibility to override. Action can be simplified as suggested.

PR #910, opened 2024-11-18
Revert to using Newtonsoft.Json for parsing instance

ivarne:
There was reported inconsistent behaviour when deserializing enums, which would likely be fixed if we add System.Text.Json annotations to Storage.Interfaces

The change was introduced in https://github.com/Altinn/app-lib-dotnet/pull/853, so it makes sense to revert.
https://github.com/Altinn/app-lib-dotnet/commit/6bac8a1264211d29cf964a7195b372ff7ce1e044#diff-509e2ea0fedd1d6a1d2d1c84b39ad128f078c03b225ce39ef5b578614beccbadL1224-R1233

## Related Issue(s)
- reported privately by @chrrust on slack

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -1,6 +1,6 @@
 using System.Globalization;
 using System.Net;
-using System.Text.Json;
+using System.Text;
 using Altinn.App.Api.Extensions;
 using Altinn.App.Api.Helpers.Patch;
 using Altinn.App.Api.Helpers.RequestHandling;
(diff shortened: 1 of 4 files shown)


martinothamar (commented):
Would like @HauklandJ to review here. Do we use string serialization of enums in our API? (as in `IOptionsMonitor<JsonOptions>.Get(JsonSettingNames.AltinnApi)`), or was it the other way around

HauklandJ:
> Would like @HauklandJ to review here. Do we use string serialization of enums in our API? (as in `IOptionsMonitor<JsonOptions>.Get(JsonSettingNames.AltinnApi)`), or was it the other way around

We use number serlialization of enums.

ivarne:
> We use number serlialization of enums.

That is not entirely correct, because in lots of places we use `[JsonConverter(typeof(JsonStringEnumConverter))]`. In this case this PR preserves the behaviour from `v8.3.x` and earlier, so I think this should be merged before releasing `v8.5`

HauklandJ (approved):

PR #914, opened 2024-11-19
Make vulnerable transitive deps not an error when running in CI

ivarne:
Dotnet 9 (which seems to be installed on some of githubs workers) has enabled checks for transitive vulnerabilities. Blocking PR builds for unrelated vulnerabilities seems wrong to me.

This is also kind of a recommendation from microsoft
> If these warnings are causing restore to fail because you are using TreatWarningsAsErrors, you can add &lt;WarningsNotAsErrors&gt;NU1901;NU1902;NU1903;NU1904&lt;/WarningsNotAsErrors&gt; to allow these codes to remain as warnings.
https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1901-nu1904#solution

I'm somewhat surprised our renovate does not catch and fix these kind of issues.

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,7 @@
 <Project>
   <PropertyGroup Condition="'$(DOTNET_TREATWARNINGSASERRORS)' == 'true' OR '$(CI)' == 'true'">
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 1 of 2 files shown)


martinothamar (approved):
LGTM, maybe we should use `global.json` so that correct version of SDK is always used as well?

HauklandJ (approved):

PR #920, opened 2024-11-21
Fix issues exposed by dotnet 9 sdk

ivarne:
* Add direct reference to new versions of JWTCookieAuthentication's insecure depencencies
* Activate security scan of transitive depencecies on net8 (but not as errors in CI)
* Warn on IDE0052: Remove unread private member
* Warn on CA1859: Use concrete types when possible for improved performance (but not in test)
* Fix related code

## Verification
- [ ] All tests run green

## Documentation


martinothamar (approved):

ivarne:
Testing with localtest exposed that it fails because `http://localhost:5101/authentication/api/v1/openid/.well-known/openid-configuration` returns a json document where `frontchannel_logout_supported` and `frontchannel_logout_session_supported` was `null` instead of `false` or missing. This is fixed in https://github.com/Altinn/app-localtest/pull/125

martinothamar:
Since we're merging this now, we should probably post a message in the product channel that people need to pull latest localtest?

ivarne:
Yes, it’s just that #921 failed the publish

PR #930, opened 2024-11-27
Populate `UserContext.SocialSecurityNumber` in `UserHelper`

danielskovli:
## Description
The main purpose of this PR is modify the method `UserHelper.GetUserContext()` to populate the property `UserContext.SocialSecurityNumber`. 

However, I found the flow of that method to be a bit hard to follow, which is the reason for the refactoring. And because those changes potentially modifies the logic of the method, I added a test.

### Investigation needed?
~I disagreed with the original logic for assignment of `UserProfile.Party`, so I changed it. At least with our test data, `UserProfile.PartyId` was not a good indicator to tell if `UserProfile.Party` contained valid information. The check now inspects `UserProfile.Party?.PartyId` instead.~

~If someone actually wants the original value, this is stored in `UserProfile.UserParty` anyway.~

~As for whether this impacts existing usage, it would be nice to get a second pair of eyes on the situation. Comparing `app-lib-dotnet` test data with `app-localtest` (for instance `1337.json`), it appears that the change is beneficial to incomplete data sets (`app-lib-dotnet`, empty `party` definition) and will not affect behaviour for complete data sets (`app-localtest`).~

After some back and forth on Slack, we have decided to leave the original `Party` assignment logic intact. Too many unknowns associated with this modification.

## Motivation
A question was recently asked about limiting app access to a list of known national identity numbers (social security numbers): https://digdir.slack.com/archives/C079ZFUSFMW/p1732624713911909

The general recommendation was to solve this in the app code as opposed to XACML. Getting hold of the information seemed easiest with the help of `UserHelper`, but for some reason this did not populate the SSN field.

Another baffling question is why the national identity number is part of the `Party` definition and not, say, `UserProfile`. But I'm sure that makes sense to someone else.

## Verification
- [x] All tests run green

## Documentation


test/Altinn.App.Api.Tests/TestUtils/AppBuilder.cs
```diff
@@ -9,7 +9,8 @@ public static class AppBuilder
     public static WebApplication Build(
         WebApplicationBuilder? builder = default,
         IEnumerable<KeyValuePair<string, string?>>? configData = default,
-        Action<IServiceCollection>? registerCustomAppServices = default
+        Action<IServiceCollection>? preRegisterCustomAppServices = default,
+        Action<IServiceCollection>? postRegisterCustomAppServices = default
```
danielskovli: We needed a post-registration hook here in order to override app default implementations

src/Altinn.App.Core/Helpers/UserHelper.cs
```diff
@@ -81,25 +69,40 @@ public async Task<UserContext> GetUserContext(HttpContext context)
...
-            userContext.PartyId = Convert.ToInt32(
-                context.Request.Cookies[_settings.GetAltinnPartyCookieName],
-                CultureInfo.InvariantCulture
-            );
-        }
+        userContext.PartyId = cookieValue is not null
+            ? Convert.ToInt32(cookieValue, CultureInfo.InvariantCulture)
+            : userContext.PartyId;
 
-        if (userContext.PartyId == userProfile.PartyId)
```
danielskovli, line 94: This is the check that was altered, re: PR description

src/Altinn.App.Core/Helpers/UserHelper.cs
```diff
@@ -81,25 +69,40 @@ public async Task<UserContext> GetUserContext(HttpContext context)
...
+            return claimWrapper.Value;
         }
-        else
+
+        public static implicit operator int(ClaimWrapper claimWrapper)
         {
-            userContext.Party = await _altinnPartyClientService.GetParty(userContext.PartyId);
+            return string.IsNullOrEmpty(claimWrapper.Value)
+                ? default
+                : Convert.ToInt32(claimWrapper.Value, CultureInfo.InvariantCulture);
```
martinothamar: When the underlying operation is fallible, I think it's nicer to have the conversion be explicit. In other words I'm not sure `ClaimWrapper` is worth it (I understand that this changes nothing functionally, but it puts more distance between the knowledge of the value and the fallible operation)

src/Altinn.App.Core/Helpers/UserHelper.cs
martinothamar: Why is this changing? I'm not sure I understood what the difference was between these two fields. Have you looked into the Profile API or something?

test/Altinn.App.Api.Tests/TestUtils/AppBuilder.cs
martinothamar: how about keeping `registerCustomAppServices` (would be nice if this mapped clearly to the corresponding function that is part of the app template, which is why this fixture was originally created), and adding `postAddAltinnAppServices`/`overrideAltinnAppServices` instead?

martinothamar (approved):
I like it, though I think over time these abstractions will be replaced with this: https://github.com/Altinn/app-lib-dotnet/pull/880

test/Altinn.App.Api.Tests/TestUtils/AppBuilder.cs
danielskovli: Makes sense to me, I like the 'override' approach 🙌

src/Altinn.App.Core/Helpers/UserHelper.cs
danielskovli: How about something like `ClaimWrapper.TryParseInt()` or something similar? If we aren't using an intermediary object, we have to go back to checking for nulls (eg. the value wasnt found), which means something along the lines of the original structure.

I mean, it clearly worked, so perhaps this was some unnecessary abstraction on my part.

src/Altinn.App.Core/Helpers/UserHelper.cs
danielskovli: I probably explained that party poorly. The only reason I've changed it is based on the test data we have available. Take for instance the user `1337.json`:

```
{
  "UserId": 1337,
  "UserName": "SophieDDG",
  "PhoneNumber": "90001337",
  "Email": "1337@altinnstudiotestusers.com",
  "PartyId": 501337,
  "Party": {},
  "UserType": 1,
  "ProfileSettingPreference": {
    "Language": "nn",
    "doNotPromptForParty": true
  }
}
```

If `PartyId` for this user matches whatever the `urn:altinn:partyid` (or `AltinnPartyId` cookie) claims, the property `UserContext.Party` will be set to an empty object instead of being looked up with the `IAltinnPartyClient`.

I'm sure this is some form of optimisation, and perhaps an empty `Party` object does not appear in the willd, but in our test environment it definitely produces a faulty logic.

So my proposed solution was to peek inside to the `Party.PartyId` property instead of relying on `PartyId` at root level.

martinothamar: I assume you're talking about localtest, but what is the behavior of the Profile API in other environments? Should `PartyId ` and `Party.PartyId` always be the same? Do we know that for a fact? I think we should be clear about this because bugs introduced here will have very broad impact

martinothamar: For example, I don't know why `Party` isn't filled in in the example you give above, like what are all the various permutations we might see of this object IRL

danielskovli: Good questions. I don't have the answers here, so perhaps we either need a more thorough investigation, or perhaps I revert the logic and update the test data file instead 🤷

src/Altinn.App.Core/Helpers/UserHelper.cs
martinothamar: Can't we have the same behavior just without the intermediary? E.g. a local function

```csharp
private static T GetClaimOrDefault(string claimType, T defaultValue);
```

Then the caller can decide what the expectation is from the claim. 
Also, not a big deal considering the number of claims, but with this approach we are going from 1 full loop over the claims to 1 loop over the claims per type we want to find, so it should be a little less efficient

src/Altinn.App.Core/Helpers/UserHelper.cs
danielskovli: As discussed externally, I've rolled back the `Party` assignment to its original logic

src/Altinn.App.Core/Helpers/UserHelper.cs
```diff
@@ -50,28 +49,33 @@ public UserHelper(
...
-        foreach (Claim claim in context.User.Claims)
+        UserContext userContext = new()
         {
-            if (claim.Type.Equals(AltinnCoreClaimTypes.UserName, StringComparison.Ordinal))
-            {
-                userContext.UserName = claim.Value;
-            }
-            else if (claim.Type.Equals(AltinnCoreClaimTypes.UserId, StringComparison.Ordinal))
+            User = context.User,
+            UserName = tokenClaims[AltinnCoreClaimTypes.UserName],
```
martinothamar: this will throw `KeyNotFoundException` if the key doesn't exist, there is `GetValueOrDefault` which does what you want I think

danielskovli: Gah, of course. Good eyes 👀

src/Altinn.App.Core/Helpers/UserHelper.cs
```diff
@@ -50,55 +49,50 @@ public UserHelper(
...
-        else
-        {
+        else if (userContext.PartyId > 0)
             userContext.Party = await _altinnPartyClientService.GetParty(userContext.PartyId);
-        }
+
+        userContext.UserParty = userProfile.Party;
+
+        userContext.SocialSecurityNumber =
+            userContext.Party?.SSN ?? userContext.Party?.Person?.SSN ?? userContext.UserParty?.SSN;
```
martinothamar: it would suck if `userContext.Party?.SSN == ""` 😄

src/Altinn.App.Core/Helpers/UserHelper.cs
```diff
@@ -50,55 +49,50 @@ public UserHelper(
...
+            },
+            AuthenticationLevel = tokenClaims.GetValueOrDefault(AltinnCoreClaimTypes.AuthenticationLevel) switch
             {
-                userContext.AuthenticationLevel = Convert.ToInt32(claim.Value, CultureInfo.InvariantCulture);
-            }
-        }
-
-        if (userContext.UserId == default)
-        {
-            throw new Exception("Could not get user profile - could not retrieve user ID from claims");
```
martinothamar: Think we should keep this, since the other exception doesn't cover `"0"`

martinothamar (approved):

src/Altinn.App.Core/Helpers/UserHelper.cs
danielskovli: Agreed. Very sucky... I don't know the likelyhood of all these edge cases, but perhaps I should check for empty and whitespace?

PR #937, opened 2024-11-28
Use 'instanceandevents' endpoint when updating process state, so that process and events are updated in a single transaction

martinothamar:
## Description
Uses the new endpoint which commits new process state and instance events in one transaction.

The related tests were a little funny. Test names said they were testing more than they actually were, it seems like most of the variation comes from the `AltinnTaskType` thingy. 

## Related Issue(s)
- https://github.com/Altinn/altinn-storage/issues/544

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -335,10 +335,6 @@ public class InstancesController : ControllerBase
 
             // create the instance
             instance = await _instanceClient.CreateInstance(org, app, instanceTemplate);
-            foreach (var instanceEvent in result.ProcessStateChange?.Events ?? [])
-            {
-                instanceEvent.InstanceId = instance.Id;
-            }
         }
         catch (Exception exception)
         {
@@ -570,10 +566,6 @@ public class InstancesController : ControllerBase
             }
 
             instance = await _instanceClient.CreateInstance(org, app, instanceTemplate);
-            foreach (var instanceEvent in processResult.ProcessStateChange?.Events ?? [])
-            {
-                instanceEvent.InstanceId = instance.Id;
-            }
 
             if (isCopyRequest && source is not null)
             {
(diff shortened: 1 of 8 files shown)


src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -683,6 +691,10 @@ [FromRoute] Guid instanceGuid
         ProcessChangeResult startResult = await _processEngine.GenerateProcessStartEvents(processStartRequest);
 
         targetInstance = await _instanceClient.CreateInstance(org, app, targetInstance);
+        foreach (var instanceEvent in startResult.ProcessStateChange?.Events ?? [])
```
martinothamar: These loops are needed because `GenerateProcessStartEvents` is creating the events, at which point `InstanceId` is null..

src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -335,6 +335,10 @@ public async Task<ActionResult<Instance>> Post(
 
             // create the instance
             instance = await _instanceClient.CreateInstance(org, app, instanceTemplate);
+            foreach (var instanceEvent in result.ProcessStateChange?.Events ?? [])
+            {
+                instanceEvent.InstanceId = instance.Id;
+            }
```
bjorntore, line 341: Was this a bug before this fix?

bjorntore (approved):

src/Altinn.App.Api/Controllers/InstancesController.cs
bjorntore: Was this a bug before this fix?

martinothamar: I don't think so, as it started to happen after switching to the new endpoint.. Don't know why, will do some digging to make sure

martinothamar: Overlooked the change in the dispatcher, it was setting `InstanceId` in the deleted function. I've now moved it into the client, but could put it into `DispatchToStorage` instead I guess.. Could also be more defensive and verify that `InstanceId` is consistent across the instance and the events 🤔

Issue #131, opened 2023-01-02
Move Altinn.Common.EFormidlingClient out of altinn-studio repo


RonnyB71:
### Description

After splitting into multiple repos this code does not belong in altinn-studio repo, but should have it's own repository.

We should also consider moving the EFormidling code in app-lib-dotnet repo out and into the same repository bundling everything related to Eformidling in the same repository.

danielskovli:
Fast forward 3 years, and we've come full circle. The client has now been re-absorbed into the `altinn-studio` repo along with the entire backend and frontend. It's been modernized at the same time, a few bugs have been fixed, and it now bundles with `altinn-studio/app/backend` per https://github.com/Altinn/altinn-studio/pull/19955.

PR #962, opened 2024-12-08
Move Altinn.Common.EFormidlingClient

martinothamar:
## Description
For now, just moved over the code.
It seems like no deployed apps are referencing this NuGet package directly, so in that case we should be able to just copy the code into Altinn.App.Core (it doesn't have any weird deps so I see no reason to keep it as a separate package)

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/131

## Verification
- [ ] All tests run green

## Documentation


danielskovli:
Superseded by https://github.com/Altinn/altinn-studio/pull/19955

PR #963, opened 2024-12-10
Add CustomOpenApiController for easier client generation for end user systems.

ivarne:
Add new strategy for auto generating openApi specifications using the actual types from the app through a custom controller that builds up the document, instead of autogenerating from controller metadata. The new document is intended to only cover the endpoints used by End user systems that integrate with an app.

image

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Api/Controllers/CustomOpenApiController.cs
```diff
@@ -0,0 +1,1248 @@
...
+        var document = new OpenApiDocument()
+        {
+            Info = new()
+            {
+                Title =
+                    "Altinn 3 App API for "
+                    + (
+                        appMetadata.Title.TryGetValue("en", out var englishTitle) ? englishTitle
+                        : appMetadata.Title.TryGetValue("nb", out var norwegianTitle) ? norwegianTitle
+                        : "Unknown"
```
martinothamar:
```suggestion
                        : appMetadata.AppIdentifier.ToString()
```

src/Altinn.App.Api/Extensions/SwaggerFilterExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+        services.Configure<SwaggerGenOptions>(c =>
+        {
+            c.DocumentFilter<DocumentFilter>();
+        });
+    }
+}
+
+internal class DocumentFilter : IDocumentFilter
+{
+    public void Apply(OpenApiDocument swaggerDoc, DocumentFilterContext context)
```
martinothamar, line 23: Could we ensure somehow that this filter applies only to swagger docs that we "own"? As in, if developers make their own docs we don't touch them

test/Altinn.App.Api.Tests/Altinn.App.Api.Tests.csproj
```diff
@@ -15,7 +15,7 @@
     <PackageReference Include="Moq" Version="4.20.72" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="7.2.0" />
     <PackageReference Include="xunit" Version="2.9.3" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="3.0.1">
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.8.2">
```
martinothamar: What happened? 😄

martinothamar (commented):
Generally LGTM. 

I wonder if we can simplify enabling this by just us doing it instead? Something like
```csharp
public static void AddAltinnAppControllersWithViews(this IServiceCollection services)
{
        ....

        services.Configure<SwaggerUIOptions>(c =>
        {
            string applicationId = StartupHelper.GetApplicationId();
            c.SwaggerEndpoint($"/{applicationId}/v1/customOpenapi.json", "Altinn Simple App API");
        });
}
```

Alternatively have a setting somewhere. I guess it depends on what we end up calling this. If it's just the "simple" variant then I think it makes sense to have enabled by default.

I also did get a build error and had to remove the packagereference. So, I think the remaining work then is

* Configuration story (how to enable this feature)
* Do we warn users that from v8.6 they will have to remove or update their packageref, or can we do something automatically here
* Docs

src/Altinn.App.Api/Extensions/SwaggerFilterExtensions.cs
ivarne: Not sure what would be a sensible restriction, but on second thought I think it makes sense to not throw if these are missing, so then the only action would be to remove our duplicate paths. It seems unlikely that someone would like to have these paths in their custom swagger doc.

I could add something like this, but I can't really se a way to detect if this is ours or theirs, that isn't likely to break. (```"Altinn App Api"``` is currently specified in the apps ``Program.cs``)
```c#
if (swaggerDoc.Info?.Title != "Altinn App Api")
{
    return;
}
```

test/Altinn.App.Api.Tests/Altinn.App.Api.Tests.csproj
ivarne: I would guess that I updated to 2.8.2 at some point, and v3 was released later and updated on main.

ivarne:
I found a way to avoid the need for modification to Program.cs by adding the configuration in our setup instead. Seems to work fine with the existing config in Program.cs
https://github.com/Altinn/app-lib-dotnet/blob/9f9d361812d2a06f4ded06b16b87204672c3db35/src/Altinn.App.Api/Extensions/ServiceCollectionExtensions.cs#L131-L136

I don't have any idea how to improve the situation regarding the version conflict on `Swashbuckle.AspNetCore`, but at some point we need to bite that bullet, and it's a pretty easy fix when updating.

martinothamar (approved):

PR #983, opened 2024-12-16
Endpoint for fetching current party roles

adamhaeger:
Adds endpoint for fetching the users current party roles.

You need this PR in localtest to test:

https://github.com/Altinn/app-localtest/pull/131

## Description
- Adds endpoint `/api/authorization/roles` to AuthorizationController
- Refactored functionality to fetch current party to reusable function
- Implemented fetching of the current parties.

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/issues/2089

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/AuthorizationController.cs
+++ b/src/Altinn.App.Api/Controllers/AuthorizationController.cs
@@ -99,7 +99,7 @@ public class AuthorizationController : Controller
         }
 
         int userId = userContext.UserId;
-        List<Role> roles = await _authorization.GetUserRolesAsync(userId, currentParty.PartyId);
+        IEnumerable<Role> roles = await _authorization.GetUserRolesAsync(userId, currentParty.PartyId);
 
         return Ok(roles);
     }
--- a/src/Altinn.App.Api/Controllers/AuthorizationController.cs
+++ b/src/Altinn.App.Api/Controllers/AuthorizationController.cs
@@ -46,7 +46,7 @@ public class AuthorizationController : Controller
     [HttpGet("{org}/{app}/api/authorization/parties/current")]
     public async Task<ActionResult> GetCurrentParty(bool returnPartyObject = false)
     {
-        Party? currentParty = await GetCurrentPartyAsync(HttpContext);
+        (Party? currentParty, _) = await GetCurrentPartyAsync(HttpContext);
 
         if (returnPartyObject)
         {
@@ -86,30 +86,30 @@ public class AuthorizationController : Controller
     /// <summary>
     /// Fetches roles for current party.
     /// </summary>
-    /// <returns>Boolean indicating if the selected party is valid.</returns>
+    /// <returns>List of roles for the current user and party.</returns>
     [Authorize]
     [HttpGet("{org}/{app}/api/authorization/roles")]
     public async Task<IActionResult> GetRolesForCurrentParty()
     {
-        Party? currentParty = await GetCurrentPartyAsync(HttpContext);
-        UserContext userContext = await _userHelper.GetUserContext(HttpContext);
-        int userId = userContext.UserId;
+        (Party? currentParty, UserContext userContext) = await GetCurrentPartyAsync(HttpContext);
 
         if (currentParty == null)
         {
             return BadRequest("Both userId and partyId must be provided.");
         }
 
+        int userId = userContext.UserId;
         List<Role> roles = await _authorization.GetUserRolesAsync(userId, currentParty.PartyId);
+
         return Ok(roles);
     }
 
     /// <summary>
-    /// Helper method to retrieve the current party from the HTTP context.
+    /// Helper method to retrieve the current party and user context from the HTTP context.
     /// </summary>
     /// <param name="context">The current HttpContext.</param>
-    /// <returns>The current party or null if none could be determined.</returns>
-    private async Task<Party?> GetCurrentPartyAsync(HttpContext context)
+    /// <returns>A tuple containing the current party and user context.</returns>
+    private async Task<(Party? party, UserContext userContext)> GetCurrentPartyAsync(HttpContext context)
     {
         UserContext userContext = await _userHelper.GetUserContext(context);
         int userId = userContext.UserId;
@@ -150,6 +150,6 @@ public class AuthorizationController : Controller
             );
         }
 
-        return userContext.Party;
+        return (userContext.Party, userContext);
     }
 }
--- a/src/Altinn.App.Api/Controllers/AuthorizationController.cs
+++ b/src/Altinn.App.Api/Controllers/AuthorizationController.cs
@@ -105,7 +105,7 @@ public class AuthorizationController : Controller
         }
 
         int userId = userContext.UserId;
-        IEnumerable<Role> roles = await _authorization.GetUserRolesAsync(userId, currentParty.PartyId);
+        IEnumerable<Role> roles = await _authorization.GetUserRoles(userId, currentParty.PartyId);
 
         return Ok(roles);
     }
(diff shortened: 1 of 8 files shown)


src/Altinn.App.Api/Controllers/AuthorizationController.cs
```diff
@@ -123,4 +82,77 @@ public async Task<IActionResult> ValidateSelectedParty(int userId, int partyId)
...
         }
     }
+
+    /// <summary>
+    /// Fetches roles for current party.
+    /// </summary>
+    /// <returns>Boolean indicating if the selected party is valid.</returns>
+    [Authorize]
+    [HttpGet("{org}/{app}/api/authorization/roles")]
+    public async Task<IActionResult> FetchRolesForCurrentParty()
```
danielskovli: All other methods are named `GetWhatever...`. It doesn't really matter, but did you mean to break the convention with `Fetch...`?

src/Altinn.App.Api/Controllers/AuthorizationController.cs
```diff
@@ -123,4 +82,77 @@ public async Task<IActionResult> ValidateSelectedParty(int userId, int partyId)
...
+        int userId = userContext.UserId;
+
+        if (currentParty == null)
+        {
+            return BadRequest("Both userId and partyId must be provided.");
+        }
+
+        List<Role> roles = await _authorization.GetUserRolesAsync(userId, currentParty.PartyId);
+
+        // TODO: implement actual logic here using currentParty
```
danielskovli: I don't understand this comment, the logic looks complete to me

danielskovli (approved):
LGTM

src/Altinn.App.Api/Controllers/AuthorizationController.cs
adamhaeger: Old, comment, will remove :)

src/Altinn.App.Api/Controllers/AuthorizationController.cs
adamhaeger: Did not mean to break convention! Will change to GetWhatever :)

src/Altinn.App.Api/Controllers/AuthorizationController.cs
```diff
@@ -123,4 +82,77 @@ public async Task<IActionResult> ValidateSelectedParty(int userId, int partyId)
...
+    /// Fetches roles for current party.
+    /// </summary>
+    /// <returns>Boolean indicating if the selected party is valid.</returns>
+    [Authorize]
+    [HttpGet("{org}/{app}/api/authorization/roles")]
+    public async Task<IActionResult> FetchRolesForCurrentParty()
+    {
+        Party? currentParty = await GetCurrentPartyAsync(HttpContext);
+        UserContext userContext = await _userHelper.GetUserContext(HttpContext);
+        int userId = userContext.UserId;
```
martinothamar: Could you return this from the helper method instead? So we don't go through `GetUserContext` twice

src/Altinn.App.Api/Controllers/AuthorizationController.cs
danielskovli: I mean, it's not a wildly adopted convention, but in the context of this PR the other methods seem to lean towards `Get` 🙂

src/Altinn.App.Core/Infrastructure/Clients/Authorization/AuthorizationClient.cs
```diff
@@ -179,4 +180,52 @@ List<string> actions
...
         return MultiDecisionHelper.ValidatePdpMultiDecision(actionsResult, response.Response, user);
     }
+
+    /// <summary>
+    /// Retrieves roles for a user on a specified party.
+    /// </summary>
+    /// <param name="userId">The user id.</param>
+    /// <param name="userPartyId">The user party id.</param>
+    /// <returns>A list of roles for the user on the specified party.</returns>
+public async Task<List<Role>> GetUserRolesAsync(int userId, int userPartyId)
```
martinothamar: I wonder if we should return `IEnumerable<>` or `IReadOnlyList<>` instead, to express the intended usage

src/Altinn.App.Core/Infrastructure/Clients/Authorization/AuthorizationClient.cs
```diff
@@ -179,4 +180,52 @@ List<string> actions
...
+{
+    List<Role> roles = new();
```
martinothamar: Usually we wrap the body in activity (distributed trace spans). You can probably find example usages of `Telemetry`

src/Altinn.App.Core/Infrastructure/Clients/Authorization/AuthorizationClient.cs
```diff
@@ -179,4 +180,52 @@ List<string> actions
...
+        }
+    }
+    catch (Exception ex)
+    {
+        _logger.LogError(
+            ex,
+            "An error occurred while retrieving roles for userId {UserId} and partyId {PartyId}",
+            userId,
+            userPartyId
+        );
```
martinothamar: I'm not sure about the error handling here. If the API fails for some reason we return an empty list, which might make it look like the user has no roles when in fact the problem was that we couldn't find out at all

adamhaeger: Good point.

src/Altinn.App.Core/Internal/Auth/IAuthorizationClient.cs
```diff
@@ -50,4 +51,12 @@ Task<bool> AuthorizeAction(
...
     /// <returns></returns>
     Task<Dictionary<string, bool>> AuthorizeActions(Instance instance, ClaimsPrincipal user, List<string> actions);
+
+    /// <summary>
+    /// Retrieves roles for a user on a specified party.
+    /// </summary>
+    /// <param name="userId">The user id.</param>
+    /// <param name="userPartyId">The user party id.</param>
+    /// <returns>A list of roles for the user on the specified party.</returns>
+    Task<IEnumerable<Role>> GetUserRolesAsync(int userId, int userPartyId);
```
martinothamar: I don't think we use `Async` suffix in the majority of cases, so maybe we should remove that (same goes for the telemetry trace)

test/Altinn.App.Api.Tests/OpenApi/swagger.json
```diff
@@ -507,6 +507,36 @@
...
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "OK"
```
ivarne: We need to annotate the response so that swagger can tell what it is

src/Altinn.App.Api/Controllers/AuthorizationController.cs
```diff
@@ -123,4 +82,74 @@ public async Task<IActionResult> ValidateSelectedParty(int userId, int partyId)
...
         }
     }
+
+    /// <summary>
+    /// Fetches roles for current party.
+    /// </summary>
+    /// <returns>List of roles for the current user and party.</returns>
+    [Authorize]
+    [HttpGet("{org}/{app}/api/authorization/roles")]
+    public async Task<IActionResult> GetRolesForCurrentParty()
```
ivarne, line 98: We need to add annotations so that the openapi generation can figure out what our endpoints returns.

Should look something like this:
```suggestion
    [ProducesResponseType(typeof(IEnumerable<Roles), StatusCodes.Status200OK)]
    [ProducesResponseType(typeof(ProblemDetails), StatusCodes.Status400BadRequest)]
    public async Task<IActionResult<IEnumerable<Roles>> GetRolesForCurrentParty()
```

martinothamar (approved):
LGTM

PR #984, opened 2024-12-17
Correspondence client updates

danielskovli:
## Description
As discussed on [Slack](https://digdir.slack.com/archives/C079AJR5TEX/p1734429151436709?thread_ts=1734352363.739069&cid=C079AJR5TEX), the Correspondence API endpoints for `send` and `details` will soon require different authorisation scopes.

The Correspondence client in general was poorly designed in this regard, with the authorisation factory attempting to provide the same scopes for every request regardless of target endpoint.

## Summary
The changes offered in this pull request aren't very large, but some things have moved around a bit. Here's the tl;dr:
- The scope requirements for each request have been moved from a static value in `CorrespondenceAuthorisationFactory`, to a configuration property in `CorrespondencePayloadBase` 
- The `CorrespondenceClient.GetStatus` method now requires a new scope: `altinn:correspondence.read`
- `CorrespondenceClient.AuthorisationResolver` has moved to `CorrespondenceAuthorisationFactory.Resolve`

## Verification
- [x] **Your** code builds clean without any errors or warnings
- [x] Manual testing done (required)
- [x] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [x] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #985, opened 2024-12-17
Correspondence debug addition

danielskovli:
## Description
Forgot to include this tiny little addition with #984. Extremely comical to go through the whole PR process for this little change, but here we are.

Had this log line been included in the first place, the debugging of an issue related to #984 would have been much smoother.


danielskovli:
Actually, maybe the same information is visible from the HttpClient logger 🤷

cammiida (approved):
Much wow. Such code 🥇

HauklandJ (approved):
lgtm

PR #998, opened 2024-12-19
Add IExceptionHandler and use it to warn about old localtest

ivarne:
This PR introduces a new concept `IExceptionHandler` that currently only handles the new `AltinnAppException` and customises the `ProblemDetails` response to include the data from the exception

The use case is to warn about old localtest, but in the future I think this Is a richer way to both ensure that exceptions gets logged, and that api users gets more helpful error messages.

## New stuff
* New RequestExceptionHandler: IExceptionHandler to create ProblemDetails
* New AltinnAppException with the fields of ProblemDetails
* Throw exception for old localtest with text for ProblemDetails about upgrading
* Modify catch blocks to let the exception bubble through.

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -23,6 +23,7 @@ using Altinn.App.Core.Internal.Process;
 using Altinn.App.Core.Internal.Profile;
 using Altinn.App.Core.Internal.Registers;
 using Altinn.App.Core.Models;
+using Altinn.App.Core.Models.Exceptions;
 using Altinn.App.Core.Models.Process;
 using Altinn.App.Core.Models.Validation;
 using Altinn.Authorization.ABAC.Xacml.JsonProfile;
@@ -368,6 +369,12 @@ public class InstancesController : ControllerBase
             // notify app and store events
             _logger.LogInformation("Events sent to process engine: {Events}", change?.Events);
             await _processEngine.HandleEventsAndUpdateStorage(instance, null, change?.Events);
+        }
+        // Let the AltinnAppException through as it is handled in IExceptionHandler
+        catch (AltinnAppException)
+        {
+            // TODO: handle all exceptions (not just this) in IExceptionHandler
+            throw;
         }
         catch (Exception exception)
         {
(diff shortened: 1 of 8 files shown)


src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -369,6 +370,12 @@ await _instanceClient.DeleteInstance(
             _logger.LogInformation("Events sent to process engine: {Events}", change?.Events);
             await _processEngine.HandleEventsAndUpdateStorage(instance, null, change?.Events);
         }
+        // Let the AltinnAppException through as it is handled in IExceptionHandler
+        catch (AltinnAppException)
```
martinothamar, line 374: The part I don't like about this is the kind of coupling we get that isn't expressed anywhere (at least not cohesively). IMO it is hard to trace and maintain over time, so I think it's not a good solution for something to be used application-wide. In this particular case, the caller and callee have to be kept in sync, but the compiler doesn't help us :( 

Since this is all internal though, we could let this through as long as we think of it as a temporary solution to this specific problem.

ivarne: Do you object to this specific case of letting an exception through (which I totally agree, but wanted to preserve behavior). Or the general concept of throwing exceptions with explicit problem details info and a central catch middleware?

If this approach is accepted it would obviously cause a bigger change to how we handle errors.

src/Altinn.App.Api/Extensions/WebApplicationBuilderExtensions.cs
```diff
@@ -16,10 +16,7 @@ public static class WebApplicationBuilderExtensions
     public static IApplicationBuilder UseAltinnAppCommonConfiguration(this IApplicationBuilder app)
     {
         var appId = StartupHelper.GetApplicationId();
-        if (app is WebApplication webApp && webApp.Environment.IsDevelopment())
-        {
-            app.UseDeveloperExceptionPage();
```
ivarne, line 21: Note that develperexeptionpage gets added by default in net6 and forwads

src/Altinn.App.Api/Controllers/InstancesController.cs
ivarne: I have also purposefully limited the available status codes that can be returned, to reduce surprises for api documentation.

martinothamar: The latter, exception based error handling, using `IExceptionHandler` in this case. `IExceptionHandler` is one of those global mechanisms that might create surprises for app devs in the same way we encountered with `JsonOptions`. So in that sense I like local try/catch better

PR #1007, opened 2025-01-06
Validate that the localtest version is new enough for the app on startup

martinothamar:
## Description
Checks that the localtest version running locally (when `GeneralSettings.HosteName` is set to `local.altinn.cloud`) is new enough (version 1)

image

Logs error and shuts down if 
* Version <= 0
* Version endpoint was not found

In other failure conditions it still logs and error and retries:

image

## Related Issue(s)
- Depends on https://github.com/Altinn/app-localtest/pull/134

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Configuration/GeneralSettings.cs
+++ b/src/Altinn.App.Core/Configuration/GeneralSettings.cs
@@ -34,6 +34,9 @@ public class GeneralSettings
     /// </summary>
     public string HostName { get; set; } = "local.altinn.cloud";
 
+    // Only here to be overridden from tests
+    internal string LocaltestUrl { get; set; } = "http://local.altinn.cloud";
+
     /// <summary>
     /// The externally accesible base url for the app with trailing /
     /// </summary>
--- a/src/Altinn.App.Core/Configuration/GeneralSettings.cs
+++ b/src/Altinn.App.Core/Configuration/GeneralSettings.cs
@@ -34,10 +34,10 @@ public class GeneralSettings
     /// </summary>
     public string HostName { get; set; } = "local.altinn.cloud";
 
-    // Only here to be overridden from tests
-    internal string LocaltestUrl { get; set; } = "http://local.altinn.cloud";
-
-    internal bool DisableLocaltestValidation { get; set; }
+    /// <summary>
+    /// Gets or sets a value indicating whether to disable localtest validation on startup.
+    /// </summary>
+    public bool DisableLocaltestValidation { get; set; }
 
     /// <summary>
     /// The externally accesible base url for the app with trailing /
--- a/src/Altinn.App.Core/Configuration/GeneralSettings.cs
+++ b/src/Altinn.App.Core/Configuration/GeneralSettings.cs
@@ -37,6 +37,8 @@ public class GeneralSettings
     // Only here to be overridden from tests
     internal string LocaltestUrl { get; set; } = "http://local.altinn.cloud";
 
+    internal bool DisableLocaltestValidation { get; set; }
+
     /// <summary>
     /// The externally accesible base url for the app with trailing /
     /// </summary>
--- a/src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
@@ -177,7 +177,8 @@ public static class ServiceCollectionExtensions
         services.Configure<FrontEndSettings>(configuration.GetSection(nameof(FrontEndSettings)));
         services.Configure<PdfGeneratorSettings>(configuration.GetSection(nameof(PdfGeneratorSettings)));
 
-        services.AddLocaltestValidation(configuration);
+        if (env.IsDevelopment())
+            services.AddLocaltestValidation(configuration);
 
         AddValidationServices(services, configuration);
         AddAppOptions(services);
--- a/src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
@@ -177,7 +177,7 @@ public static class ServiceCollectionExtensions
         services.Configure<FrontEndSettings>(configuration.GetSection(nameof(FrontEndSettings)));
         services.Configure<PdfGeneratorSettings>(configuration.GetSection(nameof(PdfGeneratorSettings)));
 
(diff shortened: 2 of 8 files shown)


HauklandJ (approved):

src/Altinn.App.Core/Configuration/GeneralSettings.cs
```diff
@@ -34,6 +34,11 @@ public class GeneralSettings
     /// </summary>
     public string HostName { get; set; } = "local.altinn.cloud";
 
+    // Only here to be overridden from tests
+    internal string LocaltestUrl { get; set; } = "http://local.altinn.cloud";
```
ivarne: How is `GeneralSettings.LocaltestUrl` used in relation to `LocaltestValidation.ExpectedHostname`. I don't think you ment to use both.

src/Altinn.App.Core/Internal/LocaltestValidation.cs
```diff
@@ -0,0 +1,210 @@
...
+{
+    public static IServiceCollection AddLocaltestValidation(
+        this IServiceCollection services,
+        IConfiguration configuration
+    )
+    {
+        services.AddSingleton<LocaltestValidation>();
+        if (configuration.GetValue<bool>("GeneralSettings:DisableLocaltestValidation"))
+            return services;
+        services.AddSingleton<IHostedService>(provider => provider.GetRequiredService<LocaltestValidation>());
```
ivarne: Why doesn't this work?
```suggestion
        services.AddSingleton<IHostedService, LocaltestValidation>();
```

ivarne:
This don't work for people that can't use port 80, but rather 8080 (or any other port they might want to use)

In [`PlatformSettings`](https://github.com/Altinn/app-lib-dotnet/blob/ivarne/betterErrorMessage/src/Altinn.App.Core/Configuration/PlatformSettings.cs) the default way to access localtest services is using `http://localhost:5101/`, and I have never had an instance where users had issues with that.

src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -176,6 +177,8 @@ IWebHostEnvironment env
         services.Configure<FrontEndSettings>(configuration.GetSection(nameof(FrontEndSettings)));
         services.Configure<PdfGeneratorSettings>(configuration.GetSection(nameof(PdfGeneratorSettings)));
 
+        services.AddLocaltestValidation(configuration);
+
```
ivarne, line 182: Should this be wrapped in `env.IsDevelopment()`?

src/Altinn.App.Core/Internal/LocaltestValidation.cs
martinothamar: Old habit of mine, but I think that becomes 2 different singletons since the service type is different?

src/Altinn.App.Core/Configuration/GeneralSettings.cs
martinothamar: `LocaltestUrl` was introduced to support overriding the HTTP client request URL, so that's the only use of it. `ExpectedHostname` is only used to check `HostName` (for when to run the code in the hosted service). There is a test for simulating DNS failure where `LocaltestUrl` differs from `HostName` (if I changed `HostName` the code wouldn't run). I found this confusing myself too tbh 😅

martinothamar: In the latest commit I removed this field and used `PlatformSettings` instead for that purpose. I think that is what you proposed in the other comment?

martinothamar:
> This don't work for people that can't use port 80, but rather 8080 (or any other port they might want to use)

Oo good point I had forgotten about that. I use `PlatformSettings` instead in the latest commit, which I think is what you suggested here. 5101 is the default for both the docker and podman compose files, and if app devs have changed it, they will probably have changed `PlatformSettings` as well

src/Altinn.App.Core/Configuration/GeneralSettings.cs
ivarne: Yes.

src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
martinothamar: Sure 👍

src/Altinn.App.Core/Internal/LocaltestValidation.cs
ivarne: Ah, so this is a hack to be able to both have a generic `IHostedService` and have the `LocaltestValidation` as an injectable service.

ivarne: Which you need for testing in order to avoid `provider.GetRequiredServices<IHostedService>().OfType<LocaltestValidation>().Single()`?

martinothamar: yup! which is not needed here at all (I started out doing things differently), so I'll cleanup

test/Altinn.App.Core.Tests/Internal/LocaltestValidationTests.cs
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    [Fact]
+    public async Task Test_Dns_Failure()
+    {
+        await using var fixture = Fixture.Create(registerCustomAppServices: services =>
+            services.Configure<PlatformSettings>(settings =>
+                settings.ApiStorageEndpoint = ReplaceHost(
+                    settings.ApiStorageEndpoint,
+                    "provoke-dns-fail.local.altinn.cloud"
```
ivarne: Is this dependent on external dns and the test will fail when someone adds a wildcard *.local.altinn.cloud registration?

ivarne (approved):

test/Altinn.App.Core.Tests/Internal/LocaltestValidationTests.cs
martinothamar: yes! Could mock out dns resolution by resolving the host in a handler or something, but I wanted to leave the `HttpClient` mostly intact so that I get all the real exceptions and behavior. I changed the address to `provoke-dns-fail.local.not-altinn-at-all.cloud` instead. If someone decides to register that I will pay the price of changing the test 😄

PR #1032, opened 2025-01-15
Fix a few issues when verifying swagger documentation

ivarne:
* Fix StartupHelper.IncludeXmlComments to only add xml files that exists
* Use Verify.VerifyJson instead of FluentAssertions to compare the full spec
* Continue saving swagger.json because it might be referenced.
* Fix .editorconfig so that editor does not trim whitespace in verify files.
* Generate xml doc for Altinn.App.Api.Tests.csproj
* Don't generate swagger.yml (it just adds noise, and isn't much easier to read and the verify file is good enough)


martinothamar (approved):

martinothamar:
> Continue saving swagger.json because it might be referenced.

what does this mean?

ivarne:
>> Continue saving swagger.json because it might be referenced.

> what does this mean?

I'll reference this file directly in the openapi branch in order to generate a client and verify that Kiota think it is valid. I considered using `swagger.verify.json`, but it seemed wrong to reference the `*.verify.*` file.

Issue #466, opened 2024-02-28
ApplicationMetadata class overrides nuget version in studio


ivarne:
https://github.com/Altinn/altinn-studio/pull/12407

We should set the version after deserialization, not as part of class initialization

PR #1044, opened 2025-01-18
Simplify PR builds, change pr version format

martinothamar:
## Description

This PR now changes the PR release version format to something like this:
`8.5.0-pr.12.signing-notification.2ae9d133`

The version building task is now looking more like the frontend version, which is nice (and simpler kind of). 

Tested these commands to mimic release builds and stable/normal builds:
```sh
find . -iname "bin" -o -iname "obj" | xargs rm -rf && \
   CI=true MINVERVERSIONOVERRIDE="8.5.0-pr.12.signing-notification.2ae9d133" dotnet restore && \
   CI=true MINVERVERSIONOVERRIDE="8.5.0-pr.12.signing-notification.2ae9d133" dotnet build --configuration Release --no-restore -p:Deterministic=true -p:BuildNumber=12 && \
   CI=true MINVERVERSIONOVERRIDE="8.5.0-pr.12.signing-notification.2ae9d133" dotnet pack AppLibDotnet.sln --configuration Release --no-restore --no-build -p:Deterministic=true -p:BuildNumber=12

git tag v8.6.0 && \
  find . -iname "bin" -o -iname "obj" | xargs rm -rf && \
  CI=true dotnet restore && \
  CI=true dotnet build --configuration Release --no-restore -p:Deterministic=true -p:BuildNumber=12 && \
  CI=true dotnet pack AppLibDotnet.sln --configuration Release --no-restore --no-build -p:Deterministic=true -p:BuildNumber=12 && \
  git tag -d v8.6.0
```

And it looks like the assembly version is being set correctly in both cases, even without removing the `Directory.Build.props` stuff....

image

### Old

The text below is the old attempt at making this sane....

We were not setting the correct value in `AltinnNugetVersion` when running a PR release. AFAICT the `Version` parsed from the assembly name will only ever hold major.minor.build/patch.revision, while we have a lot more info in our full version. MinVer will always set the `InformationalVersion` property to the full version, regardless if it is computed or comes from the environment override in CI, so this PR sets the version based on `InformationalVersion` instead of the version from the assembly name (if possible).

Tested by running the following
```sh
# PR release build
CI=true MINVERVERSIONOVERRIDE='8.6.0-preview.testing' dotnet build AppLibDotnet.sln -v m -c Release

# Normal release build
CI=true dotnet build --configuration Release --no-restore -p:Deterministic=true -p:BuildNumber=10
```

Which seems to have the same result in terms of format in the  `AssemblyInformationalVersionAttribute`

image

image

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation


ivarne:
Studio is using a rather ugly override technique in order to use this class

https://github.com/Altinn/app-lib-dotnet/issues/466

It's an issue very specific to the way studio use this class, so maybe not a huge problem.

PR #1046, opened 2025-01-20
implements default signing task validator

### LINKED ISSUE #801: Standard implementation of signing task validator
TODO: what should this entail? Input from TEs needed.
Thoughts: Minimum required signatures - count/everyone? Validate hash - if mismatch delete signatures? In practice, all signatures associated with a sign task are deleted on task entry, so this should not be necessary as data from previous tasks should be locked.

MVP: Alle delegerte har signert

cammiida:
## Description
Default validator for all signing tasks. Runs by default if not `runDefaultValidator="false"` in `process.bpmn`.

Should it be called `runDefaultValidations` instead? 🤷‍♀️ 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
+++ b/src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
@@ -124,7 +124,7 @@ internal class SigningTaskValidator : IValidator
         ];
     }
 
-    public static async Task<Tuple<Exception?, T?>> CatchError<T>(Task<T> task)
+    private static async Task<Tuple<Exception?, T?>> CatchError<T>(Task<T> task)
     {
         try
         {
--- a/src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
+++ b/src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
@@ -1,0 +1,139 @@
+using Altinn.App.Core.Features.Signing.Interfaces;
+using Altinn.App.Core.Helpers.Serialization;
+using Altinn.App.Core.Internal.App;
+using Altinn.App.Core.Internal.Data;
+using Altinn.App.Core.Internal.Instances;
+using Altinn.App.Core.Internal.Process;
+using Altinn.App.Core.Internal.Process.Elements.AltinnExtensionProperties;
+using Altinn.App.Core.Models;
+using Altinn.App.Core.Models.Validation;
+using Microsoft.Extensions.Logging;
+
+namespace Altinn.App.Core.Features.Validation.Default;
+
+/// <summary>
+/// Validates that all required parties have signed the current task
+/// </summary>
+internal class SigningTaskValidator : IValidator
+{
+    private readonly IProcessReader _processReader;
+    private readonly ISigningService _signingService;
+    private readonly IDataClient _dataClient;
+    private readonly IInstanceClient _instanceClient;
+    private readonly IAppMetadata _appMetadata;
+    private readonly ModelSerializationService _modelSerialization;
+    private readonly ILogger<SigningTaskValidator> _logger;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SigningTaskValidator"/> class.
+    /// </summary>
+    public SigningTaskValidator(
+        ILogger<SigningTaskValidator> logger,
+        IProcessReader processReader,
+        ISigningService signingService,
+        IDataClient dataClient,
+        IInstanceClient instanceClient,
+        IAppMetadata appMetadata,
+        ModelSerializationService modelSerialization
+    )
+    {
+        _logger = logger;
+        _processReader = processReader;
+        _signingService = signingService;
+        _dataClient = dataClient;
+        _instanceClient = instanceClient;
+        _appMetadata = appMetadata;
+        _modelSerialization = modelSerialization;
+    }
+
+    /// <summary>
+    /// We implement <see cref="ShouldRunForTask"/> instead
+    /// </summary>
+    public string TaskId => "*";
+
+    /// <summary>
+    /// Only run for tasks that are of type "signing"
+    /// </summary>
+    public bool ShouldRunForTask(string taskId)
+    {
+        return _processReader.GetAltinnTaskExtension(taskId)?.TaskType == "signing";
+    }
+
+    public bool NoIncrementalValidation => true;
+
+    // Should never be called because NoIncrementalValidation is true
+    public Task<bool> HasRelevantChanges(IInstanceDataAccessor dataAccessor, string taskId, DataElementChanges changes)
+    {
+        throw new NotImplementedException();
+    }
+
+    public async Task<List<ValidationIssue>> Validate(
+        IInstanceDataAccessor dataAccessor,
+        string taskId,
+        string? language
+    )
+    {
+        AltinnSignatureConfiguration? signingConfiguration = (
+            _processReader.GetAltinnTaskExtension(taskId)?.SignatureConfiguration
+        );
+
+        if (signingConfiguration == null)
+        {
+            _logger.LogError($"No signing configuration found for task {taskId}");
+            return [];
+        }
+
+        var (getAppMetadataError, appMetadata) = await CatchError(_appMetadata.GetApplicationMetadata());
+        if (getAppMetadataError != null || appMetadata == null)
+        {
+            _logger.LogError(getAppMetadataError, "Error while fetching application metadata");
+            return [];
+        }
+
+        var cachedDataMutator = new InstanceDataUnitOfWork(
+            instance: dataAccessor.Instance,
+            _dataClient,
+            _instanceClient,
+            appMetadata,
+            _modelSerialization
+        );
+
+        var (getSigneeContextsError, signeeContexts) = await CatchError(
+            _signingService.GetSigneeContexts(cachedDataMutator, signingConfiguration)
+        );
+        if (getSigneeContextsError != null || signeeContexts == null)
+        {
+            _logger.LogError(getSigneeContextsError, "Error while fetching signee contexts");
+            return [];
+        }
+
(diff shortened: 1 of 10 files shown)


src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, $"Error while fetching signing configuration for task {taskId}");
+            return false;
+        }
+
+        AltinnSignatureConfiguration? signingConfiguration = taskConfig?.SignatureConfiguration;
+
+        return signingConfiguration?.RunDefaultValidator == true && taskConfig?.TaskType == "signing";
```
danielskovli: We don't have a firm policy on this, but I've taken up the job of trying to make us use `is` and `is not` over `==` and `!=` when comparing to built-in types.

Do with that information as you please 😄

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+        IInstanceDataAccessor dataAccessor,
+        string taskId,
+        string? language
+    )
+    {
+        AltinnSignatureConfiguration? signingConfiguration = (
+            _processReader.GetAltinnTaskExtension(taskId)?.SignatureConfiguration
+        );
+
+        if (signingConfiguration == null)
```
HauklandJ:
```suggestion
        if (signingConfiguration is null)
```

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            _logger.LogError($"No signing configuration found for task {taskId}");
+            return [];
+        }
+
+        var (getAppMetadataError, appMetadata) = await CatchError(_appMetadata.GetApplicationMetadata());
+        if (getAppMetadataError != null || appMetadata == null)
```
HauklandJ:
```suggestion
        if (getAppMetadataError is not null || appMetadata is null)
```

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+            _dataClient,
+            _instanceClient,
+            appMetadata,
+            _modelSerialization
+        );
+
+        var (getSigneeContextsError, signeeContexts) = await CatchError(
+            _signingService.GetSigneeContexts(cachedDataMutator, signingConfiguration)
+        );
+        if (getSigneeContextsError != null || signeeContexts == null)
```
HauklandJ:
```suggestion
        if (getSigneeContextsError is not null || signeeContexts is null)
```

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+using Altinn.App.Core.Models;
+using Altinn.App.Core.Models.Validation;
+using Microsoft.Extensions.Logging;
+
+namespace Altinn.App.Core.Features.Validation.Default;
+
+/// <summary>
+/// Default validator for signing tasks. Validates that all parties have signed the current task.
+/// </summary>
+internal class SigningTaskValidator : IValidator
```
danielskovli: Can/should this be `sealed`?

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+        var (getSigneeContextsError, signeeContexts) = await CatchError(
+            _signingService.GetSigneeContexts(cachedDataMutator, signingConfiguration)
+        );
+        if (getSigneeContextsError != null || signeeContexts == null)
+        {
+            _logger.LogError(getSigneeContextsError, "Error while fetching signee contexts");
+            return [];
+        }
+
+        var allHaveSigned = signeeContexts.All(signeeContext => signeeContext.SignDocument != null);
```
HauklandJ:
```suggestion
        var allHaveSigned = signeeContexts.All(signeeContext => signeeContext.SignDocument is not null);
```

danielskovli (approved):
LGTM 🥇

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+                Severity = ValidationIssueSeverity.Error,
+                Description = ValidationIssueCodes.DataElementCodes.MissingSignatures,
+            },
+        ];
+    }
+
+    /// <summary>
+    /// Catch exceptions from a task and return them as a tuple with the result.
+    /// </summary>
+    private static async Task<Tuple<Exception?, T?>> CatchError<T>(Task<T> task)
```
HauklandJ: Should this be an internal utility?

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
cammiida: Haha, that's totally a blunder by me. Fixing!

bjorntore (approved):

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
HauklandJ: It would be nice to have a "TaskTypeConst" similar to LanguageConst, but I consider it outside the scope of this pr. I will create and issue for it.

HauklandJ (approved):
🙌

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+    public bool ShouldRunForTask(string taskId)
+    {
+        AltinnTaskExtension? taskConfig;
+        try
+        {
+            taskConfig = _processReader.GetAltinnTaskExtension(taskId);
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, $"Error while fetching signing configuration for task {taskId}");
```
ivarne: This log message will be noisy and confusing if a ProcessElement is missing the TaskExtension tag, because this `ShouldRunForTask` runs on all tasks (also those that are not signature tasks)

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    public bool NoIncrementalValidation => true;
+
+    /// <summary>
+    /// Should never be called because NoIncrementalValidation is true.
+    /// </summary>
+    public Task<bool> HasRelevantChanges(IInstanceDataAccessor dataAccessor, string taskId, DataElementChanges changes)
+    {
+        throw new NotImplementedException();
```
ivarne: We can add the documentation as a message, so that it is easier to understand why this is not implemented.
```suggestion
        throw new UnreachableException("HasRelevantChanges should never be called because NoIncrementalValidation is true.");
```
This might need to import `using System.Diagnostics;`

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+    )
+    {
+        AltinnSignatureConfiguration? signingConfiguration = (
+            _processReader.GetAltinnTaskExtension(taskId)?.SignatureConfiguration
+        );
+
+        if (signingConfiguration == null)
+        {
+            _logger.LogError($"No signing configuration found for task {taskId}");
+            return [];
```
ivarne, line 90: This was checked by ShouldRunForTask, so throw here as well.

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
```diff
@@ -0,0 +1,159 @@
...
+        var cachedDataMutator = new InstanceDataUnitOfWork(
+            instance: dataAccessor.Instance,
+            _dataClient,
+            _instanceClient,
+            appMetadata,
+            _modelSerialization
+        );
+
+        var (getSigneeContextsError, signeeContexts) = await CatchError(
+            _signingService.GetSigneeContexts(cachedDataMutator, signingConfiguration)
```
ivarne: You should be able to get the signeeContexts from the `dataAccessor` instead of instantiating a `Mutator` in a validator.

ivarne: You would need to update the signatures in the signing service to use `IInstanceDataAccessor`

cammiida: The `GetSigneeContexts`-function also creates new contexts if data from different sources don't align, so it seems like it needs the mutator unfortunately :/

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
cammiida: Agreed

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
cammiida: Isn't `private` sufficient since it is even more scoped down than `internal`? 🤷‍♀️

src/Altinn.App.Core/Features/Validation/Default/SigningTaskValidator.cs
cammiida: Nm, I was wrong 🙃

PR #1053, opened 2025-01-22
Improve backwards compatibility for apps that use LayoutEvaluator.RemoveHiddenData

ivarne:
SSB complains that updating to 8.5.0 is not friction less, so I made a few changes to fix it

* Un[Obsolete] `LayoutEvaluatorStateInitializer.Init()`, as it still works and we don't have a good alternative finished yet.
* Move RowRemovalOption back to original namespace
* Create RemoveHiddenDataAsync and ensure that the old version is `Wait()`ed. (Code without mulitple datamodels is syncronous anyway)

### DIFF
--- a/src/Altinn.App.Core/Internal/Expressions/LayoutEvaluator.cs
+++ b/src/Altinn.App.Core/Internal/Expressions/LayoutEvaluator.cs
@@ -125,7 +125,7 @@ public static class LayoutEvaluator
     [Obsolete("Use the async version of this method RemoveHiddenDataAsync")]
     public static void RemoveHiddenData(LayoutEvaluatorState state, RowRemovalOption rowRemovalOption)
     {
-        RemoveHiddenDataAsync(state, rowRemovalOption).Wait();
+        RemoveHiddenDataAsync(state, rowRemovalOption).GetAwaiter().GetResult();
     }
 
     /// <summary>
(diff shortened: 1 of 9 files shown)


src/Altinn.App.Core/Internal/Expressions/LayoutEvaluator.cs
```diff
@@ -122,7 +122,16 @@ await HiddenFieldsForRemovalRecurs(
     /// <summary>
     /// Remove fields that are only referenced from hidden fields from the data object in the state.
     /// </summary>
-    public static async Task RemoveHiddenData(LayoutEvaluatorState state, RowRemovalOption rowRemovalOption)
+    [Obsolete("Use the async version of this method RemoveHiddenDataAsync")]
+    public static void RemoveHiddenData(LayoutEvaluatorState state, RowRemovalOption rowRemovalOption)
+    {
+        RemoveHiddenDataAsync(state, rowRemovalOption).Wait();
```
martinothamar: `GetAwaiter().GetResult()` instead? I think behavior around exceptions is better

martinothamar (approved):

src/Altinn.App.Core/Internal/Expressions/LayoutEvaluator.cs
ivarne: Seems to be that `.Wait()` wraps exceptions in AggregateExceptions, while `.GetAwaiter().GetResult()` unwraps the exception just like `await task` does. 
```suggestion
        RemoveHiddenDataAsync(state, rowRemovalOption).GetAwaiter().GetResult();
```

PR #1059, opened 2025-01-23
Mark Correspondence client and responses as `Experimental`

danielskovli:
## Description
Since the Correspondence API has not yet stabilised, this patch marks the client + methods, along with their responses as `experimental`. The request objects are not annotated, because we are not expecting any breaking changes to the _input_ for the Correspondence API.

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Altinn.App.Core.csproj
+++ b/src/Altinn.App.Core/Altinn.App.Core.csproj
@@ -5,7 +5,7 @@
     <Description>
       This class library holds all the core features used by a standard Altinn 3 App.
     </Description>
-    <NoWarn>Correspondence001</NoWarn>
+    <NoWarn>ALTINNAPP0200</NoWarn>
   </PropertyGroup>
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
--- a/src/Altinn.App.Core/Altinn.App.Core.csproj
+++ b/src/Altinn.App.Core/Altinn.App.Core.csproj
@@ -5,6 +5,7 @@
     <Description>
       This class library holds all the core features used by a standard Altinn 3 App.
     </Description>
+    <NoWarn>Correspondence001</NoWarn>
   </PropertyGroup>
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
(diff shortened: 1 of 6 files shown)


src/Altinn.App.Core/Altinn.App.Core.csproj
```diff
@@ -5,6 +5,7 @@
     <Description>
       This class library holds all the core features used by a standard Altinn 3 App.
     </Description>
+    <NoWarn>Correspondence001</NoWarn>
```
martinothamar: should we have some kind of namespacing in the name? To make clear it comes from Altinn app (Altinn)
I worked on some diagnostics a while back in the analyzer PR: https://github.com/Altinn/app-lib-dotnet/pull/651/files#diff-696f657d1537d15ab3195f2f7652e70e135f21299764a6ecd3da455f56d95349R5-R90

danielskovli: I put absolutely no effort behind this name here, so if you have a suggestion I would love to implement it

martinothamar: Convention is full uppercase and some abbreviation of the domain/context.... I would say `ALTINN` is too broad as well, maybe `ALTAPP` or `ALTINNAPP`

danielskovli: I vote `ALTINNAPP`. Four digits, start at 0001?

martinothamar: since we will use that range for lots of things probably, should probably be another range `0200` or something random. `ALTINNAPP0001` might be a more basic/foundational error if/when we get an analyzer

martinothamar: found this: https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/choosing-diagnostic-ids#considerations
sounds like we're OK

martinothamar (approved):

src/Altinn.App.Core/Altinn.App.Core.csproj
danielskovli: I read that too... I took it to mean "do what you want" 😂 

I'll start us at 0200 then 👌

PR #1064, opened 2025-01-24
Add back AssemblyVersionTarget

martinothamar:
## Description
Hunting down a bug, not sure if this is related..

Comparing the nupkgs of 8.5.1 vs 8.5.0:
image

The Assembly name version is not the same, which is bad regardless. I think the msbuild target I removed is the only thing that can affect that, and I'm not sure why this was not reproduced when I tested this locally

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -47,6 +47,14 @@
     <PackageId>$(MSBuildProjectName).Experimental</PackageId>
   </PropertyGroup>
 
+  <Target Name="AssemblyVersionTarget" AfterTargets="MinVer" Condition="'$(MinVerVersion)'!='' AND '$(BuildNumber)' != ''">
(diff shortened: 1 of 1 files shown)


martinothamar:
Running the same build commands for testing as in https://github.com/Altinn/app-lib-dotnet/pull/1044 and inspecting the resulting dlls in dnSpy (reloading assemblies after each run, commenting and uncommenting `AssemblyVersionTarget`), I see now that the target is needed for the right `AssemblyVersion`, which makes sense considering the MinVer docs...

I think DotPeek was tricking me with caching in https://github.com/Altinn/app-lib-dotnet/pull/1044

danielskovli (approved):

PR #1067, opened 2025-01-27
Multiple new expression functions

olemartinorg:
## Description

This is huge, and sorry about that, but the bulk of it is shared expression tests (luckily).

I implemented the following new expression functions:
- `countDataElements`
- `formatDate`
  - Not really that important on the backend, but this has a lot of shared tests now to make sure we're able to parse the same dates/times on both frontend and backend. The same parser is used in other date-related functions that must be working on the backend.
  - Has existed for a while on frontend, but I have a fix (coming to frontend) for it to make sure we parse it the same way as on the backend.
- `compare`
  - Including the operators `equals`, `greaterThan`, `greaterThanEq`, `lessThan` and `lessThanEq` - all of which had dedicated functions already
  - `isBefore`, `isBeforeEq`, `isAfter`, `isAfterEq` and `isSameDay` - these are all brand new, and relies on date parsing
- `stringIndexOf`, `stringReplace`, `stringSlice`, `lowerCaseFirst` and `upperCaseFirst`

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/pull/2961
- https://github.com/Altinn/app-frontend-react/issues/1299
- https://github.com/Altinn/app-frontend-react/issues/2122

## Verification
- [x] **Your** code builds clean without any errors or warnings
- [x] Manual testing done (required)
- [x] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [x] All tests run green

## Documentation
- [x] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)
  - https://github.com/Altinn/altinn-studio-docs/pull/1983

PR #1072, opened 2025-01-30
Introduce ExpressionTypeUnion to narrow down expression variables to json types

ivarne:
## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
```diff
@@ -163,7 +221,7 @@ LayoutEvaluatorState state
 
         if (targetContext is null)
         {
-            return null;
+            return new ExpressionTypeUnion();
```
martinothamar: Could we have `ExpressionTypeUnion.Null` instead to make it a little clearer? Since this is a public API

src/Altinn.App.Core/Internal/Expressions/ExpressionTypeUnion.cs
```diff
@@ -0,0 +1,286 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Altinn.App.Core.Internal.Expressions;
+
+/// <summary>
+/// Discriminated union for the JSON types that can be arguments and result of expressions
+/// </summary>
+[JsonConverter(typeof(ExpressionTypeUnionConverter))]
+public readonly struct ExpressionTypeUnion
```
martinothamar: I'm not sure about the name, how about just `ExpressionValue` instead? Then `ValueKind` could become `Kind`

martinothamar: Also, since it's public, we should probably implement `IEquatable<>` and related functions

src/Altinn.App.Core/Models/Expressions/ExpressionFunction.cs
```diff
@@ -1,11 +1,15 @@
...
 // ReSharper disable InconsistentNaming
+
 namespace Altinn.App.Core.Models.Expressions;
 
 /// <summary>
 /// Enumeration for valid functions in Layout Expressions
 ///
 /// Note that capitalization follows the JavaScript convention of camelCase for function names
 /// </summary>
+[JsonConverter(typeof(JsonStringEnumConverter))]
```
martinothamar, line 12: Is this safe to change? Isn't the default numeric serialization?

ivarne: Yes, the default is numeric. I don't think it matters, because I have a custom `[JsonConverter(typeof(ExpressionConverter))]` on Expression and that is the only type that uses this enum. It just seems like the correct annotation to add.

src/Altinn.App.Core/Internal/Expressions/ExpressionTypeUnion.cs
ivarne: Should the `IEquatable` implementations copy the expression behaviour of `"TruE" == true == "true"`, but `"Tru" != "tru"`?

ivarne: I like `ExpressionValue`

src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
ivarne: Yes, we should. The reason I ended up not adding that, is that a `readonly record struct` needs a parameterless constructor.

A sad fact is that it has two parameterless constructors. Both the one I'm required to implement and the default constructors for arrays that sets all the memory to `000000`.

src/Altinn.App.Core/Internal/Expressions/ExpressionTypeUnion.cs
martinothamar: I think it makes sense to match the behavior of `ExpressionEvaluator.EqualsImplementation`, whatever that is. The use of `StringComparison.Ordinal` has me a bit confused. Maybe it could be replaced with `==` instead of that custom impl at  some point

martinothamar (approved):

ivarne:
This is now part of #1067

PR #1073, opened 2025-01-30
Fix bug where datamodel field was removed when `RepeatingGroup` components' hidden expression differed 

martinothamar:
## Description
An `else` branch was missing in the `RepeatingGroup` case where 1 group is saying `hidden = true` and another is saying `hidden = false`

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Core/Internal/Expressions/LayoutEvaluator.cs
```diff
@@ -68,6 +68,18 @@ ComponentContext context
                 return;
             }
         }
+        else
```
bjosttveit: This code is a bit of a mess now so I have been trying to understand the intent of this. But as far as I remember, the `nonHiddenModelBindings` is supposed to "cancel out" the `hiddenModelBindings` if a different component bound to the same field is not hidden.

I remember adding the `includeHiddenRowChildren` for expression validation so that I could know every actual hidden field, whereas when we use it to remove hidden data, we don't need to explicitly remove "child fields" if the "parent" object is already getting removed.

The only thing I am sceptical about here is the `else`. Since this will fire if either `includeHiddenRowChildren == true || !isHidden`. But since my understanding is that it should "cancel" out the block above if another component is not hidden I would think it should only be ran when 
`!includeHiddenRowChildren && !isHidden` instead.

martinothamar: Removed the flag as discussed

bjosttveit (approved):

Issue #1103, opened 2025-02-13
`AuthorizationClient.GetUserRoles` doesn't handle 404 response (no roles)


martinothamar:
### Description of the bug

`AuthorizationClient.GetUserRoles` doesn't handle 404 responses : https://github.com/Altinn/app-lib-dotnet/blob/fdf12b5353397c19da3630062232504ccce2f8bd/src/Altinn.App.Core/Infrastructure/Clients/Authorization/AuthorizationClient.cs#L201

But localtest and authorization API will return 404 when no roles are found:
Localtest: https://github.com/Altinn/app-localtest/blob/72bc0f21bf9d4ebdd1d2d3b33c3a314fe58ff5cb/src/Controllers/Authorization/RolesController.cs#L49C55-L54
Authorization: https://github.com/Altinn/altinn-authorization/blob/1472402f3268755a9691b07ef541ca9a3bc50e81/src/Authorization/Controllers/RolesController.cs#L52-L55

### Steps To Reproduce

1. Use for example localtest and login with a user that has no roles.
2. In an app, use `IAuthenticationContext` and call `LoadDetails` which will try to fetch roles

PR #1104, opened 2025-02-13
Fix case where querying for roles for a user without any roles threw exception

martinothamar:
## Description
Authorization API and localtest gives 404 when there are no roles associated with a userID, partyID pair. 
I think this means we can't tell wether a 404 is due to misconfiguration (e.g. error in URL) or if there simply are no roles.
Reported by brg

## Related Issue(s)
- #1103

## Verification
- [ ] All tests run green

## Documentation


ivarne (approved):
Approved if a fix is required before we can update the apis to not return 404

PR #1120, opened 2025-02-20
Use StatusCode constants instead of casting HttpStatusCode enum to int

ivarne:
Microsoft.AspNetCore.Mvc StatusCodes are int constants which is prettier than casting the System.Net.HttpStatusCode enums to int when an int is required

### DIFF
--- a/src/Altinn.App.Api/Controllers/DataController.cs
+++ b/src/Altinn.App.Api/Controllers/DataController.cs
@@ -157,7 +157,7 @@ public class DataController : ControllerBase
         {
             return BadRequest(await GetErrorDetails(fileValidationError.UploadValidationIssues));
         }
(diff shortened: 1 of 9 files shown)


HauklandJ (approved):
🙌

src/Altinn.App.Api/Controllers/DataController.cs
```diff
@@ -1192,7 +1192,7 @@ private async Task<
                     Title = "Data Element Not Found",
                     Detail =
                         $"Did not find data element {dataElementGuid} on instance {instanceOwnerPartyId}/{instanceGuid}",
-                    Status = (int)HttpStatusCode.BadRequest,
+                    Status = StatusCodes.Status404NotFound,
```
martinothamar, line 1195: Late to the party, but suspect copy pasta error here

Issue #1084, opened 2025-02-04
Don't run validators for 'reject' action


cammiida:
### Description of the bug

When in a signing step, a user with write rights should have the possibility to "reject" the step. That is, go back (or wherever) to the relevant step configured in bpmn.
The user would typically reject in case of an error or if any of the signees didn't receive access to the form. 
The validator triggered checks if all signees have signed, which will never happen in the case outlined above.

@ivarne might have some thoughts about this? 

### Steps To Reproduce

1. Go to the signing step in [Signering Brukerstyrt](https://altinn.studio/editor/ttd/signering-brukerstyrt/overview).
2. Make sure you have write rights.
3. Try to click the reject button.
4. See that validation messages appear and that you haven't successfully rejected the step.

PR #1124, opened 2025-02-21
Don't run validators during process next if the action is reject.

bjorntore:
We decided to not run validation on process next when the action is reject as a quick fix to support aborting tasks that don't have valid data. Does this have unforeseen consequences?

> Note: We didn't validate that reject was a valid action for the task in BPMN, so adding that. Might break some apps, but I'd argue it's more a bugfix than a breaking change.

## Related Issue(s)
- #1084 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/ProcessController.cs
+++ b/src/Altinn.App.Api/Controllers/ProcessController.cs
@@ -774,19 +774,4 @@ public class ProcessController : ControllerBase
 
         return ExceptionResponse(e, defaultMessage);
     }
-
-    /// <summary>
-    /// Validates the selected action against the ones specified for the task in process.xml
-    /// </summary>
-    private bool IsActionAllowedForTask(string currentTaskId, string attemptedAction)
-    {
-        if (string.IsNullOrEmpty(attemptedAction))
-        {
-            return false;
-        }
-
-        AltinnTaskExtension? altinnTaskExtension = _processReader.GetAltinnTaskExtension(currentTaskId);
-
-        return altinnTaskExtension?.AltinnActions?.Select(x => x.Value).Contains(attemptedAction) is true;
-    }
 }
--- a/src/Altinn.App.Api/Controllers/ProcessController.cs
+++ b/src/Altinn.App.Api/Controllers/ProcessController.cs
@@ -309,7 +309,7 @@ public class ProcessController : ControllerBase
         {
             Instance instance = await _instanceClient.GetInstance(app, org, instanceOwnerPartyId, instanceGuid);
 
-            var currentTaskId = instance.Process.CurrentTask?.ElementId;
+            string? currentTaskId = instance.Process.CurrentTask?.ElementId;
 
             if (currentTaskId is null)
             {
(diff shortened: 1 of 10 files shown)


src/Altinn.App.Api/Controllers/ProcessController.cs
```diff
@@ -342,7 +342,23 @@
...
+                {
+                    return Conflict(
+                        new ProblemDetails()
+                        {
+                            Status = StatusCodes.Status409Conflict,
+                            Title = $"The action '{processNext.Action}' is not allowed for task '{currentTaskId}'!",
+                        }
+                    );
+                }
+            }
```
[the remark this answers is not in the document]
bjorntore: Not sure it becomes easier to read if nested. If my fellow humans think so, I'll change it.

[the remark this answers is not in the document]
ivarne: I'd say just move the `_processReader` call into the first if block, and the machine will approve.

Anyways, this code bock is not clear enough that I immediately understand what it is doing and why it is here (even when reviewing), so a comment along the lines of `// Validate selected action against the ones specified for the task in process.xml` would be nice.

Should we account for unicode shenanigans and use the string from bpmn instead of the user string in further processing?

cammiida (approved):
Syns dette gir veldig mening 🙌

martinothamar (approved):

ivarne (changes requested):
We need to look at existing apps to verify that all apps specify valid actions, or verify that such apps currently break, before merging this in a 8.x release.

Otherwise we risk breaking existing apps, in a pretty subtle way that they might not know to test before updating.

The obvious alternative is to mark an issue for v9 and only verify that `reject` is in bpmn when we consider ignoring validation issues.

src/Altinn.App.Core/Internal/Process/Interfaces/IProcessReader.cs
```diff
@@ -121,4 +121,9 @@ public interface IProcessReader
     /// If the element with the given elementId is a ProcessTask, this method will return the AltinnTaskExtension for that task. If it is not, null is returned.
     /// </summary>
     public AltinnTaskExtension? GetAltinnTaskExtension(string elementId);
+
+    /// <summary>
+    /// Checks if the action is included in the AltinnActions list for the given task, which means it is allowed for the task.
+    /// </summary>
+    public bool IsActionAllowedForTask(string taskId, string action);
```
ivarne: We could make this an extension method so that we don't clutter the interface with utilities that have 1 obvious composition from the public interface.

src/Altinn.App.Api/Controllers/ProcessController.cs
```diff
@@ -756,4 +774,19 @@ private ActionResult HandlePlatformHttpException(PlatformHttpException e, string
 
         return ExceptionResponse(e, defaultMessage);
     }
+
+    /// <summary>
+    /// Validates the selected action against the ones specified for the task in process.xml
+    /// </summary>
+    private bool IsActionAllowedForTask(string currentTaskId, string attemptedAction)
```
ivarne: This is no longer in use?

ivarne (changes requested):
Delete unused code, and consider using an extension method instead of extending the interface with a utility method. You can define the extension class in the same file as the interface.

src/Altinn.App.Api/Controllers/ProcessController.cs
bjorntore: Ah, tnx, forgot to delete.

ivarne (approved):

PR #1117, opened 2025-02-20
Update MimeTypeMap.cs to include audio/x-m4a for .m4a files

ivarne:
This is used by iPone uploads of audio files

Reported on slack https://altinn.slack.com/archives/C02EJ9HKQA3/p1739963353065429

PR #1005, opened 2025-01-02
Support getting page order from navigation groups

bjosttveit:
## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/pull/2831

## Verification
- [x] **Your** code builds clean without any errors or warnings
- [x] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #1126, opened 2025-02-24
Chore/release8.5 cherry-picks

ivarne:
Cherry picks for commits to main that are relevant for 8.5 releases

## Commits from
- #1067
- #1115
- #1120
- #1121
- #1005
- #1117


martinothamar:
Should https://github.com/Altinn/app-lib-dotnet/pull/1067 be here? That sounds like a feature to me

ivarne:
Yes, but it should not be held back by (or be used to accelerate) the release of 8.6

bjosttveit (approved):

Issue #1020, opened 2025-01-13
Confirm MaskinportenClient functionality


danielskovli:
## Description
Confirm that the new MaskinportenClient contains all required functionality from the legacy version (external nuget).

Eg. Can we use the new client for eFormidling and events, for instance?

## Issues found
The new client does not provide an implementation for [IMaskinportenTokenProvider](https://github.com/Altinn/app-lib-dotnet/blob/main/src/Altinn.App.Core/Internal/Maskinporten/IMaskinportenTokenProvider.cs#L11). This interface is used by [EformidlingStatusCheckEventHandler2.AddCompleteConfirmation](https://github.com/Altinn/app-lib-dotnet/blob/main/src/Altinn.App.Core/EFormidling/Implementation/EformidlingStatusCheckEventHandler2.cs#L165).

## Tasks
- [x] Provide an implementation for `IMaskinportenTokenProvider` that automatically gets injected via [AddMaskinportenClient](https://github.com/Altinn/app-lib-dotnet/blob/main/src/Altinn.App.Core/Features/Maskinporten/Extensions/ServiceCollectionExtensions.cs#L15)

danielskovli:
Afaik all functionality has now been confirmed and tested

PR #1142, opened 2025-02-27
Legacy implementation of `IMaskinportenTokenProvider`

danielskovli:
## Description
[EformidlingStatusCheckEventHandler2](https://github.com/Altinn/app-lib-dotnet/blob/26182f1e0d8428791008a6d2ddca830b420ca5c9/src/Altinn.App.Core/EFormidling/Implementation/EformidlingStatusCheckEventHandler2.cs#L44) currently requires an `IMaskinportenTokenProvider` implementation to operate. This makes eFormidling dependent on the external MaskinportenClient package and unable to switch over to the updated, internal, client.

This PR aims to rectify that.

## Notes
Practical testing for this is proving to be difficult. We currently don't have access to an eFormidling receiver in a test environment, which means we are unable to properly test for a "message received" status. We _can_, however, check that the authentication has succeeded. I'm exploring that option now.

## Related Issue(s)
- #1020 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Features/Maskinporten/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Altinn.App.Core/Features/Maskinporten/Extensions/ServiceCollectionExtensions.cs
@@ -1,5 +1,6 @@
 using Altinn.App.Core.Extensions;
 using Altinn.App.Core.Features.Maskinporten.Models;
+using Altinn.App.Core.Internal.Maskinporten;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 
@@ -30,6 +31,10 @@ internal static class ServiceCollectionExtensions
             MaskinportenClient.VariantInternal,
             (sp, key) => ActivatorUtilities.CreateInstance<MaskinportenClient>(sp, MaskinportenClient.VariantInternal)
         );
+
+#pragma warning disable CS0618 // Type or member is obsolete
+        services.TryAddTransient<IMaskinportenTokenProvider, LegacyMaskinportenTokenProvider>();
+#pragma warning restore CS0618 // Type or member is obsolete
 
         return services;
     }
--- a/src/Altinn.App.Core/Features/Maskinporten/LegacyMaskinportenTokenProvider.cs
+++ b/src/Altinn.App.Core/Features/Maskinporten/LegacyMaskinportenTokenProvider.cs
@@ -1,0 +1,29 @@
+using Altinn.App.Core.Internal.Maskinporten;
+
+namespace Altinn.App.Core.Features.Maskinporten;
+
+#pragma warning disable CS0618 // Type or member is obsolete
+
+/// <summary>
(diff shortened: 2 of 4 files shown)


src/Altinn.App.Core/Features/Maskinporten/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -31,6 +32,10 @@ public static IServiceCollection AddMaskinportenClient(this IServiceCollection s
             (sp, key) => ActivatorUtilities.CreateInstance<MaskinportenClient>(sp, MaskinportenClient.VariantInternal)
         );
 
+#pragma warning disable CS0618 // Type or member is obsolete
+        services.TryAddTransient<IMaskinportenTokenProvider, LegacyMaskinportenTokenProvider>();
```
ivarne, line 36: So because of the usage of `TryAddTransient` this Legacy provider will only be registered if no `IMaskinportenTokenProvider` is registered before this one. If another provider is registered after this, it will override (unless someone gets `IEnumerable<IMaskinportenTokenProvider>` which seems unlikely).

This registration also only happens if a `IMaskinportenClient` is registered, so it will work.

ivarne (approved):
The code and setup looks good to me

src/Altinn.App.Core/Features/Maskinporten/Extensions/ServiceCollectionExtensions.cs
danielskovli: Yep, that was the idea. If someone wants to implement the legacy version (`AddMaskinportenJwkTokenProvider`), that should be allowed. Also if someone for whatever reason wants to override by placing their registration after ours, that's cool too.

I'm only aiming to have a default implementation available if nothing else is offered.

> This registration also only happens if a IMaskinportenClient is registered
This is correct, but we are automatically registering the `IMaskinportenClient` in `AddAltinnAppServices`. However, if no other implementation is attempting to inject either the `IMaskinportenClient` or the `IMaskinportenTokenProvider`, they will never be actualised.

src/Altinn.App.Core/Features/Maskinporten/LegacyMaskinportenTokenProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+    private readonly IMaskinportenClient _maskinportenClient;
+
+    public LegacyMaskinportenTokenProvider(IMaskinportenClient maskinportenClient)
+    {
+        _maskinportenClient = maskinportenClient;
+    }
+
+    public async Task<string> GetToken(string scopes)
+    {
+        return await _maskinportenClient.GetAccessToken([scopes]);
```
martinothamar, line 22: is this meant to be only 1 scope or, does `scopes` need to be split by whitespace before put into an array?

danielskovli: We have tests for this already, where sending "scope1 scope2 scope3" as a single string works as intended. It's a bit weird looking, but it's only because we wanted an IEnumerable<string> interface for the new client. At the end of the day the client literally does a string.Join() on that collection.

PR #1161, opened 2025-03-05
Authorize process next as 'sign' when the process task type is signing

bjorntore:
Authorize process next as 'sign' when the process task type is signing and and no action was submitted. This is how storage authorizes it, and it means that the party that should run process next when everyone has signed, needs the 'sign' access right.

Today, before we release the new signing stuff, the process next in a signature step will always have action=sign, so this will only have effect for our new functionality.

This might enable automatic process next after the last signee has signed (successful validation of signing step).

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


bjorntore:
After discussing with Camilla and Ronny we've decided to try to make so that it's enough to have sign or write to run process next in a signing step.

HauklandJ (approved):

cammiida (commented):
LGTM 👌

PR #1168, opened 2025-03-06
Improve token generation and test parameterization for TestAuthentication

martinothamar:
## Description

* Generate systemuser token using `JwtPayload` so we get correct serialization
  * Upon inspecting the systemuser tokens, the `consumer` and `authorization_details` claims where strings with JSON content, but are now plain JSON object claims instead, which is the correct representation. `Claim`'s can only hold string values 
* Make `TestJwtToken` impelement `IXunitSerializable` so that tooling/test explorer sees different test items per parameter value
* Refactor `Authenticated`, use `JwtPayload` instead of `Claim` when parsing/reading token

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Configuration/GeneralSettings.cs
+++ b/src/Altinn.App.Core/Configuration/GeneralSettings.cs
@@ -41,6 +41,8 @@ public class GeneralSettings
 
     internal bool DisableAppConfigurationCache { get; set; }
 
+    internal bool IsTest { get; set; }
+
     /// <summary>
     /// The externally accesible base url for the app with trailing /
     /// </summary>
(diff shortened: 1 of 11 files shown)


danielskovli (approved):
LGTM! 

I'm a bit lost on the use of `ref` for `ParseContext`, I can't see any mutations that require this... but perhaps related to all the action going on in `TryAssign` 🤷

martinothamar:
I just did it to not pay for copies on each call, since the struct is pretty big. I would have used `in` if the struct was immutable

danielskovli:
Living and learning... 🤓

PR #1183, opened 2025-03-12
Ignore 'Info.Version' in custom OpenAPI doc snapshot test

martinothamar:
## Description
Ignore the 'Info.Version' field in the custom OpenAPI doc snapshot test. The `AltinnNugetVersion` was giving slightly different results based on GH actions context (release vs normal build)
Also bumps the OpenAPI versjon itself.

## Related Issue(s)
- N/A

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #1186, opened 2025-03-13
Verify the actual response from api, instead of the parsed and potentially modified response

ivarne:
Follow up from https://github.com/Altinn/app-lib-dotnet/pull/1183#issuecomment-2717969882

Use .ScrubMember("version") to remove the nuget version from the info section

Use Microsoft.OpenApi.Readers only to verify that the package don't have validations that break.

Filter out the path parameter `{dataType}` that is used in the custom swagger file to allow annotation of data type schema

### DIFF
--- a/test/Altinn.App.Api.Tests/OpenApi/OpenApiSpecChangeDetection.cs
+++ b/test/Altinn.App.Api.Tests/OpenApi/OpenApiSpecChangeDetection.cs
@@ -1,10 +1,5 @@
-using System.Globalization;
-using Altinn.App.Api.Controllers;
 using Argon;
 using Microsoft.AspNetCore.Mvc.Testing;
-using Microsoft.OpenApi;
-using Microsoft.OpenApi.Extensions;
-using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Readers;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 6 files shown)


test/Altinn.App.Api.Tests/OpenApi/OpenApiSpecChangeDetection.cs
```diff
@@ -32,33 +30,26 @@ public async Task SaveCustomOpenApiSpec()
...
-        {
-            VerifySettings settings = new();
-            settings.UseStrictJson();
-            settings.DontScrubGuids();
-            settings.DontIgnoreEmptyCollections();
-            settings.AddExtraSettings(settings => settings.MetadataPropertyHandling = MetadataPropertyHandling.Ignore);
-            return settings;
-        }
+        await VerifyJson(openApi)
+            .ScrubMember("version")
```
martinothamar, line 49: Doesn't this scrub _any_ field called `version`? That sounds like something we should avoid

ivarne: Yes, but there isn't any other "version": fields in the open api spec, and if there ever was one, it would show up in the diff we accept.

PR #1199, opened 2025-03-20
Fix flakyness in tests caused by App Insights 'HostingDiagnosticListener'

martinothamar:
## Description
Some `HomeControllerTest_SetQueryParams` tests are snapshotting API responses including headers.
Application Insights will by default inject a `Request-Context` header in API responses through its [HostingDiagnosticListener](https://github.com/microsoft/ApplicationInsights-dotnet/blob/18ffbf6427ee705179052bfc94356c593621db85/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs#L708) which runs globally in process. So if these tests run parallel to other tests which used the App Insights SDK, this would trigger the linked function to append the response header. I'm not sure why it didn't do it every time, maybe it's related to sampling

This PR just disabled response injection during tests, as it is not important...

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/test/Altinn.App.Api.Tests/Program.cs
+++ b/test/Altinn.App.Api.Tests/Program.cs
@@ -19,6 +19,7 @@ using Altinn.App.Core.Internal.Profile;
 using Altinn.App.Core.Internal.Registers;
 using AltinnCore.Authentication.JwtCookie;
 using App.IntegrationTests.Mocks.Services;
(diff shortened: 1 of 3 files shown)


adamhaeger (approved):

adamhaeger (approved):
lgtm

PR #1219, opened 2025-03-30
Fix broken build

ivarne:
image

The build is unreliable, so I'm not sure these issues will fix them, but in my limited testing I see the issue without the fix and not with the fix.

* Don't reference CS files outside of the folder (copy ModuleInitializer into every project).
* Move global usings for test projects to the individual test projects
  * The current layout means that the build fails when adding a new test project before referencing Altinn.App.Tests.Common

### DIFF
--- a/test/Altinn.App.Api.Tests/Altinn.App.Api.Tests.csproj
+++ b/test/Altinn.App.Api.Tests/Altinn.App.Api.Tests.csproj
@@ -27,6 +27,11 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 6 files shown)


martinothamar:
Since it looks like you are able to reproduce, do you want to try out https://github.com/Altinn/app-lib-dotnet/pull/1220?

martinothamar (approved):
Don't think this fixes the copy error build weirdness, but looks fine either way 👍 

> The current layout means that the build fails when adding a new test project before referencing Altinn.App.Tests.Common

Daniel immediately hit this 😄

PR #1215, opened 2025-03-27
Cleanup test projects setup

martinothamar:
## Description
* `Altinn.App.Common.Tests` is just a library for common code and resources, e.g. fixtures (no tests)
  * Add namespaces to Directory.Build.props 
* Remove `Core.Tests` -> `Api.Tests` reference, moved shared code to `Common.Tests` 
* Remove some properties that I suspected weren't needed
* Remove FluentAssertions in Common.Tests

`Common.Tests` is perhaps a little weird naming considering its not holding tests 

## Related Issue(s)
- N/A

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #1220, opened 2025-03-30
Fix content include for testdata files in Altinn.App.Core.Tests which made build buggy

martinothamar:
## Description
From the error messages I've seen thus far for failing builds after merging #1215, the build has failed to copy files where `TestData` was in the path within `bin` when building `Altinn.App.Core.Tests`, so I think my "simplified" include is the culprit. VS Code solution explorer also showed the bin folder disappearing after adding this exclude clause, so this gives me some confidence.

Of course I haven't been able to reproduce this error on my machines, so hopefully someone else can test this 

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation


ivarne:
This looks promising, but I think we should use `[CallerFilePath]` instead of copying every test asset into the build folder (or add `../../` as we do in some places). As long as we use VerifyTests we can't package the tests for running in a different environment anyway.

ivarne (approved):

PR #1226, opened 2025-03-31
Translate CustomTextKey to fill description/message in validation responses

bjosttveit:
## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/812

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -22,6 +22,7 @@ using Altinn.App.Core.Internal.Prefill;
 using Altinn.App.Core.Internal.Process;
 using Altinn.App.Core.Internal.Profile;
 using Altinn.App.Core.Internal.Registers;
+using Altinn.App.Core.Internal.Texts;
 using Altinn.App.Core.Models;
 using Altinn.App.Core.Models.Process;
 using Altinn.App.Core.Models.Validation;
@@ -69,6 +70,7 @@ public class InstancesController : ControllerBase
     private readonly IHostEnvironment _env;
     private readonly ModelSerializationService _serializationService;
     private readonly InternalPatchService _patchService;
+    private readonly ITranslationService _translationService;
     private readonly InstanceDataUnitOfWorkInitializer _instanceDataUnitOfWorkInitializer;
 
     private const long RequestSizeLimit = 2000 * 1024 * 1024;
@@ -93,6 +95,7 @@ public class InstancesController : ControllerBase
         IHostEnvironment env,
         ModelSerializationService serializationService,
         InternalPatchService patchService,
+        ITranslationService translationService,
         IServiceProvider serviceProvider
     )
     {
@@ -113,6 +116,7 @@ public class InstancesController : ControllerBase
         _env = env;
         _serializationService = serializationService;
         _patchService = patchService;
+        _translationService = translationService;
         _instanceDataUnitOfWorkInitializer = serviceProvider.GetRequiredService<InstanceDataUnitOfWorkInitializer>();
     }
 
@@ -309,7 +313,11 @@ public class InstancesController : ControllerBase
         InstantiationValidationResult? validationResult = await instantiationValidator.Validate(instanceTemplate);
         if (validationResult != null && !validationResult.Valid)
         {
-            return StatusCode(StatusCodes.Status403Forbidden, validationResult);
+            InstantiationValidationResult validationResultWithMessage = await ValidationResultWithMessage(
+                validationResult,
+                language
+            );
+            return StatusCode(StatusCodes.Status403Forbidden, validationResultWithMessage);
         }
 
         instanceTemplate.Org = application.Org;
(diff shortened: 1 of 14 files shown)


src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -1313,4 +1333,23 @@ await _instanceClient.UpdateDataValues(
             );
         }
     }
+
+    private async Task<InstantiationValidationResult> ValidationResultWithMessage(
```
ivarne: Shouldn't the method name contain the word "translate", as it uses the `_translationService`?

src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -1313,4 +1333,23 @@ await _instanceClient.UpdateDataValues(
...
+            return validationResult;
+        }
+
+        return new InstantiationValidationResult()
+        {
+            Valid = validationResult.Valid,
+            CustomTextKey = validationResult.CustomTextKey,
+            Message = await _translationService.TranslateTextKey(validationResult.CustomTextKey, language),
+            ValidParties = validationResult.ValidParties,
+        };
```
ivarne: I don't think it makes much sense to create a new object here. Immutable updates isn't worth a whole lot (when we don't use the object id for fast compare), and we will get funny behaviour if someone adds a new property without adding it to the mapping.

We could change ``InstantiationValidationResult`` to a record so we can use the ``with`` keyword if we need the immutable updates.

```c#
return validationResult with
{
    Message = await _translationService.TranslateTextKey(validationResult.CustomTextKey, language),
}
```

src/Altinn.App.Core/Internal/Texts/TranslationService.cs
```diff
@@ -0,0 +1,55 @@
...
+using Altinn.App.Core.Internal.Language;
+using Altinn.App.Core.Models;
+using Altinn.Platform.Storage.Interface.Models;
+
+namespace Altinn.App.Core.Internal.Texts;
+
+/// <summary>
+/// Translation service
+/// </summary>
+public class TranslationService : ITranslationService
```
ivarne: New code style
```suggestion
internal class TranslationService : ITranslationService
```

src/Altinn.App.Core/Internal/Validation/ValidationService.cs
```diff
@@ -163,6 +169,35 @@ public async Task<List<ValidationSourcePair>> ValidateIncrementalFormData(
         return lists.OfType<ValidationSourcePair>().ToList();
     }
 
+    private async Task<List<ValidationIssue>> ValidationIssuesWithDescription(
```
ivarne: Function name should contain translate (from TranslationService)

src/Altinn.App.Core/Models/Validation/InstantiationValidationResult.cs
```diff
@@ -12,6 +12,11 @@ public class InstantiationValidationResult
     /// </summary>
     public bool Valid { get; set; }
 
+    /// <summary>
+    /// The custom text key to use for the localized text.
+    /// </summary>
```
ivarne, line 17: Mention in summary that it is used for translation in backend if message is null, and in frontend anyway.

src/Altinn.App.Core/Internal/Validation/ValidationService.cs
```diff
@@ -163,6 +169,35 @@ public async Task<List<ValidationSourcePair>> ValidateIncrementalFormData(
...
+                Description =
+                    String.IsNullOrEmpty(issue.Description) && !String.IsNullOrEmpty(issue.CustomTextKey)
+                        ? await _translationService.TranslateTextKey(issue.CustomTextKey, language)
+                        : issue.Description,
+                CustomTextKey = issue.CustomTextKey,
+                CustomTextParams = issue.CustomTextParams,
+            })
+        );
+
+        return issuesWithDescription.ToList();
```
ivarne: I think this should be a loop that mutates the issues instead. Translation doesn't typically involve slow http calls, so I don't think trying to do translation in parallel is worth it.

ivarne (commented):
Looks good, but a few comments.

This `ITranslationService` is limited and can't implement correct variable replacement (that will need a `IInstanceDataAccessor`), but it should probably at least look at `ValidationIssue.CustomTextParams`.

ivarne:
@bjosttveit Do we merge this without replacement of `CustomTextParams` (because it is broken in frontend), or do we put this on hold until we know how to handle them?

bjosttveit:
For the specific [issue](https://github.com/Altinn/app-lib-dotnet/issues/812) its probably enough with just plain text resources, so we could always enhance it later. But it would be nice to finally deal with custom parameters.

ivarne (approved):
Looks mostly good.

We know that this interface won't be the final version when we decide to support replacements properly. Could we signal that in some way?

As far as I know (now) the correct method would be
```c#
    Task<string> TranslateTextKey(
        string key,
        string? language,
        IInstanceDataAccessor? accessor = null,
        string? taskId = null,
        List<string>? customTextParams = null);
```

Adding optional parameters is fine unless a service owner decides to implement the interface, so we might be good just shipping this now.

PR #1228, opened 2025-04-02
Unobsolete GET instance copy endpoint, copy based on `Authenticated`

martinothamar:
## Description
* Unobsolete GET instance copy endpoint that is going to be used by Arbeidsflate
* Do the auth checks based on `Authenticated`

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Controllers/InstancesController.cs
+++ b/src/Altinn.App.Api/Controllers/InstancesController.cs
@@ -11,6 +11,7 @@ using Altinn.App.Core.Configuration;
 using Altinn.App.Core.Constants;
 using Altinn.App.Core.Extensions;
 using Altinn.App.Core.Features;
+using Altinn.App.Core.Features.Auth;
 using Altinn.App.Core.Helpers;
 using Altinn.App.Core.Helpers.Serialization;
 using Altinn.App.Core.Internal.App;
@@ -73,7 +74,7 @@ public class InstancesController : ControllerBase
     private readonly InternalPatchService _patchService;
     private readonly ITranslationService _translationService;
     private readonly InstanceDataUnitOfWorkInitializer _instanceDataUnitOfWorkInitializer;
-
+    private readonly IAuthenticationContext _authenticationContext;
     private const long RequestSizeLimit = 2000 * 1024 * 1024;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -597,7 +600,8 @@ await _processEngine.HandleEventsAndUpdateStorage(
 
     /// <summary>
     /// This method handles the copy endpoint for when a user wants to create a copy of an existing instance.
-    /// The endpoint will primarily be accessed directly by a user clicking the copy button for an archived instance.
+    /// The endpoint will primarily be accessed directly by a user clicking the copy button for an archived instance
+    /// from the message box in the Altinn 2 portal/Altinn 3 arbeidsflate.
```
ivarne, line 624:
```suggestion
    /// from the message box in the Altinn 2 portal/Altinn 3 arbeidsflate.
    /// WARNING: The endpoint will be removed in an updated version of the app packages as soon as we have an
    /// alternative where dialogporten/arbeidsflate where it doesn't need to rely on the user cookie propagating 
    /// through a GET request.
```

martinothamar: There's discussions going on about the future of Altinn authentication (where/how to store session info, not doing token exchange etc etc) but there is nothing concrete yet, so I don't think it's appropriate to make any assertions on what we'll do at that point. Whatever it is, it has to be planned and agreed upon with the Arbeidsflate team

ivarne: The important part is that the endpoint might be removed without warning when arbeidsflate has an alternative and stops using it for apps newer than a specific version (or however we decide to do the transition)

martinothamar: Is your point that this is only meant to be consumed by Arbeidsflate? And that we don't care about potential breakage for any other possible consumer? If so I agree that we should document that someplace, but I don't think it belongs in an XML doc comment that doesn't go anywhere

martinothamar: What do you think about the proposed changes here? https://github.com/Altinn/altinn-studio-docs/pull/2226

adamhaeger (approved):

PR #1229, opened 2025-04-03
Add party object to instance owner in APIs returning instance

martinothamar:
## Description
This is to enable frontend in using information about the party in e.g. appheader.

APIs:
* Create instance
* Get instance

APIs not touched
* Patching
* User actions
* ...etc

## Related Issue(s)
-  https://github.com/Altinn/altinn-support-private/issues/4177

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Api/Models/InstanceDto.cs
+++ b/src/Altinn.App.Api/Models/InstanceDto.cs
@@ -1,0 +1,189 @@
+#nullable disable
+using Altinn.Platform.Register.Models;
+using Altinn.Platform.Storage.Interface.Models;
+using Newtonsoft.Json;
+
+namespace Altinn.App.Api.Models;
+
+/// <summary>
+/// Represents the response from an API endpoint providing a list of key-value properties.
+/// </summary>
+[JsonObject(ItemNullValueHandling = NullValueHandling.Ignore)]
+public sealed class InstanceDto
+{
+    /// <summary>
+    /// Gets or sets the unique id of the instance {instanceOwnerId}/{instanceGuid}.
+    /// </summary>
+    [JsonProperty(PropertyName = "id")]
+    public required string Id { get; init; }
+
+    /// <summary>
+    /// Gets or sets the instance owner information.
+    /// </summary>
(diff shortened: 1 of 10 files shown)


src/Altinn.App.Api/Models/InstanceDto.cs
```diff
@@ -0,0 +1,189 @@
...
+using Altinn.Platform.Storage.Interface.Models;
+using Newtonsoft.Json;
+
+namespace Altinn.App.Api.Models;
+
+/// <summary>
+/// Represents the response from an API endpoint providing a list of key-value properties.
+/// </summary>
+[JsonObject(ItemNullValueHandling = NullValueHandling.Ignore)]
+public sealed class InstanceDto
```
HauklandJ: We are leaning heavily towards using "response" instead of "dto"

HauklandJ (approved):

src/Altinn.App.Api/Models/InstanceDto.cs
```diff
@@ -0,0 +1,189 @@
...
+using Altinn.Platform.Register.Models;
+using Altinn.Platform.Storage.Interface.Models;
+using Newtonsoft.Json;
+
+namespace Altinn.App.Api.Models;
+
+/// <summary>
+/// Represents the response from an API endpoint providing a list of key-value properties.
+/// </summary>
+[JsonObject(ItemNullValueHandling = NullValueHandling.Ignore)]
```
ivarne: Do we need to copy the Newtonsoft annotations over here?

src/Altinn.App.Api/Models/InstanceDto.cs
martinothamar: Ooo I looked at `UserDefinedMetadata` or something I think, I can go with response

src/Altinn.App.Api/Models/InstanceDto.cs
martinothamar: Nope I don't think so, I can remove them

PR #1231, opened 2025-04-03
Validate after running user action handler in process next

bjorntore:
Validate after running user action handler in process next, so that signature can be generated before validation.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


HauklandJ (commented):
LGTM 👍

ivarne (changes requested):
Looks good, but we should enforce the order in a webapplicationfactory test to ensure that we preserve behavior in the comming rewrite.

1. action runs when validation fails
2. validation doesn't run if action fails

bjorntore:
Adding integration test in https://github.com/Altinn/app-lib-dotnet/issues/1237.

bjorntore:
Need to get it into preview.

PR #1235, opened 2025-04-04
persist correspondence id in signeestate

HauklandJ:
## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


bjorntore (approved):
As agreed, store the receipt id as well.

HauklandJ:
> As agreed, store the receipt id as well.

This happens in the user action, where the SigneeContext is not used, rather the SignatureContext. The receipt is also sent after the sign persistence call. I chose to log it for now, and removed the IsReceipt sent property from SigneeState as it has never been used

PR #1239, opened 2025-04-09
Feature/ Specific token scope for IUserTokenProvider

olamathiesenBlueTree:
We are using the altinn platform on a project and we are in need of using the clients that use IUserTokenProvider with other tokens than the current logged in user's token. 

## Description
Added another implementation of IUserTokenProvider, the SpecificTokenProvider, added a SpecificTokenProviderStateContext which sets and unsets a token state within a TokenScope which inherits IDisposable. 

The SpecificTokenProviderStateContext is using the AsyncLocal<T> to store the state. 
This is the same way the default HttpContextAccessor is doing with HttpContext. So that SpecificTokenProviderStateContext can be registered as Singleton in DI.

The DI extensions unregisters the IUserTokenProvider if its already registers and registers the SpecificTokenProvider as it implementation. Can also register with a custom IUserTokenProvider to get the token if Token state is not set. 
By default it re-registers the UserTokenProvider as itself and uses that inside.

Usage can be seen in tests or image:
image

## Related Issue(s)
- #{issue number}

## Verification
- [x] All tests run green

## Documentation


danielskovli:
@olamathiesenBlueTree 

First off, please accept our apology at the extreme delay in getting back to you on this one. The `AsyncLocal` concept you propose here has been discussed at length internally, but ultimately we decided on a simplified approach that requires less contextual awareness between services: https://github.com/Altinn/app-lib-dotnet/pull/1425

Currently, only the `DataClient` has been given support for the new `AuthenticationMethod` types. If you would like to contribute a pull request that adds this support to any of the other clients, for instance the `AltinnPartyClient`, please go ahead.

If this is something you're interested in doing, and would like some help to get started and/or implementation advice, don't hesitate to [reach out](https://digdir-samarbeid.slack.com/).

PR #1241, opened 2025-04-09
Remove self identified users as a special case in `Authenticated`

martinothamar:
## Description
There's less of a difference between users and self identified users than I originally thought

* Self identified users can be thought of as users where auth level = 0
* Self identified users can't have roles right now, but might in the future (so they can submit "on behalf of" orgs at some point, and there's nothing stopping them in localtest currently)

The only thing brought over from `SelfIdentified` is the username claim which is only there for selfgregistered users.

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation


cammiida (approved):
LGTM!

PR #1247, opened 2025-04-11
Preserve empty issue list from validation on UserActions

ivarne:
The user action endpoint returns validation issues grouped by dataElementId (it kind of made sense when issues always was tied to a single element) For backwards compatibility we kept the grouping when working with validation issues

In the first implementation there was a bug that we didn't add an empty list of issues to the response when an validator returned an empty list of issues.

This mostly works fine, unless a validator removes the last issue, wich the frontend will think is still valid.

## Verification
- [ ] All tests run green

## Documentation


bjosttveit (approved):
LGTM

PR #1261, opened 2025-04-24
Convert ISigneeProvider to use ShouldRunForTask pattern as IValidator

bjorntore:
Is it better? Is it worse?

Changes in the app:
https://altinn.studio/repos/ttd/signering-brukerstyrt/pulls/2/files

### DIFF
--- a/src/Altinn.App.Core/Features/Signing/ISigneeProvider.cs
+++ b/src/Altinn.App.Core/Features/Signing/ISigneeProvider.cs
@@ -10,9 +10,17 @@ namespace Altinn.App.Core.Features.Signing;
 public interface ISigneeProvider
 {
     /// <summary>
-    /// Used to select the correct <see cref="ISigneeProvider" /> implementation for a given signing task.
+    /// The task id for the task that the signee provider is associated with or "*" if the signee provider should be available for all tasks
     /// </summary>
-    public string Id { get; init; }
+    /// <remarks>Ignored if <see cref="ShouldRunForTask"/> is implemented</remarks>
+    public string TaskId { get; }
+
+    /// <summary>
+    /// Check if this signee provider should be available for the given task.
+    ///
+    /// Default implementations check <see cref="TaskId"/>
+    /// </summary>
+    public bool ShouldRunForTask(string taskId) => TaskId == "*" || TaskId == taskId;
 
     /// <summary>
     /// Returns a list of signees for the current signing task.
--- a/src/Altinn.App.Core/Features/Signing/Services/SigneeContextsManager.cs
+++ b/src/Altinn.App.Core/Features/Signing/Services/SigneeContextsManager.cs
@@ -58,11 +58,7 @@ internal sealed class SigneeContextsManager : ISigneeContextsManager
         Instance instance = instanceDataMutator.Instance;
         string taskId = instance.Process.CurrentTask.ElementId;
 
-        SigneeProviderResult? signeesResult = await GetSigneesFromProvider(instance, signatureConfiguration);
-        if (signeesResult is null)
-        {
-            throw new SigningException("No signees returned from the signees provider");
-        }
+        SigneeProviderResult signeesResult = await GetSigneesFromProvider(instance, signatureConfiguration);
 
         List<SigneeContext> signeeContexts = [];
         foreach (ProvidedSignee signeeParty in signeesResult.Signees)
--- a/src/Altinn.App.Core/Features/Signing/Services/SigneeContextsManager.cs
+++ b/src/Altinn.App.Core/Features/Signing/Services/SigneeContextsManager.cs
@@ -61,7 +61,7 @@ internal sealed class SigneeContextsManager : ISigneeContextsManager
         SigneeProviderResult? signeesResult = await GetSigneesFromProvider(instance, signatureConfiguration);
         if (signeesResult is null)
         {
-            return [];
+            throw new SigningException("No signees returned from the signees provider");
         }
 
         List<SigneeContext> signeeContexts = [];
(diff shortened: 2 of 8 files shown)


HauklandJ:
So, instead of using the signatureConfig configuration for a task:
```diff
    <bpmn:task id="SigningTask_Founders" name="Signering stiftere">
      <bpmn:extensionElements>
        <altinn:taskExtension>
          <altinn:taskType>signing</altinn:taskType>
          <altinn:actions>
            <altinn:action>sign</altinn:action>
            <altinn:action>reject</altinn:action>
          </altinn:actions>
          <altinn:signatureConfig>
            <altinn:dataTypesToSign>
              <altinn:dataType>ref-data-as-pdf</altinn:dataType>
              <altinn:dataType>Stiftelsesdokumenter</altinn:dataType>
            </altinn:dataTypesToSign>
            <altinn:signatureDataType>signatures-founders</altinn:signatureDataType>
            <altinn:signeeStatesDataTypeId>signeeStates-founders</altinn:signeeStatesDataTypeId>
-           // Here I register the implementation of the interface that I want for the "SigningTask_Founders" task:
-           <altinn:signeeProviderId>founders</altinn:signeeProviderId>
            <altinn:signingPdfDataType>signatures-founders-pdf</altinn:signingPdfDataType>
            <altinn:correspondenceResource>apps-correspondence-integrasjon2</altinn:correspondenceResource>
            <altinn:runDefaultValidator>true</altinn:runDefaultValidator>
          </altinn:signatureConfig>
        </altinn:taskExtension>
      </bpmn:extensionElements>
      <bpmn:incoming>SequenceFlow_1oot28q</bpmn:incoming>
      <bpmn:outgoing>Flow_1if1sh9</bpmn:outgoing>
    </bpmn:task>
    <bpmn:task id="SigningTask_Auditor" name="Signering revisor">
      <bpmn:extensionElements>
        <altinn:taskExtension>
          <altinn:taskType>signing</altinn:taskType>
          <altinn:actions>
            <altinn:action>sign</altinn:action>
            <altinn:action>reject</altinn:action>
          </altinn:actions>
          <altinn:signatureConfig>
            <altinn:dataTypesToSign>
              <altinn:dataType>ref-data-as-pdf</altinn:dataType>
            </altinn:dataTypesToSign>
            <altinn:signeeStatesDataTypeId>signatures-auditor</altinn:signeeStatesDataTypeId>
            <altinn:signatureDataType>signeeStates-auditor</altinn:signatureDataType>
-           // Here I register the implementation of the interface that I want for the "SigningTask_Auditor" task:
-           <altinn:signeeProviderId>auditor</altinn:signeeProviderId>
            <altinn:signingPdfDataType>signatures-founders-pdf</altinn:signingPdfDataType>
            <altinn:correspondenceResource>apps-correspondence-integrasjon2</altinn:correspondenceResource>
          </altinn:signatureConfig>
        </altinn:taskExtension>
      </bpmn:extensionElements>
      <bpmn:incoming>Flow_0mj904m</bpmn:incoming>
      <bpmn:outgoing>Flow_1miwjt2</bpmn:outgoing>
    </bpmn:task>
```

I now tell the implementation of the interface which task it should run for:

App/logic/FounderSigneesProvider:

```diff
public class FounderSigneesProvider(IDataClient dataClient) : ISigneeProvider
{
    private readonly IDataClient _dataClient = dataClient;

-   public string Id { get; init; } = "founders";
+   public string TaskId { get; init; } = "SigningTask_Founders";

    public async Task<SigneeProviderResult> GetSigneesAsync(Instance instance)
    {
        ...
    }
}
```

App/logic/AuditorSigneesProvider:

```diff
public class AuditorSigneesProvider(IDataClient dataClient) : ISigneeProvider
{
    private readonly IDataClient _dataClient = dataClient;

-   public string Id { get; init; } = "auditor";
+   public string TaskId { get; init; } = "SigningTask_Auditor";

    public async Task<SigneeProviderResult> GetSigneesAsync(Instance instance)
    {
        ...
    }
}
```

Pros: 
- Same implementation as validation
- ?
Cons:
- ?

bjorntore:
Struggling a bit with deciding what's best.

This relation would likely have been modeled in SQL by having the foreign key on the task, not the provider. That would enforce that there is always none or one provider for a task: `task *-0..1 provider`. We would also see in BPMN what provider is being used, at a glance.

The other variant is a recognizable pattern from IValidator, which is a pro, but the patterns main strength is supporting any number of implementations, that are all invoked?


@martinothamar Any comments on this?

bjorntore:
@RonnyB71 brought up a point: Service tasks. 

They will likely look like something like this in BPMN. 

Here, you tell us the ID of the `IServiceTask` implementation you want us to execute, it's just called taskType instead of ID. This is a similar case, where we need you to give us exactly one specific implementation to call.
```
<bpmn:serviceTask id="eFormidlingTask" name="eFormidling">
      <bpmn:incoming>SequenceFlow_4ooasd</bpmn:incoming>
      <bpmn:outgoing>SequenceFlow_5assd2s</bpmn:outgoing>
      <bpmn:extensionElements>
        <altinn:taskExtension>
          <altinn:taskType>eFormidling</altinn:taskType>
        </altinn:taskExtension>
      </bpmn:extensionElements>
</bpmn:serviceTask>
```

martinothamar:
Nothing new to add I think

> but the patterns main strength is supporting any number of implementations, that are all invoked?

Agreed, so if our use-case here is sufficiently different there is nothing stopping us from going for a different pattern.
I guess one more thing is that keeping the config in the XML lets people reuse the same provider for multiple tasks as well.

Daniel had a slightly similar dilemma in expressing when a Fiks message should be sent: https://github.com/Altinn/app-lib-dotnet/blob/b1dd0c288f24c4fd4d9a61270ac4193edf954e2f/src/Altinn.App.Clients.Fiks/FiksArkiv/IFiksArkivAutoSendDecision.cs#L17
Though this is not set in stone yet (not been reviewed by the team).

There are opportunities for configuration errors by the users in both solutions, we just have to make sure we can provide diagnostics to the users in either case (e.g. `signeeProviderId` doesn't exist, or multiple providers referring to the same task)

ivarne:
If we want the relationship to be "BPMN selects implementation" and helpful diagnostics, we should probably just add the service implementation `.GetType().FullName` directly into BPMN.

```<signeeProviderClass>Altinn.App.Services.SigneeProviderImplementation<signeeProviderClass>```

For diagnostics we then have 3 cases
* Class does not exist via Assembly Scanning
* Class exists, but in another namespace (or they forgot namespace)
* Class exists, but is not registered in `services`

(Might look cleaner without the namespace, we can make that optional)

PR #1262, opened 2025-04-24
Use central package management

ivarne:
I think [Central package managment](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#enabling-central-package-management) makes sense for Altinn apps for the following reasons

* It enforces the use of the same versions of everything (apart from transitive dependencies that might resolve differently) across our projects and test projects.
* It provides a clean way to pin versions for transitive dependencies, without adding extra direct dependencies that might not be removed in time.
* SourceGenerators/analyzers bundle some dependencies, and this makes version conflicts somewhat hard to deal with. This lets us use the same version everywhere.


martinothamar (approved):
Makes sense to me. Renovate also works with cpm?

ivarne:
> Renovate also works with cpm?

Seems like it does. I found an issue claiming that it didn't work, but it was closed after several others confirmed that it worked.

PR #1278, opened 2025-04-29
Improve tests to not use greaterThanEq to compare where a failing lookup would make the test pass

ivarne:
It was too easy to break stuff without these tests failing


olemartinorg (approved):
Easy approve! :man_dancing:

PR #1279, opened 2025-04-29
Remove hidden data before validation and make clean data available in DataProcessor

### LINKED ISSUE #695: Don't use runtime reflection to read from data models (use source generation or IL code instead)
### Description

Runtime reflection is known to be very slow, and with validation we use lots of it to be able to read properties of dynamic objects. Generated code that looks like this would be way faster

```c#
public class FieldAccessor<T>
{
    public object? GetValue(T root, string path, int[] indexes)
    {
        return path switch
        {
            "name" => root.Name,
            "children.name" => root.Children?[indexes[0]]?.Name,
            ///...
        }
    }
    
    public void SetValue(T root, object? value, string path, int[] indexes)
    {...}
}
```

### Additional Information

There are two common patterns to generate code in C#
### Source Generation
* Pro
    * Actual source code is generated, which might make debugging easer, as it is possible to see the generated code while debugging.
    * Errors can be detected and notified at compile time.
    * Most performant, as it does not add to startup time.
* Con
    * Tooling support not always the best.
    
### IL generated code
* Pro
    * Fully runtime solution, no dependencies on build
* Con
    * First run will be slower in a new server app instance.
    * Harder to debug (no source will be available), but how likely are there to be issues in an auto generated property accessor.

### Tasks

- [ ] select SourceGenerator or IL generator
- [ ] write a method that uses reflection/sg equivalent to get a list of all `json` style accessors for all properties to use in the switch/case (respecting [JsonPropertyName], and supporting recursive types)
- [ ] Add the generated code
- [ ] Use the new field accessors everywhere.

### Acceptance Criterias

Use Source generation or cached IL generated code to access properties in data models using strings.

ivarne:
Now all data is cleaned for all (backend) validation if `AppSettings.RemoveHiddenData == true` and `IInstanceDataAccessor` has a new method `GetCleanDataAccessor` so that a DataProcessor (or UserAction) can easily get a copy of the data with hidden data removed for calculations or similar.

One key factor here is that we add a source generator to be able to work with data models more efficiently

During self review I tried to split the PR into smaller commits that stands on its own

* 63018b3c313d749e2b5a55  Move Altinn.App.Analyzer from .Core to .Api
* ecf795db94 Make changes for more efficient removal of hidden data
     * Introduce IFormDataWrapper
     * Deprecate DataModelWrapper
     * New helpers in IInstanceDataAccessor
* 1f015ef1d9fb Add soure generator for IFormDataWrapper implemenations in .Analyzers
    This includes two test projects.
    * SourceGenerator.Tests: Use the generator directly and verify text output
    * SourceGenerator.Integration.Tests: Reference the generator in build process and test the generated code
    
    This has to be two projects, because it is really hard to debug a source generator when it runs as part of the build process.
* 67d2df55c3 Add Clean and Previous InstanceDataAccessors
   This enables DataProcessors to get the clean versions of data models for calculations.
* df07e13cfcb Clean data before validation if AppSettings.RemoveHiddenData = true
* 6589b33747 Minor fixes

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/1157

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Analyzers/IncrementalGenerator/FormDataWrapperGenerator.cs
```diff
@@ -0,0 +1,232 @@
...
+                        Diagnostics.FormDataWrapperGenerator.AppMetadataError,
+                        FileLocationHelper.GetLocation(text, e.StartIndex, e.EndIndex),
+                        [e.Message]
+                    )
+                ),
+            ];
+        }
+    }
+
+    private static Result<ModelPathNode> CreateNodeThree(
```
martinothamar: nit: spelling `Three` -> `Tree`

src/Altinn.App.Analyzers/IncrementalGenerator/FormDataWrapperGenerator.cs
```diff
@@ -0,0 +1,232 @@
...
+                    ),
+                ];
+            }
+
+            List<ModelClassOrDiagnostic> rootClasses = [];
+            foreach (var dataType in dataTypes.GetArrayValues())
+            {
+                if (dataType.Type != JsonType.Object)
+                {
+                    continue;
```
martinothamar, line 107: This branch could emit a diagnostic, all the values in `dataTypes` should be objects

src/Altinn.App.Analyzers/IncrementalGenerator/FormDataWrapperGenerator.cs
```diff
@@ -0,0 +1,232 @@
...
+                var appLogic = dataType.GetProperty("appLogic");
+                if (appLogic?.Type != JsonType.Object)
+                {
+                    continue;
+                }
+
+                var classRef = appLogic.GetProperty("classRef");
+                if (classRef?.Type != JsonType.String)
+                {
+                    continue;
```
martinothamar, line 119: If `classRef` is non-null and not a string, this is also invalid configuration

src/Altinn.App.Analyzers/IncrementalGenerator/FormDataWrapperGenerator.cs
```diff
@@ -0,0 +1,232 @@
...
+                text.Path.Replace('\\', '/').EndsWith("config/applicationmetadata.json")
+            )
+            .SelectMany(ParseModelClassOrDiagnostic);
+
+        var modelPathNodesProvider = rootClasses.Combine(context.CompilationProvider).Select(CreateNodeThree);
+
+        context.RegisterSourceOutput(modelPathNodesProvider, GenerateFromNode);
+    }
+
+    private static ImmutableArray<ModelClassOrDiagnostic> ParseModelClassOrDiagnostic(
```
martinothamar, line 50: I think the modelling of the resulting type could be improved here given the additional opportunities for diagnostics. Maybe something like:

```csharp
private sealed record ParseModelClassesResult(
    EquatableArray<(string ClassName, Location? Location)> Models, 
    EquatableArray<EquatableDiagnostic> Diagnostics
);
``` 

Or use a record hierarchy/union pattern if it is beneficial to return a list of results (does `SelectMany` enable concurrency?)

src/Altinn.App.Analyzers/ModelPathNode.cs
```diff
@@ -0,0 +1,74 @@
...
+/// Node used to represent the shape of a model for the purpose of generating code.
+/// Somewhat similar to a JSON schema.
+/// </summary>
+[DebuggerDisplay("{_debugDisplayString}")]
+public record ModelPathNode
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ModelPathNode"/> class.
+    /// </summary>
+    public ModelPathNode(
```
martinothamar, line 16: It wasn't obvious to me why `cSharpName` and `jsonName` could be empty in some cases. That is for the root node right? Maybe we could have static named constructors like `CreateRoot` and `CreateProperty` or something. Type hierarchy/union pattern is also an option

src/Altinn.App.Analyzers/SourceTextGenerator/AddIndexToPathGenerator.cs
```diff
@@ -0,0 +1,156 @@
...
+    {
+        if (!methods.Add(node.TypeName))
+        {
+            // Already generated this method
+            return;
+        }
+        builder.Append(
+            $$"""
+
+                private void AddIndexToPathRecursive_{{node.Name}}(
```
martinothamar, line 44: Have you considered using one of those indented writers? A wrapper around `StringBuilder` that has state for current indentation

src/Altinn.App.Analyzers/SourceTextGenerator/AddIndexToPathGenerator.cs
```diff
@@ -0,0 +1,156 @@
...
+                }
+
+            """
+        );
+        GenerateRecursiveMethod(builder, rootNode, []);
+    }
+
+    private static void GenerateRecursiveMethod(StringBuilder builder, ModelPathNode node, HashSet<string> methods)
+    {
+        if (!methods.Add(node.TypeName))
```
martinothamar, line 36: Should we use the `INamedTypeSymbol` for this kind of thing instead?

src/Altinn.App.Analyzers/SourceTextGenerator/RemoveGenerator.cs
```diff
@@ -0,0 +1,170 @@
...
+            """
+        );
+        foreach (var child in modelPathNode.Properties)
+        {
+            if (
+                child is
+                {
+                    CSharpName: "AltinnRowId",
+                    JsonName: "altinnRowId",
+                    TypeName: "global::System.Guid",
```
martinothamar: This feels a little brittle. There is a `IsAltinnRowId` which does something similar

src/Altinn.App.Analyzers/SourceTextGenerator/SourceReaderUtils.cs
```diff
@@ -0,0 +1,92 @@
...
+                {
+                    return (UnwrapNullable(collectionTypeSymbol.TypeArguments[0]), namedTypeSymbol);
+                }
+
+                break;
+            case IArrayTypeSymbol arrayTypeSymbol:
+                // Arrays have never been supported in data models.
+                // Just ignore them for now.
+                // return (UnwrapNullable(arrayTypeSymbol.ElementType), arrayTypeSymbol);
+                throw new NotSupportedException(
```
martinothamar, line 85: Should this function create diagnostics instead of throwing?

src/Altinn.App.Analyzers/SourceTextGenerator/SourceReaderUtils.cs
```diff
@@ -0,0 +1,92 @@
...
+                        .ContainingNamespace
+                        .IsGlobalNamespace
+                );
+
+                if (collectionTypeSymbol is not null)
+                {
+                    return (UnwrapNullable(collectionTypeSymbol.TypeArguments[0]), namedTypeSymbol);
+                }
+
+                break;
```
martinothamar, line 80: Is it always `ICollection<>`? Feels a little brittle. We could build a set of collection symbols in an earlier stage and do set lookups instead

src/Altinn.App.Analyzers/SourceTextGenerator/SourceTextGenerator.cs
```diff
@@ -0,0 +1,132 @@
...
+                private readonly {{rootNode.TypeName}} _dataModel;
+
+                public Type BackingDataType => typeof({{rootNode.TypeName}});
+
+                public T BackingData<T>()
+                    where T : class
+                {
+                    return _dataModel as T
+                        ?? throw new InvalidCastException(
+                            $"Attempted to cast data model of type {{rootNode.TypeName}} to {typeof(T).FullName}"
```
martinothamar: nit: I saw from a snapshot that this includes `global::` as well, could probably have another variant that doesn't prefix with that

src/Altinn.App.Analyzers/Utils/EquitableDiagnostic.cs
```diff
@@ -0,0 +1,87 @@
...
+    private static Location? GetLocationTrimmed(Location? location)
+    {
+        if (location is null)
+        {
+            return null;
+        }
+        var lineSpan = location.GetLineSpan();
+        var mappedSpan = location.GetMappedLineSpan();
+
+        return Location.Create(lineSpan.Path, location.SourceSpan, lineSpan.Span, mappedSpan.Path, mappedSpan.Span);
```
martinothamar, line 37: Interesting, what does trimmed mean here?

src/Altinn.App.Analyzers/IncrementalGenerator/FormDataWrapperGenerator.cs
ivarne: 😄 

Fixed

src/Altinn.App.Analyzers/IncrementalGenerator/FormDataWrapperGenerator.cs
ivarne: Yes, I definitely could add more diagnostics. Not sure I want to do that, because diagnostics from incremental source generators is pretty buggy. I wanted to have a diagnostic for not finding `"ClassRef"` in compilation, and added some extra just because they were easy.

src/Altinn.App.Analyzers/IncrementalGenerator/FormDataWrapperGenerator.cs
ivarne: I think that `classRef` is required to be "string" whenever `appLogic` is not null/undefined.

The question is how thorough we should be for errors that can get caught with json-schema.

src/Altinn.App.Analyzers/IncrementalGenerator/FormDataWrapperGenerator.cs
ivarne: I don't see how having an object with two lists provides "additional opportunities for diagnostics" compared to a list of the union of the elements in the two lists.

src/Altinn.App.Analyzers/ModelPathNode.cs
ivarne: > That is for the root node right?

Yes, it is only for root.

I think the proper solution would be to create two different types. `Node(List<Property>)` and `Property(cSharpName, jsonName, node)`.

(Yes, I checked and that is what PolyType does have two levels)

Do you think clarity is worth the rewrite?

src/Altinn.App.Analyzers/SourceTextGenerator/AddIndexToPathGenerator.cs
ivarne: I did consider it, and I'll probably regret my choice if I want to indent code recursively, or easily change between tabs/spaces or `\r\n`/`\n`.

I think the code reads better with the current approach, and appending fewer and appending longer strings should be (very insignificantly) more efficient.

src/Altinn.App.Analyzers/SourceTextGenerator/AddIndexToPathGenerator.cs
ivarne: I don't see why that would be better, and it breaks the incremental source generator if I cary a reference to the symbol here.

src/Altinn.App.Analyzers/SourceTextGenerator/RemoveGenerator.cs
ivarne: Yes, this obviously belongs in a helper in `ModelPathNode`.

Fixed

src/Altinn.App.Analyzers/SourceTextGenerator/SourceReaderUtils.cs
ivarne: Yes, probably, but since 8.0.0 we crash at runtime if anyone has an array in their data model, and nobody complains (probably because they generate models in studio.)

src/Altinn.App.Analyzers/SourceTextGenerator/SourceReaderUtils.cs
ivarne: The proper thing to do would probably be to use whatever `System.Text.Json` / PolyType does.

I think we should either try to do the simple thing or rely on a library.

src/Altinn.App.Analyzers/SourceTextGenerator/SourceTextGenerator.cs
ivarne: Do you have name suggestion? Currently I have
* Name (without global:: and special caracters)
* TypeName (with global::)

What would the third version of the same be named?

Edit: I chose FullName

src/Altinn.App.Analyzers/Utils/EquitableDiagnostic.cs
ivarne: Without a reference to the symbol/compilation so that the "incremental" part of the generator actually works. It does not matter when I try to reference AdditionalFiles with Location.Create in the first place, but I copied the code from PolyType. I opened a PR to suggest my change, but the whole concept of providing diagnostics from incremental source generators seems broken.  https://github.com/eiriktsarpalis/PolyType/pull/163

src/Altinn.App.Analyzers/IncrementalGenerator/FormDataWrapperGenerator.cs
martinothamar: Reading up on the whole "diagnostics from incremental generators" topic, it seems like there is no intention to have a create story there. Maybe we should go in the opposite direction then? Drop all diagnostics here and rather implement analyzers if we feel it is necessary?

src/Altinn.App.Analyzers/ModelPathNode.cs
martinothamar: I think it is worth the rewrite at some point. If we do that now as part of this PR or in the future I'll leave for you to decide  😄

src/Altinn.App.Analyzers/SourceTextGenerator/AddIndexToPathGenerator.cs
martinothamar: I think it works well for larger blocks where you see the indentation differences in the same string, but when it is broken up due to C# loops etc the readability kind of goes out the window for me

ivarne:
Now that https://github.com/Altinn/app-lib-dotnet/pull/1421 is merged, there is no more need for this

PR #1286, opened 2025-05-02
Cancellation token for signing

HauklandJ:
## Description
Token sources: Signing-, Process- and Action controllers
"Consumers": Client.Send, response.ReadAsStringAsync

## Related Issue(s)
- #{issue number}

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/Altinn.App.Core/Features/IUserAction.cs
+++ b/src/Altinn.App.Core/Features/IUserAction.cs
@@ -17,6 +17,7 @@ public interface IUserAction
     /// Method for handling the user action
     /// </summary>
     /// <param name="context">The user action context</param>
+    /// <param name="ct">Cancellation token</param>
     /// <returns>If the handling of the action was a success</returns>
-    Task<UserActionResult> HandleAction(UserActionContext context);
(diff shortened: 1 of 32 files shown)


src/Altinn.App.Core/Features/IUserAction.cs
```diff
@@ -17,6 +17,7 @@ public interface IUserAction
     /// Method for handling the user action
     /// </summary>
     /// <param name="context">The user action context</param>
+    /// <param name="ct">Cancellation token</param>
     /// <returns>If the handling of the action was a success</returns>
-    Task<UserActionResult> HandleAction(UserActionContext context);
+    Task<UserActionResult> HandleAction(UserActionContext context, CancellationToken ct);
```
bjorntore: I think it's quite likely that some apps have implemented custom IUserAction-s. Can you make this non-breaking by adding default value?

HauklandJ: I missed the ImplementableByApps attribute, good catch. I think using _default_ which is _CancellationToken.None_ is the way to go

bjorntore (approved):

PR #1291, opened 2025-05-07
Use Register party lookup API that handles system users tokens as well

martinothamar:
## Description
Some Register APIs are not ready for system users to use (one of which is the target of `IAltinnPartyClient.GetParty`): https://github.com/Altinn/altinn-authorization-tmp/issues/88

This PR makes use of a different Register API to get parties which doesn't validate based on access, it's just a straight lookup. I put it in a new internal client with doc comments describing the constraints.

We should probably also go through the rest of the use of `IAltinnPartyClient`, as there may be more issues with system users

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation


ivarne (approved):
Should we do something with the method names of registrer client add emphasis to the fact that this is not an authorised api?

 eg `GetPartyUnchecked` or similar

martinothamar:
good idea, done

PR #1297, opened 2025-05-09
The schema for GridCell accepts null, so backend must filter out nulls

ivarne:
image

https://altinncdn.no/toolkits/altinn-app-frontend/4/schemas/json/layout/layout.schema.v1.json


bjosttveit (approved):
LGTM!

PR #1309, opened 2025-05-13
Fix null pointer exception in obsolete single data element validation

ivarne:
The previous code crashed when validating a `dataElement` where `dataType.TaskId` was `null`. I spent some time debugging, so it seems worth fixing even though it did not matter in the end.

### DIFF
--- a/src/Altinn.App.Api/Controllers/ValidateController.cs
+++ b/src/Altinn.App.Api/Controllers/ValidateController.cs
@@ -129,7 +129,9 @@ public class ValidateController : ControllerBase
             return NotFound();
         }
 
-        if (instance.Process?.CurrentTask?.ElementId == null)
+        var taskId = instance.Process?.CurrentTask?.ElementId;
+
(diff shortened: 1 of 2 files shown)


src/Altinn.App.Api/Controllers/ValidateController.cs
```diff
@@ -152,10 +154,9 @@ public async Task<IActionResult> ValidateData(
             throw new ValidationException("Unknown element type.");
         }
 
-        string taskId = instance.Process.CurrentTask.ElementId;
-
         // Should this be a BadRequest instead?
-        if (!dataType.TaskId.Equals(taskId, StringComparison.OrdinalIgnoreCase))
+        // The element will likely not be validated at all if the taskId is not the same as the one in the element
```
ivarne:
```suggestion
        // The element will likely not be validated at all if the taskId is not the same as the one in the dataType
```

martinothamar (approved):

PR #1310, opened 2025-05-13
Make expression engine work with explicit dataType in subform

ivarne:
Studio seems to prefer explicit dataType in "dataModelBindings", but backend only supported explicit reference when the target has MaxCount > 1

* Add nullable dataType to DataElementIdentifier (for efficiency)
* Check if defaultDataElement from context is of the correct type before complaining about MaxCount


src/Altinn.App.Core/Models/DataElementIdentifier.cs
```diff
@@ -37,17 +42,27 @@ public DataElementIdentifier(Guid guid)
...
     }
 
+    /// <summary>
+    /// Constructor that initializes from a <see cref="DataElement"/>
+    /// </summary>
+    public DataElementIdentifier(DataElement dataElement)
+    {
+        Guid = Guid.Parse(dataElement.Id);
+        Id = dataElement.Id;
+        DataTypeId = dataElement.DataType;
```
martinothamar, line 52: Why would we name it `DataTypeId` here as opposed to `DataType`?

martinothamar (approved):

src/Altinn.App.Core/Models/DataElementIdentifier.cs
ivarne: Because I get confused whether something is the dataType object or the ID. I can't change `dataElement.DataType` to `dataElement.DataTypeId`. Do you think I should follow the naming in storage?

martinothamar: I see that we use `DataTypeId` a couple of more places (signing and process stuff), so I'm fine with it

PR #1330, opened 2025-05-28
Fix inconsistencies and improve test coverage on Expression and ExpressionValue

ivarne:
When implementing removal of hidden data for validations, this seemed useful, but is better presented in a separate PR.

### DIFF
--- a/src/Altinn.App.Core/Internal/Expressions/ExpressionValue.cs
+++ b/src/Altinn.App.Core/Internal/Expressions/ExpressionValue.cs
@@ -1,4 +1,5 @@
 using System.Diagnostics;
+using System.Text.Encodings.Web;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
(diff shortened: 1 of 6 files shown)


ivarne:
Code checkers seems to be very eager to remind me to implement `IEquatable<>`, but apparently not about writing tests for it.

src/Altinn.App.Core/Internal/Expressions/ExpressionValue.cs
```diff
@@ -268,7 +337,6 @@ internal class ExpressionTypeUnionConverter : JsonConverter<ExpressionValue>
     /// <inheritdoc />
     public override ExpressionValue Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
     {
-        reader.Read();
```
martinothamar, line 271: Was this call moved somewhere or was it actually not needed? :O

martinothamar (approved):

bjorntore (approved):

src/Altinn.App.Core/Internal/Expressions/ExpressionValue.cs
ivarne: It is just plain wrong, and made it ard to write tests for the code. You don't call `.Read()` when reading primitive types (only on object/array).

I might not actually need this converter (yet), because I always use `Expression` for serialisation, but with proper test coverage I think it makes sense to have.

PR #1337, opened 2025-06-06
Add CancellationToken to IUserAction.HandleAction in a backwards compatible way.

ivarne:
Add CancellationToken to IUserAction.HandleAction in a backwards compatible way.

Apps that update to 8.6 (signing release) and implements `IUserAction` currently needs to manually add the `CancellationToken` parameter to their implementation.

This adds a default implementation to the interface (as we have done in other instances)

Pros:
* Apps that upgrade don't need to make changes

Cons:
* Apps that want to use the cancellation token needs to implement the old method without token as a "dummy" method
* Apps that already have updated (and added the parameter) needs to remove the parameter or add the dummy method too.

### Alternative
Remove the `CancellationToken` parameter and let users that need cancellation token fetch it from `IHttpContextAccessor.HttpContext.RequestAborted`

### DIFF
--- a/src/Altinn.App.Core/Features/Action/PaymentUserAction.cs
+++ b/src/Altinn.App.Core/Features/Action/PaymentUserAction.cs
@@ -37,7 +37,7 @@ internal class PaymentUserAction : IUserAction
     public string Id => "pay";
 
     /// <inheritdoc />
-    public async Task<UserActionResult> HandleAction(UserActionContext context, CancellationToken ct)
+    public async Task<UserActionResult> HandleAction(UserActionContext context)
     {
         if (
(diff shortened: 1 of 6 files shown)


HauklandJ (approved):

src/Altinn.App.Core/Features/Action/PaymentUserAction.cs
```diff
@@ -37,7 +37,7 @@ ILogger<PaymentUserAction> logger
     public string Id => "pay";
 
     /// <inheritdoc />
-    public async Task<UserActionResult> HandleAction(UserActionContext context, CancellationToken ct)
+    public async Task<UserActionResult> HandleAction(UserActionContext context)
```
HauklandJ, line 40:
```suggestion
    public async Task<UserActionResult> HandleAction(UserActionContext context, CancellationToken? ct = null)
```

PR #1358, opened 2025-06-24
Feature/binary data copy 1305

adamhaeger:
## Description
  
  This PR enhances the instance copy functionality to include binary data (file attachments) when creating copies of archived instances,  addressing a feature gap compared to Altinn 2 where users expect both form fields and file attachments to be copied.

  ### Changes made:

  1. **Add binary data copying support**
     - Implements copying of binary data elements (attachments) alongside existing form data
     - Detects binary data types as those without `AppLogic.ClassRef`
     - Respects the existing `ExcludedDataTypes` configuration
     - Includes error handling with warning logs for individual attachment failures
     - Continues operation even if some attachments fail to copy

  ## Related Issue(s)
  - #1305

  ## Verification
  - [x] All tests run green

  ## Documentation
  - [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs)
  (if applicable)


src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -1024,6 +1024,91 @@ await _dataClient.InsertFormData(
...
+            {
+                continue;
+            }
+
+            if (binaryDataTypes.Any(dt => dt.Id.Equals(de.DataType, StringComparison.Ordinal)))
+            {
+                try
+                {
+                    // Get binary data from source instance
+                    var binaryData = await _dataClient.GetDataBytes(
```
martinothamar: Use `GetBinaryData` instead? Since that already produces a stream

src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -1024,6 +1024,91 @@ await _dataClient.InsertFormData(
                 await UpdateDataValuesOnInstance(application.DataFields, targetInstance, dt.Id, data);
             }
         }
+
+        if (application.CopyInstanceSettings?.IncludeAttachments is null or false)
```
ivarne:
```suggestion
        if (application.CopyInstanceSettings?.IncludeAttachments != true)
```

src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -1024,6 +1024,91 @@ await _dataClient.InsertFormData(
...
+                    );
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogWarning(
+                        ex,
+                        "Failed to copy binary data element {DataElementId} of type {DataType}",
+                        de.Id,
+                        de.DataType
+                    );
```
ivarne: Hvorfor trenger vi logge exceptions ulikt basert på type?

src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -1024,6 +1024,91 @@ await _dataClient.InsertFormData(
...
+                    _logger.LogWarning(
+                        ex,
+                        "Invalid argument encountered while processing binary data element {DataElementId} of type {DataType}",
+                        de.Id,
+                        de.DataType
+                    );
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogWarning(
```
martinothamar: What is the reasoning for the error handling here? Do we do this for other data elements as well or do we exit if we can't completely copy?

src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -1024,6 +1024,91 @@ await _dataClient.InsertFormData(
...
+            return;
+        }
+
+        // Copy binary data elements (files/attachments)
+        List<DataType> binaryDataTypes = application
+            .DataTypes.Where(dt => dt.AppLogic?.ClassRef == null)
+            .Where(dt =>
+                dt.TaskId != null
+                && dt.TaskId.Equals(targetInstance.Process.CurrentTask.ElementId, StringComparison.Ordinal)
+            )
```
ivarne, line 1041: Looks like you still filter based on taskId, but the PR description says you don't?

adamhaeger: Removed that from description, was from initial implementation. After discussing with Ronny, we are keeping this logic.

src/Altinn.App.Api/Controllers/InstancesController.cs
adamhaeger: Mulig vi ikke trenger det, var forslag fra den github boten for å gjøre det tydeligere hva som har feilet. Kan kjøre generisk om det er bedre.

src/Altinn.App.Api/Controllers/InstancesController.cs
adamhaeger: well, I guess the question is what do we want to happen. For the other datatypes we prefill the new instance with the data from the old instance so I guess its a bit less error prone. 

When it comes to attachments Id say its a bit more error prone as your're potentially dealing with a lot more data, my reasoning is that it might be sufficiant to not fail the entire copy operation if one attachment fails, the worst case then is that the user needs to re-upload one of the attachments.

What do you think?

martinothamar: Not sure. We're essentially saying to the user that we made a copy but then we didn't really if attachment copies fail. I think I'm leaning towards not having this as a third possible outcome/state and rather having copying either succeed or fail.. Seems like the least suprising of the alternatives. If attachment copies often fail I guess we have something to fix/improve.

But speaking of error handling, I see that copying from `PostSimplified` will have any exceptions during copying mapped to a 500 response with a handwritten error message for the user (since this is POST I think this is meant for service owner initiated copies? I.e. serverside). For `CopyInstance` the user is just being linked and if we fail with a 500 (default ASP.NET Core error handler in this case) the user will get some bad error handling here. Not sure if we should handle errors here better (redirect back to referrer or something?). Maybe we should ask Arbeidsflate as they are the ones meant to redirect users to this endpoint

adamhaeger: Yes good point, Ill  ask them :)

martinothamar (approved):
Before merging, can we rethrow (or not catch) the exception so that we keep the same failure behavior for all data types?

src/Altinn.App.Api/Controllers/InstancesController.cs
adamhaeger: Talked to Arbeidsflate and they have no concrete need for any other behaviour atm, they will get back to us if that changes.

PR #1383, opened 2025-07-03
Custom text parameters

bjosttveit:
## Description

- Implement support for a new custom text parameters field in `ValidationIssue` and `InstantiationValidationResult`.
- Obsolete `CustomTextParams`.
- Add support in `TranslationService`.
- Also added this to `GenericFormDataValidator.CreateValidationIssue`, this is bit annoying since you get positional parameters and would need to specify `null` for `CustomTextParams`. Is there a way to `[Obsolete]` positional parameters? 😅 
- I decided not to make `CustomTextParams` work by adding the index numbers as string keys in the new dictionary type, it never worked before, so I don't see a reason to make two new ways of doing it.

## Related Issue(s)
- App-Frontend changes: https://github.com/Altinn/app-frontend-react/pull/3526

## Verification
- [x] All tests run green

## Documentation


src/Altinn.App.Core/Internal/Texts/TranslationService.cs
```diff
@@ -95,18 +115,23 @@ ComponentContext context
...
         }
 
+        if (variable.DataSource == "customTextParameters" && customTextParameters != null)
+        {
+            return customTextParameters.GetValueOrDefault(variable.Key);
+        }
+
         _logger.LogWarning(
-            "Text resource variable with dataSource '{DataSource}' is not supported. Only 'dataModel.*', instanceContext and applicationSettings is supported. In text resource with id = {TextResourceId}",
+            "Text resource variable with dataSource '{DataSource}' is not supported. Only 'dataModel.*', instanceContext, applicationSettings, and customTextParameters is supported. In text resource with id = {TextResourceId}",
```
bjosttveit, line 164: Now, this depends on which method you called, should this be reflected in the log message?

ivarne: I think the fallback message is ok, but when you add the other conditions (eg. `&& state != null`) they should be in a separate branch that logs an error that the source can't be used in this context.

src/Altinn.App.Core/Features/Validation/GenericFormDataValidator.cs
```diff
@@ -58,13 +58,15 @@ public bool HasRelevantChanges(object current, object previous)
...
+    /// <param name="customTextParameters">Dictionary of parameters to replace after looking up the translation.</param>
     protected void CreateValidationIssue<T>(
         Expression<Func<TModel, T>> selector,
         string textKey,
         ValidationIssueSeverity severity = ValidationIssueSeverity.Error,
         string? description = null,
         string? code = null,
-        List<string>? customTextParams = null
+        List<string>? customTextParams = null,
+        Dictionary<string, string>? customTextParameters = null
```
ivarne, line 67: Can you search for the usage of this parameter? If only a few, I think it is OK to support the List version here. It just adds confusion.

ivarne: Just change the type as you did in `ITranslationService`. The previous code didn't work anyway, and those getting a build failure will need a notice that "Now it works".

ivarne:
Apart from the minor comments, I want to have a ready answer for when the service owners starts asking what is custom about these parameters. Is there any standard or non-custom parameters?

bjosttveit:
> what is custom about these parameters. Is there any standard or non-custom parameters?

Well, my thinking was:
1. they are custom compared to the standard parameters in app-frontend, which I hope we get support for overriding texts using those.
2. The name customTextParameters goes well with customTextKey, and suggests a connection, which there is.

Open to better naming ofc

src/Altinn.App.Core/Features/Validation/GenericFormDataValidator.cs
bjosttveit: Doesn't look like anyone is using `customTextParams` here

ivarne: Another reason to not attempt compatibility

src/Altinn.App.Core/Internal/Texts/TranslationService.cs
bjosttveit: Ait, I tried adding logging and early returns to the branches

ivarne (approved):

PR #1396, opened 2025-07-22
Renames Maskinporten extension methods

danielskovli:
## Description
As part of the ongoing attempt to unify the language of public API components, this PR renames the following methods from British spelling to American:
- UseMaskinportenAuthorization
- UseMaskinportenAltinnAuthorization

The `TokenAuthorities` enum has also been renamed to `TokenAuthority`, which is more in line with other enum namings.

The old spelling is still available as aliases, which will generate a compiler warning due to the  `Obsolete` annotation.


cammiida (approved):
🙏

cammiida (approved):
SLGTM 🙏

PR #1410, opened 2025-07-31
Chore/improve maskinporten client tests

danielskovli:
## Description
The aim of this PR is to tighten up the tests around MaskinportenClient's caching.

Initially, the `Get[...]Token()` methods returned a reference type object, which could be asserted as being the same (or not same) pointer as another reference. With the introduction of value type `JwtToken` at some point in the past, this assertion is no longer possible.

Additionally, the mocked `MaskinportenTokenResponse` did not contain a unique `jti`, which meant that the logic behind equality in `JwtToken` for these tests was faulty.

In the process of breaking out some internals of the MaskinportenClient class for testability, a fair bit of restructuring has taken place. The class is getting somewhat large, but I still feel that it makes sense to keep it as a single service in a single file. Opinions may differ here though.

## Summary of changes
- [Add `jti` to mocked tokens](https://github.com/Altinn/app-lib-dotnet/pull/1410/files#diff-6daf223707377be2440ee3c2d0ec103a8881977a92f89e501f5a9a04bfde7a46R640)
- [Add test to assert that the underlying cache reuses objects instead of serializing+deserializing](https://github.com/Altinn/app-lib-dotnet/pull/1410/files#diff-1ea89fe33a6a1f3517a1b5953d1a7111215026c4254a271cea7ebd0ed60c61feR291)
- [Replace false positives for "not same" checks with token equivalence (which are now always generated uniquely, because of `jti`)](https://github.com/Altinn/app-lib-dotnet/pull/1410/files#diff-1ea89fe33a6a1f3517a1b5953d1a7111215026c4254a271cea7ebd0ed60c61feR375)

## Verification
- [x] All tests run green

## Documentation


test/Altinn.App.Core.Tests/Features/Maskinporten/MaskinportenClientTest.cs
```diff
@@ -265,16 +274,49 @@ public async Task GetAccessToken_ThrowsExceptionWhenTokenIsExpired(string varian
...
         // Assert
-        await act1.Should().ThrowAsync<MaskinportenTokenExpiredException>();
-        await act2.Should().ThrowAsync<MaskinportenTokenExpiredException>();
+        await Assert.ThrowsAsync<MaskinportenTokenExpiredException>(act1);
+        await Assert.ThrowsAsync<MaskinportenTokenExpiredException>(act2);
+    }
+
+    [Theory]
+    [MemberData(nameof(Variants))]
+    public async Task TokenCache_ReusesObjects_DoesNotSerialize(string variant)
```
cammiida: Not sure I quite got the name here. Could this be an alternative? 🤷‍♀️ 
```suggestion
    public async Task TokenCache_Returns_SameInstanceForIdenticalRequests(string variant)
```

cammiida (approved):
LGTM 🙌

test/Altinn.App.Core.Tests/Features/Maskinporten/MaskinportenClientTest.cs
danielskovli: I guess what I meant was that the HybridCache wants really badly to serialize objects put into cache, and deserialize them on the way out again. That's why the `TokenCacheEntry` is annotated with `ImmutableObject(true)`, to prevent this.

But that said, I'm not arguing that the intial test name was in any way _good_ 😆

PR #1413, opened 2025-08-04
Stream Correspondence attachments to support large file upload

Ceredron:
## Description
Implement support for streaming attachments instead of storing in-memory to better support very large files that previously exhausted the memory of the Altinn Studio app.

## Related Issue(s)
- [#1196](https://github.com/Altinn/altinn-correspondence/issues/1196)

## Verification
- [X] Manual testing done (required)
- [X] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [X] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)


src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceAttachmentBuilder.cs
```diff
@@ -41,6 +41,14 @@ public ICorrespondenceAttachmentBuilderData WithSendersReference(string sendersR
     public ICorrespondenceAttachmentBuilder WithData(ReadOnlyMemory<byte> data)
     {
         BuilderUtils.NotNullOrEmpty(data, "Data cannot be empty");
+        var memStream = new MemoryStream(data.ToArray());
+        _data = memStream;
+        return this;
+    }
+
+    /// <inheritdoc/>
+    public ICorrespondenceAttachmentBuilder WithData(Stream data)
```
martinothamar, line 49: We should think about how to control disposal of streams based on input

martinothamar: If we want to actually save memory the caller has to ensure that the incoming stream is not just yet another `MemoryStream`. Our `DataClient` implementation doesn't let the user get a response where the whole body isn't buffered. We could consider updating the dataclient method returning a stream to not buffer the whole response with `HttpCompletionOption`. We should probably also have an XML doc that explains this

Ceredron: I noticed that DataClient has a method, Task<Stream> GetBinaryData. Could that one be used? It seems to return the HttpResponseMessage body as a stream directly.

martinothamar: Yes, but by default the .NET HTTP client buffers the body in memory. So we would have to use `HttpCompletionOption.ResponseHeadersRead` to signal that we can use the response as soon as the headers are ready. AFAIK the underlying body is a `MemoryStream` or similar unless this flag is set (when the flag is set the stream becomes a `NetworkStream` I think)

martinothamar: For some reason that method uses an extension method that explicitly sets `HttpCompletionOption.ResponseContentRead` (which is not what we want for large bodies that can just be passed along as a stream)

Ceredron: I see. Then we'd have to use ResponseHeadersRead in the GetAsync call that the DataClient makes. Is there any reason not to do this for the streaming call?

martinothamar: We have this problem where we aren't disposing responses anywhere... I wonder if we should add another overload to `DataClient` that retuns a wrapped disposable object that ensures disposal of resources (and doens't use `PlatformHttpException` since that captures the response object)

Ceredron: Stream inherits IDisposable out-of-the-box so I think DataClient.GetBinaryStream should handle disposals correctly. But perhaps we should use "using HttpResponseMessage response..." on line 182 in DataClient.cs.

martinothamar: I'm assuming disposing the response also disposes the stream, as the stream can not live without the underlying socket/connection owned by the response in our case (since by using the aforementioned flag the whole body won't just be buffered in a `MemoryStream` anymore). So logical order of operations should be like this:

```csharp
using var response = await client.GetAsync(...);
using var stream = await response.Content.ReadAsStreamAsync();
// stream is disposed first
// then response
```

> "using HttpResponseMessage response..." on line 182 in DataClient.cs.

So we can't do this because we actually want the response to "escape"/return to the caller. If we don't want to return the whole response I think we should have a disposable stream wrapper that does something like the above, e.g.

```csharp
public sealed record StreamResponse : IDisposable // IAsyncDisposable??
{
    internal StreamResponse(Stream stream, HttpResponseMessage response)
    {
        Stream = stream; // public prop
        _response = response;
    }

    public void Dispose()
    {
        Stream.Dispose();
        _response.Dispose();
    }
}
```

Ceredron: That's true. I made a wrapper that inherits from Stream now such that we still just return a stream (it just proxies calls stream calls to the stream it wraps).

src/Altinn.App.Core/Features/Correspondence/CorrespondenceClient.cs
```diff
@@ -191,6 +191,11 @@ CancellationToken cancellationToken
     )
     {
         using HttpClient client = _httpClientFactory.CreateClient();
+
+        // Configure HttpClient for large file uploads
+        client.Timeout = TimeSpan.FromMinutes(30);
```
bjorntore, line 196: Maybe a bit much with 30 minutes?

danielskovli:
Implemented by #1734 and #1772

Issue #1353, opened 2025-06-19
Support some way of annotating that some fields should not be removed from data model even though the component is hidden


ivarne:
### Description

A layout can define some components as conditionally `"hidden"` with an expression. If the user enters some information in a component before changing (a presumably earlier) answer, so that a components `"hidden"` expression changes value, it usually makes sense to remove the associated data. Because users might explore the logic of the app, the data needs to show up if he reverts the choice that triggered the condition.

Removal of data is done in 3 cases
* `appLogic.RemoveHiddenData` adds a step when finishing a task to remove all hidden fields from the data model(s)
* Validators that implement `ShouldRunAfterRemovingHiddenData => true` will not see hidden fields
* DataProcessors that uses `dataMutator.GetCleanDataAccessor().GetFormData<Model>()`

Sometimes we want to hide components in the frontend, without actually removing the data. There are multiple ways to implement this, and we need to make the right choice.

There are 4 distinct outcomes that are possible

1. Field is not hidden and not removed
2. Field is hidden and removed
3. Field is hidden but not removed
4. Field is not hidden but removed (maybe not relevant?)

As we have 1 boolean expression `"hidden"`, it makes sense to expand to 4 options by adding another boolean option.

### 1. New expression property `"remove"` on component
The most obvious solution would be to add a new expression `"remove"`, but for backwards compatibility it needs to default to the current value of `"hidden"` if undefined or null.

Is it obvious what `"remove"` would mean?

### 2. New property `"removeWhenHidden"` on component
This does not support outcome 4, and it is slightly awkward that the default will depend on `appSettings.RemoveHiddenData`, so we have a 3 way boolean
It has a few nice properties.
* Mostly not needed
* Adds a path to remove the global setting `RemoveHiddenData`, because we can set `removeWhenHidden` to false on all fields in a upgrade script.

### 3. New property `"visuallyHidden"` on component
For visual representations (frontend) it is evaluated and or'ed with `"hidden"`, but when removing data it is ignored.

### Other alternatives (better naming)
???

rvessb:
Jeg lurer på om det hadde vært en ide å opprettet en setting i selve datamodellen som sier noe on variabelen skal kunne slettes ved innsending....?
Ved å ha dette på komponeten (som du drodler rundt her), så får man fort situasjonen at man kan ha motstridende setting på forskjellige komponenter som peker mot samme felt i datamodellen.
[...]

PR #1421, opened 2025-08-06
Support working with clean data in dataProcessors and validators and removeWhenHidden

ivarne:
The main features of this PR is
* Add `ShouldRunAfterRemovingHiddenData` property to all the validator interfaces, so that validation can run with data from hidden components removed.
* Add `GetCleanAccessor` (and `GetPreviousAccessor`) to `IInstanceDataAccessor` so that data processors (and others) can see how the data looked like originally and how it looks with data from hidden components removed.
* Add `removeWhenHidden` to all components so that they can opt out of being removed on `/process/next` if they have a `hidden` expression that evaluates to true.

## Description
The PR is separated into 5 commits, that code rabbit has reviewed independently and it might be good for reviewers to do the same.
* Get up to speed on layout changes where non-repeating-groups (like accordion, tabs, grids...) might need to be part of a repeating group. https://github.com/Altinn/app-lib-dotnet/pull/1411
* Add source generator for IFormDataWrapper implementations https://github.com/Altinn/app-lib-dotnet/pull/1463
* Use the generated `IFormDataWrapper` in app code. https://github.com/Altinn/app-lib-dotnet/pull/1470
* Support an easy mechanism for validators to only work on "clean" data models. (`IValidator.ShouldRunAfterRemovingHiddenData`)  and Support an easy way to access data models where data referenced (only) by hidden components is removed in dataProcessors for calculations. (`IInstanceDataAccessor.GetCleanAccessor` and `IInstanceDataAccessor.GetPreviousDataAccessor`) https://github.com/Altinn/app-lib-dotnet/pull/1473
* Add new `RemoveWhenHidden` property to all components #1353 https://github.com/Altinn/app-lib-dotnet/pull/1485

## Verification
- [ ] All tests run green

## Documentation


src/Altinn.App.Analyzers/FormDataWrapperGenerator.cs
```diff
@@ -0,0 +1,254 @@
...
+    {
+        try
+        {
+            var textContent = text.GetText(token)?.ToString();
+            if (textContent is null)
+            {
+                return
+                [
+                    new(
+                        new EquatableDiagnostic(
```
martinothamar: If I remember correctly, I think we agreed to either
* Remove diagnostics from codegen
* Include a comment explaining that we should not expand the generator with diagnostics (pointing to relevant issues describing the issues you discovered)

src/Altinn.App.Core/Helpers/DataModel/FormDataWrapperFactory.cs
```diff
@@ -0,0 +1,55 @@
...
+                            pathAccessors.Add(
+                                KeyValuePair.Create(formDataWrapperInterface.GenericTypeArguments[0], type)
+                            );
+                        }
+                    }
+                }
+                catch (Exception)
+                {
+                    // Just ignore it if we can't load the assembly types for some reason
+                    // It is likely not the relevant assembly anyway
```
martinothamar: Completely empty catch clauses are kind of scary. What if we make a bug in the try block? Can you think of any good ways to let us disover potential bugs here (aside from the assembly thing)? E.g.

* Make this type non-static, inject logger and emit warning?
* Debug assertion so that we are made aware during local testing?

src/Altinn.App.Core/Internal/Data/IFormDataWrapper.cs
```diff
@@ -0,0 +1,290 @@
...
+
+namespace Altinn.App.Core.Features;
+
+/// <summary>
+/// Interface for a wrapper around a data model, that allows for easy access to fields and rows in the model.
+///
+/// Implementations for each data model type will be created by a source generator and retrieved by the
+/// <see cref="FormDataWrapperFactory"/>
+/// </summary>
+public interface IFormDataWrapper
```
martinothamar, line 15: Have you discussed this naming internally in the squad? Specifically `Wrapper` feels a little to generic. We've probably discussed this before, but about using the `Accessor` term for this as well?

src/Altinn.App.Core/Features/Validation/Default/LegacyIInstanceValidatorFormDataValidator.cs
```diff
@@ -42,7 +42,7 @@ public string ValidationSource
         {
             var type = _instanceValidator?.GetType() ?? GetType();
             Debug.Assert(type.FullName is not null, "FullName does not return null on class/struct types");
-            return type.FullName;
+            return type.FullName + "_FormData";
```
martinothamar, line 45: Can this break anyone? E.g. someone passing the source in `ignoredValidators` or someting

src/Altinn.App.Core/Internal/Data/CleanInstanceDataAccessor.cs
```diff
@@ -0,0 +1,156 @@
...
+    private readonly DataElementCache<IFormDataWrapper> _cleanCache = new();
+
+    private readonly Lazy<Task<List<DataReference>>> _hiddenFieldsTask;
+
+    public Instance Instance => _dataAccessor.Instance;
+
+    public IReadOnlyCollection<DataType> DataTypes => _dataAccessor.DataTypes;
+
+    public IReadOnlyDictionary<DataType, StorageAuthenticationMethod> AuthenticationMethodOverrides =>
+        throw new NotImplementedException();
```
martinothamar: Is this an accident? Since the override is implemented below I kind of expected `_dataAccessor.AuthenticationMethodOverrides`

src/Altinn.App.Core/Helpers/DataModel/DataModel.cs
```diff
@@ -9,6 +9,9 @@ namespace Altinn.App.Core.Helpers.DataModel;
 /// <summary>
 /// Get data fields from a model, using string keys (like "Bedrifter[1].Ansatte[1].Alder")
 /// </summary>
+[Obsolete(
+    "This class is deprecated and will be removed in a future version. Use the new LayoutEvaluatorState instead."
```
martinothamar: Is `LayoutEvaluatorState` correct here? Or is it supposed to be the wrapper

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
```diff
@@ -372,16 +372,14 @@ private LayoutSetComponent LoadLayout(LayoutSet layoutSet, List<DataType> dataTy
...
-                )
-                ?? throw new InvalidDataException(page + ".json is \"null\"")
+            var pagePath = Path.Join(folder, page + ".json");
+            PathHelper.EnsureLegalPath(folder, pagePath);
+            var pageBytes = File.ReadAllBytes(pagePath);
+            using var document = JsonDocument.Parse(
+                pageBytes,
+                new JsonDocumentOptions() { AllowTrailingCommas = true, CommentHandling = JsonCommentHandling.Skip }
             );
+            pages.Add(new PageComponent(document.RootElement, page, layoutSet.Id));
```
martinothamar: Is it safe to dispose the document while the component holds on to a reference to the document? Typically when something is disposable and is embedded/owned by something else, the owner has to also become disposable and take care of that. Maybe include a comment about why this is safe/OK?

src/Altinn.App.Core/Internal/Data/DataElementCache.cs
```diff
@@ -21,7 +21,7 @@ public async Task<T> GetOrCreate(DataElementIdentifier key, Func<Task<T>> valueF
                 _cache.Add(key.Guid, lazyTask);
             }
         }
-        return await lazyTask.Value;
+        return await lazyTask.Value.ConfigureAwait(false);
```
martinothamar, line 24: Did `ConfigureAwait(false)` do anything here?

src/Altinn.App.Core/Internal/Data/InstanceDataUnitOfWork.cs
```diff
@@ -63,11 +68,13 @@ public InstanceDataUnitOfWork(
...
         IInstanceClient instanceClient,
         ApplicationMetadata appMetadata,
+        ITranslationService translationService,
         ModelSerializationService modelSerializationService,
         IAppResources appResources,
         IOptions<FrontEndSettings> frontEndSettings,
         string? taskId,
-        string? language
+        string? language,
+        Telemetry? telemetry
```
martinothamar: Maybe do the `= null` here for symmetry?

test/Altinn.App.Core.Tests/LayoutExpressions/TestDataModel.cs
```diff
@@ -13,44 +12,42 @@ namespace Altinn.App.Core.Tests.LayoutExpressions;
...
-
     [Fact]
     public void TestSimpleGet()
     {
         var model = new Model { Name = new() { Value = "myValue" } };
-        var modelHelper = new DataModelWrapper(model);
-        modelHelper.GetModelData("does.not.exist").Should().BeNull();
-        modelHelper.GetModelData("name.value").Should().Be(model.Name.Value);
-        modelHelper.GetModelData("name.value", [1, 2, 3]).Should().Be(model.Name.Value);
+        var modelHelper = new ReflectionFormDataWrapper(model);
```
martinothamar, line 19: Do we have tests for `DataModelWrapper` elsewhere as well or do we just lose coverage on these now? It was just obsoleted right?

test/Altinn.App.SourceGenerator.Tests/ModuleInitializer.cs
```diff
@@ -0,0 +1,13 @@
+using System.Runtime.CompilerServices;
+
+namespace Altinn.App.SourceGenerator.Tests;
+
+internal static class ModuleInitializer
+{
+    [ModuleInitializer]
+    public static void Init()
+    {
+        // VerifierSettings.AutoVerify(includeBuildServer: false);
```
martinothamar: We have this pattern in the other initializers:

```csharp
        var isCi = !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("CI"));
        if (BuildServerDetector.Detected && BuildServerDetector.IsWsl && !isCi)
            BuildServerDetector.Detected = false; // WSL is not a build server
        VerifierSettings.AutoVerify(includeBuildServer: false);
```

martinothamar (approved):
Lots of great changes 🚀 

Didn't spend too much time on the layout and expression stuff as I have very limited understanding there unfortunatley.

In terms of public API I raised the question naming (`Wrapper`), which we talked about before, but apart from that I find it a little hard to follow the different variants of `IInstanceDataAccessor`. Reading the XML doc for `GetPreviousDataAccessor` it sounds more like the "initial" or "original" data accessor or something? There's also someting about "clean" and "hidden" (the connection between these terms) that doesn't feel quite right. Maybe this is something you can discuss in the squad (the naming of the new public APIs specifically).

It would be nice to get benchmarks on some of this. You're doing a lot of heavy lifting for perf here, love to see it 💪

src/Altinn.App.Core/Helpers/DataModel/FormDataWrapperFactory.cs
ivarne: The real fault situation here is when we have to return `ReflectionFormDataWrapper` instead of a source generated one, so that's probably the correct time to log a warning. 

I rewrote this part to let all the generated classes have a static `[ModuleInitializer]` and register themself instead of relying on assembly scanning, but in reality it did not change the possible failure of not finding a generated wrapper (apart from removing the need to catch exceptions)

src/Altinn.App.Core/Internal/Data/IFormDataWrapper.cs
ivarne: I somewhat agree, but using `Accessor` twice does not sit well with me. I'll try to invite some discussion and see if anyone suggests a better naming scheme. In the next major we can also rename InstanceData `Mutator`/`Accessor` if that helps.

src/Altinn.App.Core/Features/Validation/Default/LegacyIInstanceValidatorFormDataValidator.cs
ivarne: I would be really surprised if that could possibly be an issue.

The reason for this change is that the `[Obsolete]` `IInstanceValidator` has two methods that needs two different wrappers that runs at different times and thus should have two different validationSources.

src/Altinn.App.Core/Internal/Data/CleanInstanceDataAccessor.cs
ivarne: Definitely a rebase accident. Fixed!

src/Altinn.App.Core/Helpers/DataModel/DataModel.cs
ivarne: I removed the comment about alternative. The restructuring makes it different anyway.

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
ivarne: It is safe because the constructor does not keep a reference to the `JsonElement`. I'll rewrite the constructors to static `.Parse` methods so that it becomes more obvious that we don't keep a reference to the json element.

src/Altinn.App.Core/Internal/Data/DataElementCache.cs
ivarne: Probably not, but CodeRabbit insisted only in this file and as far as I understand it can't hurt.

I'm also not totally confident about the concept of awaiting the task multiple times, so I accepted the suggestion.

src/Altinn.App.Core/Internal/Data/InstanceDataUnitOfWork.cs
ivarne: OK.

test/Altinn.App.Core.Tests/LayoutExpressions/TestDataModel.cs
ivarne: I could restore the tests, but I'm not sure what good it would be to run tests on an obsolete helper that has no depencencies that could change, nor any reason to change itself before we remove it.

test/Altinn.App.SourceGenerator.Tests/ModuleInitializer.cs
ivarne: Yes, fixed!

ivarne:
@martinothamar Thanks!

I pushed 3 more commits to fix the issues you noted.

I did try to write some benchmarks, but micro benchmarks that only proves that reflection is much slower than generated code is not very informative (we know that to be the case). I looked at the open telemetry traces for a patch request, and the communication with storage and xml (de)serialisation seemed to be the relevant places to improve now.

I'm working on removing the diagnostics next.

HauklandJ:
With regards to clean being the opposite of hidden, clean could be "after different types of mutations" - if we only see it being used to represents "hidden data removed" we could do with being more specific.

Not quite sure I understand, so please enlighten me:

previous -instance data before any changes were made
dirty(?) - instance data with changes -  with hidden data
clean - instance data with changes - with hidden data removed 
?

src/Altinn.App.Core/Helpers/DataModel/FormDataWrapperFactory.cs
martinothamar: I was more referring to if we introduced bug in the old try block, it would be hard to find out ;) But yes I agree it would be nice to have some logging for the reflection vs src gen thing as well. Perhaps during startup, but I'm fine with leaving that for later. 

The new solution looks fine to me, a couple of drawbacks:
* Slower lookups during runtime
* Exposes a public function that isn't really meant to be called by any other codepath

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
martinothamar: Ahhhh, sorry should have checked my assumptions ;)

test/Altinn.App.Core.Tests/LayoutExpressions/TestDataModel.cs
martinothamar: Typically one does this because it's hard to predict future changes, but this is your call, resolving 😄

src/Altinn.App.Core/Implementation/AppResourcesSI.cs
ivarne: > Ahhhh, sorry should have checked my assumptions ;)

Anyways, if you read it wrong, it was bad design.

src/Altinn.App.Analyzers/FormDataWrapperGenerator.cs
ivarne: I restructured the analyser to use common code with the generator in a separate analyser step in https://github.com/Altinn/app-lib-dotnet/pull/1421/commits/9d2e3c2de71334a34908bafc2dcd320c7cc8df26

src/Altinn.App.Core/Helpers/DataModel/FormDataWrapperFactory.cs
ivarne: I don't really see why this would cause slower lookups during runtime.

I use `Dictionary<Type, (object)=>IFormDataValidator>` instead of `FrozenDictionary<Type,Type>` followed by `Activator.CreateInstance`. Do you think the frozen dictionary should be obviously faster?

The public function is not ideal, but not something I think we should worry about.

Issue #1402, opened 2025-07-24
Add documentation for Sensitive data elements


danielskovli:
Write up the appropriate docs related to new sensitive data setup and flow.

Re: https://github.com/Altinn/app-lib-dotnet/issues/1376

Issue #1376, opened 2025-07-02
Protect and/or hide sensitive data elements


danielskovli:
## Description
Certain apps require the ability to store and interact with data elements that should be hidden or otherwise restricted from the active user. The primary use case is to provide the receiving system and case owner with enough information to answer the case. Some of that information may be information not available to the active user, but available to the service owner owning the service.

The app should be able to create these data elements without the user gaining access to read the content.

## Proposed implementation
With the current app authorization infrastructure, the app itself does not have an identity. For this reason, we cannot allow the app to read back the data of sensitive data elements without an appropriate token. Should the app require to read this data after initially writing it, an org-token (via Maskinporten) will be required.

1. Create new properties for `Storage.DataType`:
   - `ActionRequiredToRead`: An optional user-action that is required to **read** (get) data elements of this type. In addition to `instance.read`, which always required
   - `ActionRequiredToWrite`: An optional user-action that is required to **write** (post/put/del) data elements of this type. In addition to `instance.write`, which is always required
2. The new properties are specified in `applicationmetadata.json` along with the existing configuration in the `dataTypes` section
3. The relevant user-actions are assigned to matching users in `policy.xml`
4. `App.DataController` __and__ `Storage.DataController` adds the following auth checks:
   - Get: Requires `ActionRequiredToRead`
   - Post/put/patch/del: Required `ActionRequiredToWrite`
5. Every time the app returns a list of data elements, it should filter out items the current user does have **read** access to. This filtration will occur in the following general areas:
   - Instance endpoint
   - ~~Process endpoint~~
   - Datacontroller/Patch endpoints
      - This filtration will incur additional calls to authorization, in order to determine if the user can read objects of certain types. However, these calls will only need to be performed if data elements of restricted types are found for a given task
6. The filtration of data elements happens in the `App` not in `Storage`, so the element IDs are more readily accessible for backend hooks (validators, etc) should they need to read this data.
   - However, should the app need to read a restricted data element outside of an appropriate user context, an org-token must be generated with the `IMaskinportenClient` to be used for this purpose
      - This will come with additional setup requirements for the app developer
7. Because of the abovemention filtration, the frontend code does not need to be adjusted to support this feature

## Notes/questions
[...]

PR #1425, opened 2025-08-07
DataClient authentication improvements

danielskovli:
## Description
The fact that DataClient only permits authentication with the current user's token has been a low-key problem for a while now. But with the introduction of [support for sensitive data elements](https://github.com/Altinn/app-lib-dotnet/issues/1376), this lack of flexibility has become more pressing.

The changes proposed in this PR are by definition breaking, but only if users have for some reason implemented their own version of `IDataClient` or are manually calling the `DataClient` constructor.  However, it has been decided internally that this usage is not something we wish to support, which means that an update like what's proposed here can still be released as a minor package version.

For all "normal" usage of `DataClient`, the changes are non-breaking and do not require any action by the consumers  unless they wish to make use of the new features.

Naming of types and services is very much up for discussion, please propose alternatives if you have concerns about the current naming strategy.

## Summary of changes
- Introduces new "union" type `AuthenticationMethod` and concrete exposing type `StorageAuthenticationMethod`
- Introduces a new service `AuthenticationTokenResolver`, which takes an `AuthenticationToken` request and maps this to an actual call to the relevant token service
- Extends `DataClient` to accept a `StorageAuthenticationMethod` parameter for all applicable methods

## Related issues
- https://github.com/Altinn/app-lib-dotnet/issues/1376
- https://github.com/Altinn/app-lib-dotnet/issues/1402


src/Altinn.App.Core/Infrastructure/Clients/Storage/AuthenticationTokenResolver.cs
```diff
@@ -0,0 +1,114 @@
...
+        IHostEnvironment hostEnvironment,
+        IOptionsMonitor<PlatformSettings> platformSettings
+    )
+    {
+        _userTokenProvider = userTokenProvider;
+        _maskinportenClient = maskinportenClient;
+        _appMetadata = appMetadata;
+        _httpClientFactory = httpClientFactory;
+        _platformSettings = platformSettings;
+        _isDev = hostEnvironment.IsDevelopment();
```
martinothamar: I think we use a different heuristic for the `LocaltestValidation` thing, we should probably do the same thing and potentially pull it out into some common internal thing

src/Altinn.App.Core/Infrastructure/Clients/Storage/AuthenticationTokenResolver.cs
```diff
@@ -0,0 +1,114 @@
...
+    private readonly IMaskinportenClient _maskinportenClient;
+    private readonly IAppMetadata _appMetadata;
+    private readonly IHttpClientFactory _httpClientFactory;
+    private readonly IOptionsMonitor<PlatformSettings> _platformSettings;
+
+    private readonly bool _isDev;
+
+    public AuthenticationTokenResolver(
+        IHttpClientFactory httpClientFactory,
+        IUserTokenProvider userTokenProvider,
```
martinothamar: Possible to use `IAuthenticationContext` instead?

src/Altinn.App.Core/Infrastructure/Clients/Storage/AuthenticationTokenResolver.cs
```diff
@@ -0,0 +1,114 @@
...
+using Altinn.App.Core.Models;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
+
+namespace Altinn.App.Core.Infrastructure.Clients.Storage;
+
+/// <summary>
+/// Resolves authentication tokens based on the specified authentication method.
+/// </summary>
+public interface IAuthenticationTokenResolver
```
martinothamar: could probably be internal?

src/Altinn.App.Core/Infrastructure/Clients/Storage/AuthenticationTokenResolver.cs
```diff
@@ -0,0 +1,114 @@
...
+        string token = await response.Content.ReadAsStringAsync(cancellationToken);
+        return JwtToken.Parse(token);
+    }
+
+    private static async Task EnsureSuccessStatusCode(HttpResponseMessage response)
+    {
+        if (response.IsSuccessStatusCode)
+            return;
+
+        throw await PlatformHttpException.CreateAsync(response);
```
martinothamar: If we use this then we can't dispose the response. So either not dispose at all like other clients or just conditionally dispose (dispose in OK branches but not when we end up here) somehow

src/Altinn.App.Core/Infrastructure/Clients/Storage/AuthenticationTokenResolver.cs
martinothamar: I see you did something there, so perhaps you meant to change this

martinothamar (commented):
I like this, simple and clear approach

src/Altinn.App.Core/Infrastructure/Clients/Storage/AuthenticationTokenResolver.cs
danielskovli: So I guess I did a half/half approach here, using the BaseUrl logic from `LocalTestValidation`, but changed by mind a few times on how to trigger the "are we running in localtest" condition. I have no prefs here, either way will work fine I think, but perhaps the `IsDevelopment` method is a bit more vulnerable because it's an env var not really under our control?

src/Altinn.App.Core/Infrastructure/Clients/Storage/AuthenticationTokenResolver.cs
danielskovli: I had it as internal first, which means that `DataClient` needs to inject `IServiceProvider`, which tempted me to use the service provider for all service resolution, which in turn made me realised I had created quite a big diff from the original implementation.

Other than that word salad, my preference is definitely `internal` here.

src/Altinn.App.Core/Infrastructure/Clients/Storage/AuthenticationTokenResolver.cs
danielskovli: I had it as internal first, which means that `DataClient` needs to inject `IServiceProvider`, which tempted me to use the service provider for all service resolution, which in turn made me realised I had created quite a big diff from the original implementation.

Other than that word salad, my preference is definitely `internal` here.

src/Altinn.App.Core/Infrastructure/Clients/Storage/AuthenticationTokenResolver.cs
danielskovli: Ooh, good idea, I forgot that this existed!

src/Altinn.App.Core/Infrastructure/Clients/Storage/AuthenticationTokenResolver.cs
danielskovli: This gets me every time

bjorntore (approved):

HauklandJ (approved):
⭐

src/Altinn.App.Core/Internal/Auth/IAuthenticationTokenResolver.cs
```diff
@@ -0,0 +1,18 @@
+using Altinn.App.Core.Models;
+using Altinn.App.Core.Models.AuthenticationMethod;
+
+namespace Altinn.App.Core.Internal.Auth;
+
+/// <summary>
+/// Resolves authentication tokens based on the specified authentication method.
+/// </summary>
+internal interface IAuthenticationTokenResolver
```
martinothamar, line 9: nit: siden dette er internal så trenger vi egentlig ikke interfacet

test/Altinn.App.Core.Tests/PublicApiTests.PublicApi_ShouldNotChange_Unintentionally.verified.txt
```diff
@@ -4186,6 +4186,18 @@ namespace Altinn.App.Core.Models
         public Altinn.Platform.Register.Models.Party UserParty { get; set; }
     }
 }
+namespace Altinn.App.Core.Models.AuthenticationMethod
+{
+    public abstract class AuthenticationMethod : System.IEquatable<Altinn.App.Core.Models.AuthenticationMethod.AuthenticationMethod> { }
```
martinothamar: trengte denne å være public? Eller skulle dette bare være det interne maskineriet som alle variantene bruker

martinothamar (approved):

test/Altinn.App.Core.Tests/PublicApiTests.PublicApi_ShouldNotChange_Unintentionally.verified.txt
martinothamar: Også ble jeg plutselig litt usikker på namespace her, vet ikke hvor godt definert `Models` er, og det er ofte litt problematisk å ha et typenavn som er identisk med en del i et namespace (`AuthenticationMethod`)

src/Altinn.App.Core/Internal/Auth/IAuthenticationTokenResolver.cs
danielskovli: har vi en preferanse her? interfacet kan gjøre det litt lettere med mocking og testing 🤷

test/Altinn.App.Core.Tests/PublicApiTests.PublicApi_ShouldNotChange_Unintentionally.verified.txt
danielskovli: Hvis vi skal beholde implicit conversion fra StorageAuthenticationMethod til AuthenticationMethod, så må den være public (fordi operator må være public).

Jeg synes det var en nice måte å bruke det på, men vi kan også så klart wrappe up StorageAuthenticationMethod.Request eller hva den proppen heter.

danielskovli: > det er ofte litt problematisk å ha et typenavn som er identisk med en del i et namespace (AuthenticationMethod)

Det brant meg i testene, og var ganske irriterende 😬 

Skal vi flytte denne? Jeg la den i egen mappe/namespace fordi den skulle ligge sammen med alle arvtakerne (hittils bare Storage)

src/Altinn.App.Core/Internal/Auth/IAuthenticationTokenResolver.cs
danielskovli: For referanse valgte vi å beholde interface

PR #1427, opened 2025-08-08
Integration test improvements

martinothamar:
## Description

- Refactor and parameterize BasicAppTests
- Claude commands for porting and implementing new integration test
- Classfixture to reuse containers for tests that don't need a clean state
- Deterministic nupkg builds so that Docker's layer cache works (deterministic .NET builds does not mean deterministic nupkgs. The sha256 checksum which I believe Docker uses is different due to timestamps and such)
- Remove CliWrap and do custom sync/threaded command running. See the code comment for more info, there has been some deadlock/hanging issue that I'm having trouble reproducing and diagnosing

Tests run atleast twice as fast on my machine now

Also, some flakyness stemming from issue described here: https://github.com/Altinn/app-localtest/pull/169

Tests can be run like this:

```sh
TEST_LOCALTEST_BRANCH="fix/storage-instancerepo-race" make ti
```

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation


bjorntore (approved):
I **think** it's good! I'm thinking we just merge and let the rest of the team try building tests this way?

PR #1451, opened 2025-08-25
Add endpoint for updating tags on a dataElement

cammiida:
## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/1070

## Verification
- [x] All tests run green

## Documentation

## Summary by CodeRabbit

- **New Features**
  - Added a PUT endpoint to replace all tags on a data element. Validates tag formats, applies changes, and returns updated tags plus optional validation issues via new request/response models.

- **Documentation**
  - OpenAPI updated with the new endpoint and schemas.
  - Fixed typo in tags description (“string values”).

- **Tests**
  - Added comprehensive integration tests for GET/POST/DELETE/PUT tag operations, validation rules, clearing tags, ignored-validators behavior, and not-found cases.


src/Altinn.App.Api/Controllers/DataTagsController.cs
```diff
@@ -179,4 +192,95 @@
...
+            validationIssues = await _validationService.ValidateInstanceAtTask(
+                dataAccessor,
+                taskId,
+                ignoredValidators,
+                true,
+                language
+            );
+        }
+        return validationIssues;
+    }
```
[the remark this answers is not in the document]
cammiida: This is a conscious choice because of how other endpoints that also do validation work. However, this is open for debate if someone feels strongly about this.

src/Altinn.App.Api/Controllers/DataTagsController.cs
```diff
@@ -179,4 +192,110 @@ [FromRoute] string tag
...
+        [FromRoute] Guid instanceGuid,
+        [FromRoute] Guid dataGuid,
+        [FromBody] SetTagsRequest? setTagsRequest,
+        [FromQuery] string? ignoredValidators = null,
+        [FromQuery] string? language = null
+    )
+    {
+        if (!ModelState.IsValid || setTagsRequest is null)
+        {
+            return BadRequest(ModelState);
```
martinothamar: What does ModelState do here? I think we should use `ProblemDetails` like it seems we've standardized on (that goes for the other responses below as well)

ivarne: Probably CodeQL suggestion because the controller is not marked with `[ApiController]` and thus the framework will call your action method even if not all parameters could be set to valid values from the request.

I think it would be wise to remove this block and add `[ApiController]` instead.

src/Altinn.App.Api/Controllers/DataTagsController.cs
```diff
@@ -179,4 +192,110 @@ [FromRoute] string tag
...
+        }
+
+        var tags = setTagsRequest.Tags;
+
+        if (tags.Any(tag => string.IsNullOrWhiteSpace(tag) || !LettersRegex().IsMatch(tag)))
+        {
+            return BadRequest("Tags may only contain letters, '-' and '_'.");
```
ivarne: As Martin said, we should return ProblemDetails for errors.
```suggestion
            return BadRequest(
                new ProblemDetails
                {
                    Title = "Invalid tag name",
                    Detail = "Tags may only contain letters, '-' and '_'.",
                    Status = StatusCodes.Status400BadRequest,
                }
            );
```
There are more, but I don't think a comment on each line is very helpful.

src/Altinn.App.Api/Controllers/DataTagsController.cs
```diff
@@ -179,4 +192,110 @@ [FromRoute] string tag
...
+            return NotFound("Unable to find data element based on the given parameters.");
+        }
+
+        // Clear existing tags
+        dataElement.Tags ??= [];
+        dataElement.Tags.Clear();
+
+        // Add new tags
+        var normalizedTags = tags.Distinct(StringComparer.Ordinal).ToList();
+        dataElement.Tags.AddRange(normalizedTags);
```
ivarne: Nit:
Would't this be the same as
```suggestion
        // Add new tags
        dataElement.Tags = tags.Distinct(StringComparer.Ordinal).ToList();
```

src/Altinn.App.Api/Models/SetTagsRequest.cs
```diff
@@ -0,0 +1,15 @@
...
+/// <summary>
+/// Represents the request body for setting a set of tags on a data element.
+/// </summary>
+public sealed class SetTagsRequest
+{
+    /// <summary>
+    /// A list of tags to set on the data element represented as string values.
+    /// </summary>
+    [JsonPropertyName("tags")]
+    public List<string> Tags { get; init; } = [];
```
ivarne: I think this can be marked required, so that openapi also knows it is required.
```suggestion
    public required List<string> Tags { get; init; }
```

src/Altinn.App.Api/Controllers/DataTagsController.cs
```diff
@@ -179,4 +192,110 @@ [FromRoute] string tag
...
+    /// <summary>
+    /// Sets a set of tags on a data element.
+    /// </summary>
+    /// <param name="org">The short name for the application owner.</param>
+    /// <param name="app">The name of the application.</param>
+    /// <param name="instanceOwnerPartyId">The party id of the owner of the instance.</param>
+    /// <param name="instanceGuid">The id of the instance.</param>
+    /// <param name="dataGuid">The id of the data element.</param>
+    /// <param name="setTagsRequest">The request body.</param>
+    /// <param name="ignoredValidators">comma separated string of validators to ignore.</param>
```
ivarne:
```suggestion
    /// <param name="ignoredValidators">comma separated string of validators to ignore. If missing we don't run validation</param>
```

src/Altinn.App.Api/Models/SetTagsResponse.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// A list of tags represented as string values.
+    /// </summary>
+    [JsonPropertyName("tags")]
+    public List<string> Tags { get; init; } = [];
+
+    /// <summary>
+    /// List of validation issues that changed as a result of updating tags.
+    /// </summary>
+    [JsonPropertyName("validationIssues")]
+    public required List<ValidationSourcePair> ValidationIssues { get; init; }
```
ivarne: Given that we only validate when setting `ignoredValidators` to something not null, the list should be optional `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]`

The Tags property should probably be `required`

ivarne (changes requested):
Litt småplukk jeg fant ved gjennomlesning. Si fra hvis du vil jeg bare skal pushe endringene jeg foreslår til branchen.

src/Altinn.App.Api/Controllers/DataTagsController.cs
cammiida: Haha true, I think so. Good catch 😅

bjorntore (approved):
Struggling to see the reason why the class is partial, but if there is a reason I can't see, fine! :P

danielskovli:
@bjorntore 
> Struggling to see the reason why the class is partial, but if there is a reason I can't see, fine! :P

Its that `GeneratedRegex` generator that requires this pattern

src/Altinn.App.Api/Controllers/DataTagsController.cs
```diff
@@ -179,4 +221,111 @@ [FromRoute] string tag
...
+            return Problem(
+                title: "Data element not found",
+                detail: "Unable to find data element based on the given parameters.",
+                statusCode: StatusCodes.Status404NotFound
+            );
+        }
+
+        // Set dataElement tags to be the new tags
+        dataElement.Tags = [.. tags.Distinct(StringComparer.Ordinal)];
+        var updatedElement = await _dataClient.Update(instance, dataElement);
```
danielskovli, line 288: Do we want to save these tags on the data element, even if the `ValidateTags` call fails below?

danielskovli (approved):
LGTM, just a small question below

src/Altinn.App.Api/Controllers/DataTagsController.cs
cammiida: Good question! I think the answer is yes. It's not as much a "this tag is a valid tag"-validation as "this is the end state we want to be in before process/next"-validation. We can't really control what they put in these validators either or which will run at this point. So there might be a totally unrelated validator that is run here and that might return validation issues, which probably shouldn't affect the tags.

cammiida: So in the case of the app that currently needs this, they just need to validate that they have uploaded certain document types. That is, one for each required tag. So if we were to not set tags if there were validation issues, you would never reach a valid state because you need to set many tags before you reach that state 😄

ivarne (approved):

PR #1465, opened 2025-09-08
Also interpret 'altinn3local.no' as running in localtest

martinothamar:
## Description

* Also interpret 'altinn3local.no' as running in localtest
* Slight refactoring to pull some of these functions into `LocaltestDiscovery` (naming?)
* Adding `Options.Create` directly during tests leaves `IOptionsMonitor<>` still injectable but referring to something else... Affected tests use `Configure` instead, and it a little brittle in other ways :sweat: we should remove all `Options.Create` manual registrations in a separate PR

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

- Bug Fixes
  - Improved detection of local test platform, making authentication and token retrieval more reliable in local environments.

- Refactor
  - Centralized environment logic into a runtime environment service used across authentication, token resolution, and startup.
  - Simplified platform base-URL resolution and related validation.

- Tests
  - Updated tests to obtain environment settings via DI and the new runtime environment service.

- Chores
  - Registered the runtime environment service during app startup for consistent environment handling.

### DIFF
--- a/src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
@@ -194,6 +194,7 @@ public static class ServiceCollectionExtensions
         services.Configure<FrontEndSettings>(configuration.GetSection(nameof(FrontEndSettings)));
         services.Configure<PdfGeneratorSettings>(configuration.GetSection(nameof(PdfGeneratorSettings)));
 
+        services.AddLocaltestDiscovery();
         if (env.IsDevelopment())
             services.AddLocaltestValidation();
 
--- a/src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
+++ b/src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
@@ -194,7 +194,7 @@ public static class ServiceCollectionExtensions
         services.Configure<FrontEndSettings>(configuration.GetSection(nameof(FrontEndSettings)));
         services.Configure<PdfGeneratorSettings>(configuration.GetSection(nameof(PdfGeneratorSettings)));
 
-        services.AddLocaltestDiscovery();
+        services.AddRuntimeEnvironment();
         if (env.IsDevelopment())
             services.AddLocaltestValidation();
 
--- a/src/Altinn.App.Core/Internal/Auth/AuthenticationTokenResolver.cs
+++ b/src/Altinn.App.Core/Internal/Auth/AuthenticationTokenResolver.cs
@@ -23,15 +23,15 @@ internal class AuthenticationTokenResolver : IAuthenticationTokenResolver
         IMaskinportenClient maskinportenClient,
         IAppMetadata appMetadata,
         IAuthenticationContext authenticationContext,
-        LocaltestDiscovery localtestDiscovery
+        RuntimeEnvironment runtimeEnvironment
     )
     {
         _maskinportenClient = maskinportenClient;
         _appMetadata = appMetadata;
         _httpClientFactory = httpClientFactory;
         _authenticationContext = authenticationContext;
-        _isDev = localtestDiscovery.IsRunning();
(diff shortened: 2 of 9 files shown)


src/Altinn.App.Core/Internal/Auth/AuthenticationTokenResolver.cs
```diff
@@ -25,16 +25,15 @@ public AuthenticationTokenResolver(
...
+        LocaltestDiscovery localtestDiscovery
     )
     {
         _maskinportenClient = maskinportenClient;
         _appMetadata = appMetadata;
         _httpClientFactory = httpClientFactory;
         _authenticationContext = authenticationContext;
-        _isDev = LocaltestValidation.IsLocaltest(generalSettings.Value);
-        _localtestBaseUrl = LocaltestValidation.GetLocaltestBaseUrl(platformSettings.Value);
+        _isDev = localtestDiscovery.IsRunning();
```
danielskovli: `isDev = localtestDiscovery.IsRunning()` doesn't feel quite right. It's hard to understand what this method is actually returning. Maybe I'll circle back to some suggestions after reading the rest of the code.

bjosttveit (approved):

src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -194,6 +194,7 @@ IWebHostEnvironment env
         services.Configure<FrontEndSettings>(configuration.GetSection(nameof(FrontEndSettings)));
         services.Configure<PdfGeneratorSettings>(configuration.GetSection(nameof(PdfGeneratorSettings)));
 
+        services.AddLocaltestDiscovery();
         if (env.IsDevelopment())
```
danielskovli, line 198: Is it still the right logic to only check for localtest if `ASPNETCORE_ENVIRONMENT=Development`?

src/Altinn.App.Core/Internal/LocaltestDiscovery.cs
```diff
@@ -0,0 +1,32 @@
...
+internal static class LocaltestDiscoveryDI
+{
+    public static IServiceCollection AddLocaltestDiscovery(this IServiceCollection services)
+    {
+        services.AddSingleton<LocaltestDiscovery>();
+        return services;
+    }
+}
+
+internal sealed class LocaltestDiscovery(
```
danielskovli: What about `LocaltestDiscovery -> RuntimeEnvironment` or something along those lines? Then `IsRunning` can become `IsLocaltest` and `GetBaseUrl` can remain as it is.

src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
danielskovli: Eg. asking because we're not relying on that variable to determine if we're on localtest, so by the same logic should we perhaps not make the same assumption here?

src/Altinn.App.Core/Internal/LocaltestDiscovery.cs
martinothamar: yeah much better name/concept, thanks :+1:

src/Altinn.App.Core/Extensions/ServiceCollectionExtensions.cs
martinothamar: It probably is sufficient to just check the hostname, but I guess it's another safety measure to make sure the localtest validation never runs in tt02/prod etc. It was suggested back when we originally added the localtest validation. So for tt02 and prod, 2 things have to go wrong for it to trigger - we have to run in "Development" mode _and_ someone has to explicitly set the hostname to one of the localtest ones

danielskovli: Makes sense to me 🙂

PR #1468, opened 2025-09-10
Signing: Simple signing integration test

martinothamar:
## Description
Simple signing integration test

Rebased on https://github.com/Altinn/app-lib-dotnet/pull/1587

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

- Tests
  - Added end-to-end integration tests for a signing workflow (user and system user), covering instance creation, process progression, data downloads (JSON/XML/PDF), and logs.
  - Enhanced test infrastructure to support JSON data handling and snapshot verification for downloaded data.
  - Introduced test app configuration, process definition, and authorization policy to exercise signing scenarios, including signature metadata storage and task actions (sign/reject).


martinothamar:
Found one issue in localtest, implementation seems to be a little out of sync: https://github.com/Altinn/app-localtest/pull/178

But there is still the issue of `Could not find party for person or organization. A valid SSN or OrgNr must be provided`
which ultimately stems from `SigneeContext.From` not working as expected. I know I tested signing with system users before and I have no idea why I thought it worked (I guess I didn't pay attention to console and still got 200 like I did initially here). You can run this test command to include the localtest fix PR:

```
TEST_LOCALTEST_BRANCH="fix/signdocument-systemuser" make ti
```

But I guess the fundamental issue in `SigneeContext` is that I don't know if we can go from `systemUserId` -> `orgNr`. The code expects the `orgNr` to be set but when signing with systemuser only the system user ID is set. I am not sure if we can go from only system user ID -> org nr. The details lookup in `Authenticated` relies on the systemuser claim in the token which details which org the systemuser represents. If there is an API that maps from system user ID -> org nr I think it would exist in altinn-authentication. Not sure

ivarne:
We must look into the test failure. I have seen lots of similar runs with intermittent failures a lot lately.
image

Not sure how the content length could be different on different runs.

PR #1472, opened 2025-09-15
Implement `AuthenticationMethod` support in `IInstanceDataAccessor`

danielskovli:
## Description
This PR implements `AuthenticationMethod` support for `IInstanceDataAccessor`, and uses this ability to automatically support _restricted data_ in the `SigningService`.

## Related Issue(s)
- closes https://github.com/Altinn/team-apps-private/issues/15
- https://github.com/Altinn/altinn-support-private/issues/4841
- https://github.com/Altinn/altinn-studio-docs/pull/2361

## Verification
- [x] All tests run green

## Documentation

## Summary by CodeRabbit

- Bug Fixes
  - Restricted data types now use service-owner access where needed, preventing authorization failures during signing and when loading signee contexts.

- Refactor
  - Signing flow and data-access now derive data-type access rules from application metadata, improving reliability with no change to user-facing behavior.

- Behavior Change
  - Validation now surfaces failures when metadata or signee information cannot be retrieved (exceptions propagate).

- Tests
  - Tests updated to cover metadata-driven data-type handling and the stricter validation behavior.


src/Altinn.App.Core/Features/IInstanceDataAccessor.cs
```diff
@@ -39,6 +39,13 @@ public interface IInstanceDataAccessor
     /// <param name="dataTypeId">DataType.Id (from applicationmetadata.json)</param>
     /// <returns>The data type (or null if it does not exist)</returns>
     DataType? GetDataType(string dataTypeId);
+
+    /// <summary>
+    /// Set the authentication method associated with a given data type.
```
ivarne:
```suggestion
    /// Set the authentication method used when reading and writing data of the given data type.
```

src/Altinn.App.Core/Internal/Data/InstanceDataUnitOfWork.cs
```diff
@@ -47,6 +47,10 @@ internal sealed class InstanceDataUnitOfWork : IInstanceDataMutator
     // We want to make sure that the data elements are updated in the instance object
     private readonly ConcurrentBag<DataElement> _savedDataElements = [];
 
+    private readonly ConcurrentDictionary<string, StorageAuthenticationMethod> _authenticationMethodOverrides = [];
+    private static readonly StorageAuthenticationMethod _defaultAuthenticationMethod =
+        StorageAuthenticationMethod.CurrentUser();
```
ivarne, line 59: Nit: Could the static readonly live in the StorageAuthenticationMethod class and not in every class that uses it?

src/Altinn.App.Core/Internal/Data/InstanceDataUnitOfWork.cs
```diff
@@ -75,6 +79,12 @@ public InstanceDataUnitOfWork(
 
     public Instance Instance { get; }
 
+    /// <inheritdoc />
+    public void OverrideAuthenticationMethod(DataType dataType, StorageAuthenticationMethod method)
+    {
+        _authenticationMethodOverrides[dataType.Id] = method;
```
ivarne: What to do if this gets set multiple times to different values? Should probably throw an exeption?

ivarne (approved):
Quick read through of IInstanceDataAccessor related code found just a few minor nits

src/Altinn.App.Core/Internal/Data/InstanceDataUnitOfWork.cs
danielskovli: We made this idempotent/safe to call multiple times on purpose. Is there an assumption from the user that overriding an override is not permitted? Open question.

src/Altinn.App.Core/Internal/Data/InstanceDataUnitOfWork.cs
danielskovli: I only placed it here because it's this particular implementer that has determined the `CurrentUser` to be default. Other than that, I don't really have much of an argument for or against.

This also doesn't need to be a staticly tracked internal, it could just as well be part of the `authenticationMethod:` param logic wherever we call `DataClient`.

Do you have a strong preference here?

src/Altinn.App.Core/Internal/Data/InstanceDataUnitOfWork.cs
ivarne: I guess it would cause problems if one part of the code uses custom and another wants to set maskinporten.

I only want the error when the existing override is incompatible with the new override

src/Altinn.App.Core/Internal/Data/InstanceDataUnitOfWork.cs
danielskovli: I'm going to close this as resolved. The internal mechanics of this can be changed later, if we want to.

src/Altinn.App.Core/Internal/Data/InstanceDataUnitOfWork.cs
danielskovli: We're leaning towards last-override-wins internally in the team here. As a compromise, I've exposed the `AuthenticationMethodOverrides` register so callers have more contextual data available if they need it: https://github.com/Altinn/app-lib-dotnet/pull/1472/commits/cd00624ffd07f8938ee5d676c3e2002b49c22d68

How does this sit with you?

ivarne: I'm still 99.9 % sure that if multiple users of the data accessor sets **different** authentication overrides on the same data type, that would merit an exception, but it feels so unlikely to ever happen that I don't think it is worth discussing, so in reality your approach is acceptable.

I have no idea what the extra property would be good for.

danielskovli: Resolving this as discussed on Slack.

PR #1423, opened 2025-08-07
Use streaming endpoints for streams in Correspondence

Ceredron:
## Description
These endpoints handle big files better.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

Issue #1300, opened 2025-05-09
Validate signature hash when leaving signature step


bjorntore:
### Description

Right now you can leave the signature step with a hash that is no longer valid.

HauklandJ:
Removing myself as assigned because I did not start this issue before the summer, and will not be able to within the next 3 weeks due to support + vacation

PR #1474, opened 2025-09-17
Validate signature hashes

bjorntore:
Add a standard validator that validates that the signature hashes are still valid when leaving a signature task.

## Related Issue(s)
- #1300
- https://github.com/Altinn/app-lib-dotnet/pull/1423
- https://github.com/Altinn/app-lib-dotnet/pull/1413

## Verification
- [x] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * Streaming support for large binary files with cancellation-aware transfers.
  * Cancellation support added to HTTP operations.
  * Automatic signature-hash validation for signing tasks to detect tampered data.
  * Richer, sanitized HTTP error snapshots for clearer failure information.

* **Bug Fixes**
  * Improved diagnostics and handling for platform HTTP failures, yielding clearer error responses.

* **Tests**
  * Comprehensive unit tests added for streaming, validation, and HTTP snapshot behaviors.

* **Localization**
  * Backend fallback text for invalid signature-hash validation added.


bjorntore:
Ser jeg glemte å ta i bruk GetStreamingAsync.... WIP

martinothamar (approved):
I see that this matches the corresponding code in Storage: https://github.com/Altinn/altinn-storage/blob/6259f1a741cb0c6fc6b625575af867bc455bcf2b/src/Storage/Services/DataService.cs#L110

Maybe make a comment/note saying how this is coupled together?

danielskovli (approved):

bjorntore:
@danielskovli @martinothamar Take another look please. Ended up bringing in some more from the correspondence PR.
1. Reverted name back to GetStreamingAsync.
2. Added the ResponseWrapperStream class to ensure the http response is disposed when the stream is disposed, like in the PR.
3. I throw platform exception if Storage returns 404 not found. Is that OK? The old GetBinaryData method returns null, even though the return type is not nullable. I can either set return type to Stream? and return null, or do what I do now. Thoughts?

bjorntore:
I have not attempted to add any http client timeout override param to GetBinaryDataStream. Would not be thread safe to set that on the client in DataClient.

We could maybe consider having a separate http client for streaming, with a much higher timeout than the default, and use that in this method?

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
```diff
@@ -204,6 +205,51 @@ public async Task<Stream> GetBinaryData(
...
+            catch (Exception)
+            {
+                response.Dispose();
+                throw;
+            }
+        }
+
+        var exception = await PlatformHttpException.CreateAsync(response);
+        response.Dispose();
+        throw exception;
```
martinothamar: I don't think it's safe to let `PlatformHttpException` have a reference to a disposed response. I think I tested once that trying to access e.g. StatusCode through the exception yields `ObjectDisposedException`

bjorntore: I looked at it and though it immediately took the status code, reason and content, and wrote it to a string, but on a second look I can see it keeps a reference to the response. Crap.

Suggestion?

martinothamar: Yeah I couldn't figure out a non-breaking way to use it without not disposing at all (which unfortunately is the current practice). So I think the only option is to not use `PlatformHttpException`

bjorntore: Can you take a look at the suggested new exception type added in the last two commits? 

Do we want to go that way? Or do we want a different platform api error strategy? The last commit adds inheritance to HttpPlatformException, so that old code most likely won't crash in cases when this is thrown instead of the old exception.

@danielskovli @martinothamar

danielskovli: I don't have very many opinions here. The only thing that strikes me is that the new `PlatformHttpResponseSnapshotException` seems like a fair bit of work. The information it contains is pretty useful stuff, but perhaps a simple log error + discard the response could be good enough?

As a consumer of exceptions I imagine I would be interested in an http error code + reason, at most. I'm probably just passing it forward anyway, perhaps with some logic to pivot on error codes.

danielskovli:
@bjorntore 
> I have not attempted to add any http client timeout override param to GetBinaryDataStream. Would not be thread safe to set that on the client in DataClient.
> 
> We could maybe consider having a separate http client for streaming, with a much higher timeout than the default, and use that in this method?

I don't really know what our options are here. I think it's safe to assume that consumers of the `DataClient` may have different timeout requirements depending on their use -- across multiple methods. But if there's no way to pass these preferences to the http client from the request methods, I'm not immediately coming up with any good proposals for a solution. 

Edit: Should we investigate the feasibility of disabling the HttpClient.Timeout (e.g. infinite) and using a linked CancellationToken to set the per-request timeout preference?

bjorntore:
> @bjorntore
> 
> > I have not attempted to add any http client timeout override param to GetBinaryDataStream. Would not be thread safe to set that on the client in DataClient.
> > We could maybe consider having a separate http client for streaming, with a much higher timeout than the default, and use that in this method?
> 
> I don't really know what our options are here. I think it's safe to assume that consumers of the `DataClient` may have different timeout requirements depending on their use -- across multiple methods. But if there's no way to pass these preferences to the http client from the request methods, I'm not immediately coming up with any good proposals for a solution.
> 
> Edit: Should we investigate the feasibility of disabling the HttpClient.Timeout (e.g. infinite) and using a linked CancellationToken to set the per-request timeout preference?

Can you check the latest commit with a separate http client with long timeout for streaming? @danielskovli @martinothamar

danielskovli (approved):

src/Altinn.App.Core/Helpers/PlatformHttpResponseSnapshotException.cs
```diff
@@ -0,0 +1,204 @@
...
+            string content = response.Content is null
+                ? string.Empty
+                : await response.Content.ReadAsStringAsync(cancellationToken);
+
+            bool truncated = content.Length > MaxCapturedContentLength;
+            if (truncated)
+            {
+                content = content[..MaxCapturedContentLength];
+            }
+
```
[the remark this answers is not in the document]
martinothamar: Agreed with Coderabbit on this one

src/Altinn.App.Core/Helpers/PlatformHttpResponseSnapshotException.cs
```diff
@@ -0,0 +1,204 @@
...
+        void Append(string prefix, IEnumerable<KeyValuePair<string, IEnumerable<string>>>? headers)
+        {
+            if (headers is null)
+                return;
+            foreach ((string key, IEnumerable<string> values) in headers)
+            {
+                sb.Append(prefix).Append(": ").Append(key).Append(": ").AppendLine(string.Join(", ", values));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
martinothamar: Sounds like a good idea

src/Altinn.App.Core/Helpers/PlatformHttpResponseSnapshotException.cs
```diff
@@ -0,0 +1,204 @@
...
+            };
+
+            // Copy normal headers
+            foreach (KeyValuePair<string, IEnumerable<string>> h in response.Headers)
+            {
+                safeResponse.Headers.TryAddWithoutValidation(h.Key, h.Value);
+            }
+
+            // Attach a diagnostic snapshot body for legacy consumers (text only, truncated)
+            string mediaType = response.Content?.Headers?.ContentType?.MediaType ?? "text/plain";
```
martinothamar: Probably shouldn't default to anything here if the point is to build a completely equal response message

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
```diff
@@ -55,6 +57,18 @@ public DataClient(HttpClient httpClient, IServiceProvider serviceProvider)
         httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
         httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/xml"));
         _client = httpClient;
+
+        // Configure streaming client
+        var httpClientFactory = serviceProvider.GetRequiredService<IHttpClientFactory>();
+        _streamingClient = httpClientFactory.CreateClient("DataClient.Streaming");
```
martinothamar: I'd say this doesn't really have anything to do with stream? Expected timeout should probably be based on payload size and not globally for "streaming requests". One option is to use `CancellationToken`. We can create a `CancellationTokenSource` is the methods where we might get large payloads, set the timeout to something pretty large and then callers can still override that with the passed in `CancellationToken`. What do you think? I think that's a little more to the point as opposed to creating an entire HttpClient for this

src/Altinn.App.Api/Controllers/ProcessController.cs
```diff
@@ -680,6 +680,19 @@ private ObjectResult ExceptionResponse(Exception exception, string message)
     {
         _logger.LogError(exception, message);
 
+        if (exception is PlatformHttpResponseSnapshotException phse)
```
martinothamar, line 683: Do we need to explicitly catch these exceptions? Just wondering since we seem to have "safe" responses on the base exception now?

src/Altinn.App.Core/Helpers/PlatformHttpResponseSnapshotException.cs
```diff
@@ -0,0 +1,204 @@
...
+/// backward compatible with <see cref="PlatformHttpException"/>.
+/// <para>
+/// This class derives from <see cref="PlatformHttpException"/> so existing
+/// catch blocks continue to work. It passes a sanitized, non-streaming
+/// <see cref="HttpResponseMessage"/> to the base class to avoid keeping any
+/// live network resources, and it exposes string-based snapshot properties
+/// for safe logging and persistence.
+/// </para>
+/// </summary>
+public sealed class PlatformHttpResponseSnapshotException : PlatformHttpException
```
martinothamar: Could this be internal as well? Like if `PlattformHttpException` is fully usable now?

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
martinothamar: Added a couple of comments. Agreed that it is a bit of work, and I'm not sure if we want to go with the exception way in the next major? I'm sort of hoping we can treat IO failures as non-exceptional in the future :pray: But given the current state of the code this suddenly seems to make `PlattformHttpException` safe while also disposing the responses which is a benefit :exploding_head: cool trick!

martinothamar: But if we are going this route, it would feel better to have some tests that don't just mock the http handlers directly (create respones manually). Either using some WireMock stuff or some integration tests that trigger this exception with

* Both <400 and >=400 status codes
* With and without response bodies (also exercising the buffering/truncation logic at 16KiB)
* Having a consumer that touches the `HttpResponseMessage` stuff to make sure nothing unexpectedly throws or doesn't work (I assume it doesn't since we don't dispose `safeResponse` here)

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
bjorntore: My probably not perfect research concluded that even if I use a cancellationTokenSource with a long timeout, the http client timeout would still cancel the request after the default amount of time. Is that wrong? That's what I'm trying to get past.

bjorntore: I can still add the cts with a timeout anyway. But get back to me on the http client timeout.

src/Altinn.App.Api/Controllers/ProcessController.cs
bjorntore: Think I just followed the existing pattern here. :P 

Looking at it now, if I remove this, would this not return a hard coded 500 status code? It doesn't inherit from ServiceException.

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
martinothamar: Good point, we could set the timeout on the client to infinite and then always control cancellation using cts perhaps. I guess thats the only way we can set defaults that are still overridable by the caller

test/Altinn.App.Api.Tests/Mocks/DataClientMock.cs
```diff
@@ -131,6 +131,40 @@
...
+        }
+
+        var fs = new FileStream(
+            path,
+            FileMode.Open,
+            FileAccess.Read,
+            FileShare.Read,
+            bufferSize: 64 * 1024,
+            options: FileOptions.Asynchronous | FileOptions.SequentialScan
+        );
```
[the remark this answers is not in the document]
bjorntore: .

test/Altinn.App.Core.Tests/Features/Validators/Default/SignatureHashValidatorTests.cs
```diff
@@ -0,0 +1,404 @@
...
+        _processReaderMock
+            .Setup(x => x.GetAltinnTaskExtension("signing-task"))
+            .Returns(new AltinnTaskExtension { SignatureConfiguration = signingConfiguration });
+
+        _appMetadataMock.Setup(x => x.GetApplicationMetadata()).ReturnsAsync(applicationMetadata);
+
+        _signingServiceMock
+            .Setup(x =>
+                x.GetSigneeContexts(
+                    It.Is<IInstanceDataAccessor>(d => d == _dataAccessorMock.Object),
```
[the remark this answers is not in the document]
bjorntore: .

src/Altinn.App.Core/Features/Validation/Default/SignatureHashValidator.cs
```diff
@@ -0,0 +1,162 @@
...
+                        new ValidationIssue
+                        {
+                            Code = ValidationIssueCodes.DataElementCodes.InvalidSignatureHash,
+                            Severity = ValidationIssueSeverity.Error,
+                            Description = ValidationIssueCodes.DataElementCodes.InvalidSignatureHash,
+                        },
+                    ];
+                }
+            }
+        }
```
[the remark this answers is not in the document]
bjorntore: .

src/Altinn.App.Core/Features/Validation/Default/SignatureHashValidator.cs
```diff
@@ -0,0 +1,178 @@
...
+                    dataElementSignature,
+                    instance,
+                    applicationMetadata
+                );
+
+                if (validationIssue != null)
+                {
+                    return [validationIssue];
+                }
+            }
```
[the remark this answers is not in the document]
bjorntore: .

src/Altinn.App.Core/Helpers/PlatformHttpResponseSnapshotException.cs
```diff
@@ -0,0 +1,183 @@
...
+
+    /// <summary>
+    /// Creates a new <see cref="PlatformHttpResponseSnapshotException"/> by snapshotting
+    /// the provided <see cref="HttpResponseMessage"/> into immutable string values,
+    /// constructing a sanitized clone for the base class, and then disposing the original response.
+    /// </summary>
+    /// <param name="response">The HTTP response to snapshot and dispose.</param>
+    /// <param name="cancellationToken">A cancellation token to cancel reading the content.</param>
+    public static async Task<PlatformHttpResponseSnapshotException> CreateAndDisposeHttpResponse(
+        HttpResponseMessage response,
```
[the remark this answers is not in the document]
bjorntore: .

src/Altinn.App.Core/Helpers/PlatformHttpResponseSnapshotException.cs
```diff
@@ -0,0 +1,183 @@
...
+        finally
+        {
+            try
+            {
+                response.Dispose();
+            }
+            catch
+            {
+                /* ignore dispose failures */
+            }
```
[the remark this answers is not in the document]
bjorntore: .

src/Altinn.App.Core/Helpers/PlatformHttpResponseSnapshotException.cs
```diff
@@ -0,0 +1,204 @@
...
```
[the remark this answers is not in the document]
bjorntore: .

src/Altinn.App.Core/Helpers/PlatformHttpResponseSnapshotException.cs
```diff
@@ -0,0 +1,284 @@
...
+        if (!string.IsNullOrWhiteSpace(charset))
+        {
+            try
+            {
+                encoding = Encoding.GetEncoding(charset);
+            }
+            catch
+            {
+                // Fallback to UTF8 if charset is invalid
+            }
```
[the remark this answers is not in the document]
bjorntore: .

test/Altinn.App.Core.Tests/PublicApiTests.PublicApi_ShouldNotChange_Unintentionally.verified.txt
```diff
@@ -230,10 +230,11 @@ namespace Altinn.App.Core.Extensions
...
+        public static System.Threading.Tasks.Task<System.Net.Http.HttpResponseMessage> GetStreamingAsync(this System.Net.Http.HttpClient httpClient, string authorizationToken, string requestUri, string? platformAccessToken = null, System.Threading.CancellationToken cancellationToken = default) { }
+        public static System.Threading.Tasks.Task<System.Net.Http.HttpResponseMessage> PostAsync(this System.Net.Http.HttpClient httpClient, string authorizationToken, string requestUri, System.Net.Http.HttpContent? content, string? platformAccessToken = null, System.Threading.CancellationToken cancellationToken = default) { }
+        public static System.Threading.Tasks.Task<System.Net.Http.HttpResponseMessage> PutAsync(this System.Net.Http.HttpClient httpClient, string authorizationToken, string requestUri, System.Net.Http.HttpContent? content, string? platformAccessToken = null, System.Threading.CancellationToken cancellationToken = default) { }
```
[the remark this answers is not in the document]
bjorntore: The change does not have binary compatibility, but it's source compatible, which should be good enough since we distribute our dll through NuGet, and comsumers will rebuild their code whenever they upgrade to this version.

src/Altinn.App.Core/Internal/Data/IDataClient.cs
```diff
@@ -126,6 +127,28 @@ Task<Stream> GetBinaryData(
...
+    /// <param name="app">Application identifier which is unique within an organisation.</param>
+    /// <param name="instanceOwnerPartyId">The instance owner id</param>
+    /// <param name="instanceGuid">The instance id</param>
+    /// <param name="dataId">the data id</param>
+    /// <param name="authenticationMethod">An optional specification of the authentication method to use for requests</param>
+    /// <param name="cancellationToken">An optional cancellation token</param>
+    /// <exception cref="Altinn.App.Core.Helpers.PlatformHttpException">Thrown when the data element is not found or other HTTP errors occur</exception>
+    Task<Stream> GetBinaryDataStream(
+        string org,
+        string app,
```
ivarne: Vi trenger ikke org og app her :)

src/Altinn.App.Core/Features/Validation/Default/SignatureHashValidator.cs
```diff
@@ -0,0 +1,185 @@
...
+                instance.Id,
+                dataElementSignature.Sha256Hash,
+                sha256Hash
+            );
+
+            return new ValidationIssue
+            {
+                Code = ValidationIssueCodes.DataElementCodes.InvalidSignatureHash,
+                Severity = ValidationIssueSeverity.Error,
+                Description = ValidationIssueCodes.DataElementCodes.InvalidSignatureHash,
```
ivarne: You should set `CustomTextKey` and `CustomTextParams` instead of `Description`

And add a fallback text in TranslationService if the user does not provide

src/Altinn.App.Core/Infrastructure/Clients/Storage/DataClient.cs
```diff
@@ -87,12 +94,15 @@ public async Task<DataElement> InsertFormData<T>(
     )
         where T : notnull
     {
+        using var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        timeoutCts.CancelAfter(_httpOperationTimeout);
+
```
ivarne: This approach of combining tokens in all the methods is kind of noisy. Could we rather inject `IHttpClientFactory` and use GetClient in all the methods, with a custom timeout argument to be used when we need it overridden?

PR #1497, opened 2025-09-28
Make ["component"] lookups work in expression validation

ivarne:
For historical reasons ["component"] lookups don't assume component ids are unique across pages, thus a component lookup wanted to know the current page and failed if no component is available. Frontend has since enforced unique component ids across all pages in a layout set, so this restriction no longer applies, but the lookup code is very simple to keep in backend.

## TODO:
* ~~Copy shared tests from frontend (or write them)~~

## Related issues:
* https://github.com/Altinn/app-frontend-react/pull/3735

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability when evaluating expressions that reference missing components or pages.
  * More robust handling of optional page identifiers to avoid abrupt failures and inconsistent error paths.
  * Safer null handling to reduce edge-case crashes across layouts.

* **Tests**
  * Added validation test covering lookups of hidden components.
  * Updated test fixtures and expectations to reflect nullable page handling and cleaned-up test data.


olemartinorg (approved):
:raised_hands:

PR #1517, opened 2025-10-10
Fix use of checked AltinnPartyClient in PrefillSI

martinothamar:
## Description
Found a trace related to the bug report: c3e48a1a59a09a417e2098b0b6a6ce9d

This is a `PostSimplified` instantiation trace. During process init we are prefilling.
The caller is a system user instantiating on behalf of someone else (meaning the system user org party ID is not the instance owner). When that happens the modified switch statement falls through to `GetParty` which doesn't work for system users.

## Related Issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/1514

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

- Refactor
  - Prefill service updated to use centralized dependency resolution for improved reliability and maintainability.
  - Streamlined party lookup path for more robust data retrieval.
- Tests
  - Converted a core test to async and aligned test setup with the new dependency approach.
- Documentation
  - Updated constructor parameter documentation to reflect the revised dependency model.
- Chores
  - Updated public API verification artifacts to match the adjusted constructor signature.

### DIFF
--- a/src/Altinn.App.Core/Implementation/PrefillSI.cs
+++ b/src/Altinn.App.Core/Implementation/PrefillSI.cs
@@ -7,6 +7,7 @@ using Altinn.App.Core.Internal.Prefill;
 using Altinn.App.Core.Internal.Registers;
 using Altinn.Platform.Profile.Models;
 using Altinn.Platform.Register.Models;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json.Linq;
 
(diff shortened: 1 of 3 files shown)


ivarne (approved):
It does look like there is some sort of authorisation on partyId on every instance where this method is called, so the assumption that we can use `GetPartyUnchecked` seems valid to me.

I'm not sure I understand why we don't just use `GetPartyUnchecked` regardless of the `currentAuth`. Is it because `user.LookupSelectedParty()` and `systemUser.LoadDetails()` can reuse cached party objects?

martinothamar:
Yes I asked myself the same question and figured the potential for a cached response was reason enough to keep it

PR #1545, opened 2025-10-29
Adding support for Likert component in `HiddenFieldsForRemoval`

olemartinorg:
## Description
I'm working on getting test-apps running against the latest app-lib-dotnet version (main, specifically), but one of our test-apps (multiple-datamodels-test) failed when removing data from hidden fields because it had a Likert component. Using claude code to implement something that seems to work, including the wonky filtering functionality this component has. The filtering only affects which row indexes are visible, so it also affects which rows specifically are hidden.

To get this working properly I got claude to add the `ShouldSkipBindingForHiddenEvaluation` function as well. I don't quite know why this was needed, but I suspect it's because the Likert component is kind-of repeating (it binds to repeating rows, `object[]`), but it's also self-contained and doesn't have any children (although app-frontend actually can't handle that, so it has its own trick of injecting a `LikertItem` component as a child for every repeating row).

It also seems a bit strange to me that this should be marking both `root.questions[0]` and `root.questions[0].Answer` as hidden, so I hope that will work fine that way.

## Related Issue(s)
- https://github.com/Altinn/altinn-studio/issues/16640

## Verification
- [x] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * Added a Likert survey component for layouts with row-based questions and selectable responses.

* **Improvements**
  * Default behavior to determine which data references are removed when components are hidden, improving data cleanup for conditional fields.

* **Tests**
  * Added unit tests covering Likert component behavior when conditionally hidden.


ivarne (approved):
Fixed and approved 💯

PR #1554, opened 2025-11-03
Feat/variables in pdf name

bjorntore:
## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation

## Release Notes

* **New Features**
  * Added tracking of which task generated PDF and signature data elements, enabling automatic deletion when tasks are revisited during process re-execution.

* **Improvements**
  * Enhanced data management for generated documents through improved task-origin traceability.

### DIFF
--- a/src/Altinn.App.Core/Features/IInstanceDataMutator.cs
+++ b/src/Altinn.App.Core/Features/IInstanceDataMutator.cs
@@ -33,6 +33,11 @@ public interface IInstanceDataMutator : IInstanceDataAccessor
     /// <summary>
     /// Add a new data element without app logic to the instance.
     /// </summary>
+    /// <param name="dataTypeId">The data type identifier from application metadata</param>
+    /// <param name="contentType">The MIME content type of the binary data</param>
+    /// <param name="filename">Optional filename for the data element</param>
+    /// <param name="bytes">The binary data to store</param>
+    /// <param name="generatedFromTaskId">Optional reference to the task that generated the data element. If set, the data element will be deleted during task initialization if the task is revisited by the process engine. Used for PDF and signatures among potentially other things.</param>
     /// <remarks>
     /// Saving to storage is not done until the instance is saved, so mutations to data might or might not be sent to storage.
     /// </remarks>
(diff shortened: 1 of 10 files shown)


bjorntore:
@ivarne Bare å commite justeringer til branchen om du så for deg noe annet.

src/Altinn.App.Core/Features/IInstanceDataMutator.cs
```diff
@@ -33,14 +33,20 @@ public interface IInstanceDataMutator : IInstanceDataAccessor
     /// <summary>
     /// Add a new data element without app logic to the instance.
     /// </summary>
+    /// <param name="dataTypeId">The data type identifier from application metadata</param>
+    /// <param name="contentType">The MIME content type of the binary data</param>
+    /// <param name="filename">Optional filename for the data element</param>
+    /// <param name="bytes">The binary data to store</param>
+    /// <param name="generatedFromTaskId">Optional reference to the task that generated the data element. If set, the data element will be deleted during task initialization if the task is revisited by the process engine. Used for PDF and signatures among potentially other things.</param>
```
ivarne, line 40: Will this still be a concept of the new Process engine?

src/Altinn.App.Core/Features/IInstanceDataMutator.cs
```diff
@@ -33,14 +33,20 @@ public interface IInstanceDataMutator : IInstanceDataAccessor
...
     /// <remarks>
     /// Saving to storage is not done until the instance is saved, so mutations to data might or might not be sent to storage.
     /// </remarks>
     BinaryDataChange AddBinaryDataElement(
         string dataTypeId,
         string contentType,
         string? filename,
-        ReadOnlyMemory<byte> bytes
+        ReadOnlyMemory<byte> bytes,
+        string? generatedFromTaskId = null
```
ivarne, line 49: We have lots of data element metadata that will need to be exposed in some way, and adding optional arguments gets somewhat messy.

I suggest we consider other possible patterns.

* Just mutate the `DataElementChange` after adding (but before saving). I see that you made the property `{ get; set; }`, but we should probably have some way to lock the change for mutations after saving.
* Add a "Metadata" value object that can be passed in where we add optional parameters without creating a huge method signature
* Same as previous but use `Action<Metadata> configure` so that you can use an inline lambda to set values
* Add some fluent interface `AddBinaryDataElement(...).SetGeneratedFromTask()`. This also allows for setting the current task as default without having to add it as an argument.

ivarne (commented):
I think we should find a more scalable approach to add metadata to data elements than to just keep adding parameters to AddBinaryData and AddFormData. Please see suggestions.

PR #1555, opened 2025-11-03
Delay throwing exceptions for invalid expressions untill thy actually gets evalutated.

ivarne:
After https://github.com/Altinn/app-lib-dotnet/pull/1421 we parse `textResourceBindings` on components in backend, and that causes issues with expressions that are not implemented. We don't (yet) evaluate the bindings if they are expressions, so delaying the exceptions restores the previous behaviour

I also renamed (and inverted) `Expression.IsFunctionExpression` to `!Expression.IsLiteralValue`, because I for a second was confused what an expression would be if it was not a function expression.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearer, more consistent error reporting for invalid or unimplemented expressions.
  * More robust parsing/handling of malformed expressions to reduce surprising failures.

* **Refactor**
  * Literal values are handled consistently across evaluation and parsing, improving stability and null-safety.
  * Validation, boolean/number handling, and evaluation order tightened for more predictable results.
  * Internal argument representation streamlined while preserving compatibility.

* **Tests**
  * Updated tests to allow more flexible error-message matching.

### DIFF
--- a/src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
+++ b/src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
@@ -133,11 +133,9 @@ public static class ExpressionEvaluator
             ExpressionFunction.gatewayAction => state.GetGatewayAction(),
             ExpressionFunction.language => state.GetLanguage(),
             ExpressionFunction.INVALID => throw new ExpressionEvaluatorTypeErrorException(
-                $"Function {expr.Args.FirstOrDefault()} not implemented in backend" + expr.ToString()
+                $"Function {expr.Args.FirstOrDefault()} not implemented in backend {expr}"
             ),
-            _ => throw new UnreachableException(
-                $"Function {(int)expr.Function} not a valid enum value" + expr.ToString()
-            ),
+            _ => throw new UnreachableException($"Function {(int)expr.Function} not a valid enum value {expr}"),
         };
         return ret;
     }
@@ -777,7 +775,7 @@ public static class ExpressionEvaluator
         var all = true;
         foreach (var arg in args)
         {
-            // Ensure all args gets converted, because they might throw an Exception
+            // the LINQ All() method would short-circuit and not evaluate all args, so we do it manually to ensure exceptions are thrown correctly
             if (!PrepareBooleanArg(arg))
             {
                 all = false;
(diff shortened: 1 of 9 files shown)


src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
```diff
@@ -770,9 +772,17 @@ private static bool PrepareBooleanArg(ExpressionValue arg)
             throw new ExpressionEvaluatorTypeErrorException("Expected 1+ argument(s), got 0");
         }
 
-        var preparedArgs = args.Select(arg => PrepareBooleanArg(arg)).ToArray();
-        // Ensure all args gets converted, because they might throw an Exception
-        return preparedArgs.All(a => a);
+        var all = true;
+        foreach (var arg in args)
+        {
+            // the LINQ All() method would short-circuit and not evaluate all args, so we do it manually to ensure exceptions are thrown correctly
```
HauklandJ, line 778: That's correct, but no need to keep this comment.

src/Altinn.App.Core/Internal/Expressions/ExpressionEvaluator.cs
```diff
@@ -804,9 +814,17 @@ ExpressionValue[] args
             throw new ExpressionEvaluatorTypeErrorException("Expected 1+ argument(s), got 0");
         }
 
-        var preparedArgs = args.Select(arg => PrepareBooleanArg(arg)).ToArray();
-        // Ensure all args gets converted, because they might throw an Exception
-        return preparedArgs.Any(a => a);
+        bool any = false;
+        foreach (var arg in args)
+        {
+            // the LINQ Any() method would short-circuit and not evaluate all args, so we do it manually to ensure exceptions are thrown correctly
```
HauklandJ, line 820: That's correct, but no need to keep this comment.

src/Altinn.App.Core/Models/Expressions/ExpressionConverter.cs
```diff
@@ -119,21 +129,24 @@ private static ExpressionFunction ExpressionFunction(string? stringFunction)
     /// <inheritdoc />
     public override void Write(Utf8JsonWriter writer, Expression value, JsonSerializerOptions options)
     {
-        if (value.IsFunctionExpression)
+        if (value.IsLiteralValue)
+        {
+            // Just serialize the literal value
```
HauklandJ, line 134: Just explains how the line below works

bjorntore (approved):

HauklandJ (approved):
Nice, I like the change from !IsFunctionExpression to IsLiteralValue

PR #1562, opened 2025-11-10
Fix gateway expression evaluation crash with invalid data element identifiers

olemartinorg:
## Description

My earlier fix in #1539 regressed when #1530 was merged. Applying a fix for this problem again so that the relevant cypress tests in `ttd/stateless-app` works again in the monorepo.

## Related Issue(s)
- #{issue number}

## Verification
- [x] **Your** code builds clean without any errors or warnings
- [x] Manual testing done (required)
- [x] Relevant automated test added (if you find this hard, leave it and we'll help out) (in the monorepo!)
- [x] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where invalid default data element identifiers could be processed in certain scenarios. The system now validates that default identifiers contain valid values before using them.

PR #1563, opened 2025-11-11
Fix ExpressionExclusiveGateway when no default data element is found

ivarne:
This also requires ComponentContext.DataElementIdentifier to be nullable

## Related Issue(s)
- Replaces #1562

## Summary by CodeRabbit

* **Refactor**
  * Enhanced data element identifier handling to support optional scenarios where identifiers may not be available, improving system flexibility and robustness.

* **Tests**
  * Updated API verification tests to reflect updated parameter and property nullability changes.


olemartinorg (approved):
Verified to be working when running e2e-tests in monorepo :muscle:

PR #1572, opened 2025-11-14
Implement IAppOptionsProvider for code lists from org library

olavsorl:
## Description
Adds support for using code lists from org library in Apps. The implementation is based on the existing IAppOptionsProvider

## Related Issue(s)
- #{issue number}

## Verification
- [x] **Your** code builds clean without any errors or warnings
- [x] Manual testing done (required)
- [x] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [x] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

## Summary by CodeRabbit

* **New Features**
  * Retrieve Altinn3 library code lists as rich options (value, multilingual label, description, help text, tags) and register via AddAltinn3CodeList.

* **Behavioral Improvements**
  * Language fallback for labels, descriptions and help text (prefers Norwegian then English).
  * Short-term caching (15 minutes) to reduce repeated API requests and improve performance.

* **Configuration**
  * New setting to configure the Altinn3 library API endpoint.

* **Tests**
  * Added tests for fetching, language fallback, caching, request URIs and error handling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1588, opened 2025-12-01
Add "Unknown" `Authenticated` type, to support custom APIs with custom auth 

martinothamar:
## Description

Some apps have completely custom APIs with custom auth, in these cases we may fail to identify/parse the token based on normal app usage, in which case I think it would be nice to have this "Unknown" type. This could be viewed as a breaking change, but I don't believe we have to alter/handle any of our switch statements any differently here as our own APIs are either invokved as `None` or with some valid/authorized party (in which case we know we have an expected `Authenticated` case). Any app that has custom APIs and custom auth will simply be able to use this abstraction without getting an exception

See the added tests
- The first token represents a completely custom/unknown issuer
- The second token is a Maskinporten token but not including a scope related to Altinn Studio (we match on relevant scopes)

Was a bit unsure of the classification here. It is technically a feature that solves what I would characterize as a bug? Re: related issue. Therefore not sure if we should backport this or not

## Related Issue(s)
- https://github.com/Altinn/altinn-studio/issues/17080

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * Added support for handling authentication requests with unknown authentication types.
  * Added capability to detect Altinn instance scopes in token validation.

* **Tests**
  * Added comprehensive test coverage for unknown authentication type scenarios.
  * Added test cases for scope detection validation.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/Altinn.App.Core/Features/Auth/Authenticated.cs
+++ b/src/Altinn.App.Core/Features/Auth/Authenticated.cs
@@ -89,6 +89,16 @@ public abstract class Authenticated
     }
 
     /// <summary>
+    /// Type to indicate that the current request is authenticated, but we don't know the type of the authentication.
+    /// It may be custom for the app.
+    /// </summary>
+    public sealed class Unknown : Authenticated
+    {
+        internal Unknown(ref ParseContext context)
+            : base(ref context) { }
+    }
+
+    /// <summary>
     /// The logged in client is a user (e.g. Altinn portal/ID-porten)
     /// </summary>
     public sealed class User : Authenticated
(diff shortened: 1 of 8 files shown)


test/Altinn.App.Tests.Common/Utils/OtelVisualizers.cs
```diff
@@ -7,12 +7,22 @@ namespace Altinn.App.Tests.Common.Utils;
...
-        List<LogRecord> logs,
+        List<Activity> activitiesInput,
+        List<LogRecord> logsInput,
         bool includeDuration = true,
         int initialIndent = 0
     )
     {
+        // This is a little racy potentially,
+        // got some enumeration errors that hopefully is avoided by taking a copy of the input first
+        // The inputs can only grow, so worst case we miss a few elements if the race hits
```
ivarne: I think the issue here might be caused by not disposing the service provider or flushing logs manually before printing the logs.

https://github.com/open-telemetry/opentelemetry-dotnet/issues/5102

I should probably just follow the instructions and dispose the _serviceProvider before visualising telemetry, instead of trying to ForceFlush.


The issue was that the API gave me no LoggerProvider to call `ForceFlush` on.
`var loggerProvider = app.Services.GetRequiredService<LoggerProvider>();`

ivarne (approved):
Does the "Closed class hierarchy" stil make sense when you add an `Unknown` option and how is Unknown different from None?

Regarding backporting I think we should look more at need versus risk, than the classification of bug versus feature (especially when the distinction is unclear).

test/Altinn.App.Tests.Common/Utils/OtelVisualizers.cs
ivarne: I got the order wrong 😭 
https://github.com/Altinn/app-lib-dotnet/blob/1192ac7c3d6ea7714f3168b64684cca901007f88/test/Altinn.App.Tests.Common/Fixtures/MockedServiceCollection.cs#L403-L425

ivarne: I think this should be fixed in https://github.com/Altinn/app-lib-dotnet/pull/1572

martinothamar:
Thinking was that
* None = not authenticated
* Unknown = there is some authentication, we just don't recognize it

I don't think we should describe it as a closed hierarchy, moreso that it is not currently open for extension by consumers. Altinn Authentication may change over time so we might end up getting more cases in there. Not sure how well documented that is though

PR #1601, opened 2025-12-17
Deny writes to ref-data-as-pdf through API. Considered to be the apps…

bjorntore:
Deny writes to ref-data-as-pdf through API. Considered to be the app´s data type.

Not using app:owned, since that would require us to change config in all apps. With this solution it´s enough that they update the NuGet.

## Related Issue(s)
- [#17269](https://github.com/Altinn/altinn-studio/issues/17269)

## Verification
- [x] All tests run green

## Summary by CodeRabbit

* **Bug Fixes**
  * PDF data elements are now protected from modification; create/update/delete attempts are blocked and return a 403 Forbidden response.
* **Tests**
  * Added tests verifying PDF elements cannot be created, updated, or deleted and that appropriate 403 responses and messages are returned.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/Altinn.App.Core/Internal/Data/DataElementAccessChecker.cs
+++ b/src/Altinn.App.Core/Internal/Data/DataElementAccessChecker.cs
@@ -97,7 +97,7 @@ internal class DataElementAccessChecker : IDataElementAccessChecker
     {
         auth ??= _authenticationContext.Current;
 
-        if (dataType.Id == PdfService.PdfElementType)
+        if (dataType.Id.Equals(PdfService.PdfElementType, StringComparison.OrdinalIgnoreCase))
         {
             return new ProblemDetails
             {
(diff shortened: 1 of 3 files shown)


src/Altinn.App.Core/Internal/Data/DataElementAccessChecker.cs
```diff
@@ -122,6 +133,7 @@ public async Task<bool> CanDelete(
             {
                 Title = "Forbidden",
                 Detail = "User is not a valid contributor to the data type",
+                Status = StatusCodes.Status403Forbidden,
```
bjorntore, line 136: Was returning 500 by mistake.

tjololo:
This looks like a good fix to me. that dataType.id is as far as I remember a special identifier where we store the pdf we generate, stopping others from changing it makes sense to me!

ivarne (approved):

PR #1661, opened 2026-02-13
chore: runtime configuration improvements

martinothamar:
## Description

- Expose a setting that we can flip through our syncroot/runtime when we have an OTel collector in place
- Make `/mnt/app-secrets/` JSON files automatically added, so we can extend configuration over time without relying on app-lib changes for every little thing

## Related Issue(s)
- N/A

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * Option to route telemetry via an OpenTelemetry collector or Azure Monitor.
  * Load runtime JSON configuration from a secrets directory with ordered override handling.
  * New configuration settings to control telemetry routing and runtime secrets location.

* **Improvements**
  * Health-check requests are excluded from telemetry to reduce noise.

* **Tests**
  * Added unit tests for runtime configuration file loading, ordering, and duplicate avoidance.

### DIFF
--- a/src/Altinn.App.Api/Extensions/WebHostBuilderExtensions.cs
+++ b/src/Altinn.App.Api/Extensions/WebHostBuilderExtensions.cs
@@ -1,5 +1,8 @@
+using Altinn.App.Core.Configuration;
 using Altinn.App.Core.Extensions;
 using Altinn.App.Core.Features.Maskinporten.Extensions;
+using Microsoft.Extensions.Configuration.Json;
+using Microsoft.Extensions.FileProviders;
 
 namespace Altinn.App.Api.Extensions;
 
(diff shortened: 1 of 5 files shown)


mirkoSekulic (approved):

src/Altinn.App.Api/Extensions/WebHostBuilderExtensions.cs
```diff
@@ -40,8 +42,106 @@ public static void ConfigureAppWebHost(this IWebHostBuilder builder, string[] ar
...
             }
         );
     }
+
+    private static void AddRuntimeConfigFiles(
+        this IConfigurationBuilder configBuilder,
+        IHostEnvironment hostEnvironment
+    )
+    {
+        const string secretsDirectory = "/mnt/app-secrets";
```
mirkoSekulic: Can we make this path configurable? I feel like this code should be unaware of our infrastructure(mount paths etc). What do you think?

martinothamar: Very good point 🫡

martinothamar: fixed?

martinothamar: maybe a little strange to have a public member for this, could be just some env var as well

danielskovli (approved):

PR #1689, opened 2026-03-09
Add fields parameter to party query endpoint URL

ivarne:
Altinn register seems to require that we request `user.id` when using `urn:altinn:person:legacy-selfidentified:` urns for lookup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Party lookup queries now request only the specific fields needed (party id and associated user id), reducing data transferred.
  * Resulting performance improvements make searching for and loading party-related information noticeably snappier with lower network usage.

### DIFF
--- a/src/Altinn.App.Core/Infrastructure/Clients/Register/AltinnPartyClient.cs
+++ b/src/Altinn.App.Core/Infrastructure/Clients/Register/AltinnPartyClient.cs
@@ -135,7 +135,7 @@ public class AltinnPartyClient : IAltinnPartyClient
     public async Task<int?> GetPartyIdByUrn(string urn)
     {
         using var activity = _telemetry?.StartLookupPartyActivity();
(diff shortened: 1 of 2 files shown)


src/Altinn.App.Core/Infrastructure/Clients/Register/AltinnPartyClient.cs
```diff
@@ -135,7 +135,7 @@ await response.Content.ReadAsStringAsync()
     public async Task<int?> GetPartyIdByUrn(string urn)
     {
         using var activity = _telemetry?.StartLookupPartyActivity();
-        string endpointUrl = "access-management/parties/query";
+        string endpointUrl = "access-management/parties/query?fields=id,user.id";
```
ivarne:
```suggestion
        string endpointUrl = "apps/parties/query?fields=id,user.id";
```

HauklandJ (approved):
Hope they fix this on their end

PR #1694, opened 2026-03-12
fix: check prop name without case-sensitivity.

Konrad-Simso:
## Description
Comapre strings without case-sensitivity. 
We've seen some examples where the prop name starts with a lowercase `a`, while the if check only accepts an uppercase `A`.

## Verification
- [x] All tests run green

## Documentation


Konrad-Simso:
Closing due to not fixing root the problem.

PR #1696, opened 2026-03-16
Fix nullability issues in source generator

ivarne:
Vi fikk noen nullability issues for apper som ikke genererer modeller med nullable reference types enabled.

For å fikse det må vi håndtere value types annerledes og legge til en `!` på copy generatoren.

## Summary by CodeRabbit

* **New Features**
  * Recognizes C# value types during type classification to improve serialization and assignment behavior.

* **Bug Fixes**
  * Refined nullability handling in generated code to suppress nullable warnings and align annotations.
  * Broadened direct-assignment logic to handle value-type elements and non-nullable lists more robustly.

* **Tests**
  * Added tests and model properties covering non-nullable lists and value-type scenarios.

### DIFF
--- a/src/Altinn.App.Analyzers/ModelPathNode.cs
+++ b/src/Altinn.App.Analyzers/ModelPathNode.cs
@@ -110,6 +110,8 @@ public record ModelPathNode
             "global::System.DateTime" => true,
             "global::System.DateTimeOffset" => true,
             "global::System.TimeSpan" => true,
+            "global::System.DateOnly" => true,
+            "global::System.TimeOnly" => true,
             "global::System.Int32" => true,
             "global::System.Int64" => true,
             "global::System.UInt32" => true,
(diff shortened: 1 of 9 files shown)


src/Altinn.App.Analyzers/ModelPathNode.cs
```diff
@@ -101,4 +101,42 @@ public bool IsAltinnRowId()
...
+            "global::System.Double" => true,
+            "global::System.Decimal" => true,
+            "global::System.Boolean" => true,
+            "global::System.Char" => true,
+            "global::System.Byte" => true,
+            "global::System.SByte" => true,
+            "global::System.Int16" => true,
+            "global::System.UInt16" => true,
+            "int" => true,
+            "long" => true,
```
HauklandJ, line 129: Missing global::System.DateOnly and global::System.TimeOnly

HauklandJ (approved):

src/Altinn.App.Analyzers/ModelPathNode.cs
ivarne: Fixed

PR #1698, opened 2026-03-16
Make component expressions subform aware

ivarne:
Previously you could not use "component" expressions in subforms, because the context lookup did error because multiple contexts would match.

## Breaking
Previously backend supported the same component ID on different pages (using the page id to break ties). Frontend errors out on duplicate ids (even across pages), and has since the early days of v4, so this is safe to remove now.

## Summary by CodeRabbit

* **Refactor**
  * Improved component context resolution for complex/subform scenarios for more accurate component lookups.
  * Marked the legacy lookup path as deprecated in favor of the new context-aware approach.

* **Tests**
  * Updated test suite to align with the new lookup behavior and adjusted a subform field validation scenario.


test/Altinn.App.Core.Tests/LayoutExpressions/FullTests/SubForm/SubFormTests.cs
```diff
@@ -178,7 +178,7 @@ private record SubFormModel(
                       "dataModelBindings": {
                         "simpleBinding": "Address"
                       },
-                      "required": true
+                      "required": ["notEquals", ["component", "Name"], "Always required, bu test component lookup"]
```
HauklandJ:
```suggestion
                      "required": ["notEquals", ["component", "Name"], "Always required, by test component lookup"]
```

HauklandJ (approved):

PR #1714, opened 2026-03-26
Delete instance on instantiation failiure

ivarne:
When instantiation fails, we previously left a half initialised instance that likely misses the data element with appLogic and fails to open in our frontend.

This attempts to delete the instance if an exception happens in the init code, but we still have the issue of application crashes that might cause deletion to fail or the full app container to crash.

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically attempts to delete partially created instances when creation, prefill/storage, or copy operations fail, reducing orphaned data.
  * Cleanup attempts are best-effort: deletion failures are caught and logged while original errors are returned.
  * Only attempts cleanup when an instance identifier is available, avoiding secondary errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Altinn/app-lib-dotnet/pull/1714?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/Altinn.App.Api/Controllers/InstancesController.cs
```diff
@@ -723,6 +730,18 @@ await _processEngine.HandleEventsAndUpdateStorage(
         }
         catch (Exception exception)
         {
+            if (instance?.Id is not null)
+            {
+                // Try to delete instance if creation fails to avoid leaving orphaned instances in storage.
+                // If deletion also fails, there is not much we can do about it, but we should at least log the error.
+                await _instanceClient.DeleteInstance(
```
martinothamar: should we wrap these calls in a try catch as well and log warnings or something? so the user isnt affected if we fail to delete. Or do we want the request to fail in that case

martinothamar (approved):

src/Altinn.App.Api/Controllers/InstancesController.cs
ivarne: The request should obviously fail (500 internal server error) if there is an exception, and still fail if the cleanup after the exception fails.

The issue here is that we actually don't log the original exception, but I think I'll just log the error.

PR #1750, opened 2026-05-13
feat: Move task data cleanup responsibility to Storage

### LINKED ISSUE #1735: Legacy PDF generation leaves stale PDF when a data task loops back to itself before ending
### Description of the bug

  Legacy PDF generation can leave a stale PDF when a data task loops back to itself and later exits directly to an end event.

  The issue happens with enablePdfCreation: true on a data type connected to a task that can repeat through a gateway:

  task4 -> gateway -> task4

  Each time the task ends, PdfServiceTaskLegacy generates a new PDF. Cleanup is normally triggered when a task starts through ProcessTaskInitializer, which deletes data elements whose generatedFromTask reference matches the starting task.

  However, when the task exits directly to an end event:

  task4 -> gateway -> EndEvent

  there is no new start event for task4, so the last stale loop PDF is not cleaned before the final PDF is generated. The result is two PDFs for the same task: one stale PDF from the previous loop iteration and one final PDF from the end transition.

  Only one XML/form data element remains, so the issue seems isolated to generated PDF data elements.

### Steps To Reproduce

  1. Create an app with a data task that can loop back to itself through a gateway.
  2. Enable legacy PDF generation for the task’s data type:
```json
  {
    "id": "model4",
    "taskId": "task4",
    "appLogic": {
      "autoCreate": true,
      "classRef": "Altinn.App.Models.model4.model4"
    },
    "enablePdfCreation": true
  }
```
  3. Configure the BPMN so the task can repeat and later exit directly to an end event:
```xml
  <bpmn:task id="task4">
    <bpmn:extensionElements>
      <altinn:taskExtension>
        <altinn:taskType>data</altinn:taskType>
      </altinn:taskExtension>
    </bpmn:extensionElements>
    <bpmn:incoming>Flow_ToTask4</bpmn:incoming>
    <bpmn:incoming>Flow_LoopBackToTask4</bpmn:incoming>
    <bpmn:outgoing>Flow_ToGateway</bpmn:outgoing>
  </bpmn:task>

  <bpmn:exclusiveGateway id="Gateway_Repeat">
    <bpmn:incoming>Flow_ToGateway</bpmn:incoming>
    <bpmn:outgoing>Flow_LoopBackToTask4</bpmn:outgoing>
    <bpmn:outgoing>Flow_ToEnd</bpmn:outgoing>
  </bpmn:exclusiveGateway>

  <bpmn:sequenceFlow id="Flow_LoopBackToTask4" sourceRef="Gateway_Repeat" targetRef="task4">
    <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">
      ["lessThan", ["dataModel", "repeat"], 1000]
    </bpmn:conditionExpression>
  </bpmn:sequenceFlow>

  <bpmn:sequenceFlow id="Flow_ToEnd" sourceRef="Gateway_Repeat" targetRef="EndEvent_1">
    <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">
      ["greaterThanEq", ["dataModel", "repeat"], 1000]
    </bpmn:conditionExpression>
  </bpmn:sequenceFlow>
```
  4. Complete task4 once with data that causes the gateway to loop back to task4.
  5. Complete task4 again with data that causes the gateway to go to EndEvent_1.
  6. Inspect instance data elements.

  Expected result:
  There should be only one generated PDF for task4, representing the final state.

  Actual result:
  There are two generated PDFs for task4: one stale PDF from the loop iteration and one final PDF from the end transition.

### Additional Information

_No response_

danielskovli:
## Summary

- Deletes `IProcessTaskCleaner` / `ProcessTaskCleaner` and the two call sites in `ProcessTaskInitializer` and `SubformPdfServiceTask`.
- Storage's `PutInstanceAndEvents` (PUT `instances/{partyId}/{guid}/process/instanceandevents`) will perform the `generatedFromTask` cleanup transactionally as part of process transitions. App-lib no longer attempts this from its side, eliminating the user-auth failure mode where the caller is authorized to advance the process but not to write/delete each affected data element.
- Pure subtraction — no new endpoints, no new interfaces. The Storage-side change is a separate PR (companion to this one).

## Why

Tracks three issues that turned out to be different surface effects of the same problem (cleanup running in app-lib against the in-memory `Instance` rather than the persisted state):
- Altinn/altinn-studio#18504 — symptom: cleanup fails when user lacks write permission during `process/next`.
- Altinn/altinn-studio#18465 — proposal: handle cleanup server-side as part of the same Storage transaction.
- #1735 — symptom: legacy PDF generation leaves a stale PDF when a data task loops back to itself before exiting to an EndEvent.

The originally considered app-lib approach (a new internal Storage endpoint called from a new `IDataClient` method using a service-owner token) was scoped out in favour of letting Storage own the cleanup transactionally inside `PutInstanceAndEvents`. Service-task entries also produce to/from task events, so Storage's logic covers the `SubformPdfServiceTask` "clean up previous failed attempts" case too — no separate path needed. The same move also resolves #1735 as a side effect: Storage runs cleanup against the persisted instance loaded from DB, which does see the legacy PDF that `PdfServiceTaskLegacy` wrote via `InsertBinaryData`.

## Companion PRs

- https://github.com/Altinn/altinn-storage/pull/977
- https://github.com/Altinn/app-localtest/pull/209

App-lib's implementation is blocked by Storage. See deployment order below.

### Deployment order
1. https://github.com/Altinn/altinn-storage/pull/977
2. https://github.com/Altinn/app-localtest/pull/209
3. https://github.com/Altinn/app-lib-dotnet/pull/1750

## Test plan

- [x] `dotnet build solutions/All.sln` — clean
- [x] `dotnet test test/Altinn.App.Core.Tests/` — 2533 pass (3 cleanup-specific tests removed)
- [x] `dotnet test test/Altinn.App.Api.Tests/` — 413 pass
- [x] Public-API snapshot (`PublicApiTests.PublicApi_ShouldNotChange_Unintentionally.verified.txt`) updated to reflect interface removal + `ProcessTaskInitializer` ctor change.
- [ ] Verify alongside the Storage PR in an integration environment that re-entering a task removes stale `generatedFromTask` data elements (PDFs, signatures) for both user and service-task transitions, and that the #1735 loop-to-EndEvent scenario leaves exactly one PDF.

## Breaking changes

`IProcessTaskCleaner` was a `public interface` in the `Internal/` namespace. A search across known app repos turned up zero references, so the practical breakage is nil — but technically this violates the "binary compatible within a major version" rule in AGENTS.md. Flagging for reviewer awareness.

## Summary by CodeRabbit

* **Refactor**
  * Internal process task handling reorganized: cleanup component removed and initializer/finalizer/data-locking wiring updated.

* **Behavioral Change**
  * Subform PDF generation no longer performs prior cleanup of previously generated artifacts.

* **Validation**
  * Local test validation now requires localtest API version 5 or higher (was 3).

* **Tests**
  * Tests updated to reflect removal of cleanup behavior and the version requirement change.

* **Chores**
  * CI pinned a tooling version used in integration tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Altinn/app-lib-dotnet/pull/1750?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ivarne (commented):
Approved (men ikke på ordentlig for denne kan ikke merges før https://github.com/Altinn/altinn-storage/pull/977 er deploya, og det tar fort litt tid)

martinothamar (approved):

PR #1763, opened 2026-05-22
Prepare for multi targeting net8 and net10

ivarne:
Note that Directory.Pacages.props needs some more work in order to reference correct versions for net8 and net10

## Related Issue(s)
- Fixes https://github.com/Altinn/altinn-studio/issues/18607
- See also https://github.com/Altinn/altinn-studio/pull/18793

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -36,7 +36,7 @@ jobs:
       uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
       with:
         dotnet-version: |
-          8.0.x
+          10.0.x
         include-prerelease: true
 
     - name: Checkout repository
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -27,8 +27,9 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
         with:
-          dotnet-version: |
-            8.0.x
+          dotnet-version:
+            - 8.0.x
+            - 10.0.x
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of analysis
--- a/.editorconfig
+++ b/.editorconfig
@@ -142,7 +142,12 @@ dotnet_diagnostic.CA1725.severity = warning
 dotnet_diagnostic.IDE0080.severity = error
 
 # CA1859: Use concrete types when possible for improved performance
-dotnet_diagnostic.CA1859.severity = warning
+# Keep disabled because this can narrow public or extension seams used by app code.
+dotnet_diagnostic.CA1859.severity = none
+
+# CA1873: Avoid evaluating expensive arguments when logging is disabled
+# Keep disabled because it is too noisy for the existing logging style after moving to the .NET 10 analyzer set.
+dotnet_diagnostic.CA1873.severity = none
 
 # CA1716: Rename namespace "" so that it no longer conflicts with the reserved language keyword 'Interface'
 # TODO: fixing this would be breaking
--- a/.github/workflows/dotnet-format.yml
+++ b/.github/workflows/dotnet-format.yml
@@ -12,7 +12,9 @@ jobs:
     - name: Setup .NET
       uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
       with:
-        dotnet-version: 8.0.x
+        dotnet-version:
+         - 8.0.x
+         - 10.0.x
     - name: Install csharpier
       run: dotnet tool restore
     - name: Run csharpier
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -28,8 +28,9 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
         with:
-          dotnet-version: |
-            8.0.x
+          dotnet-version:
+            - 8.0.x
+            - 10.0.x
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -31,8 +31,9 @@ jobs:
       - name: Install dotnet8
         uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
         with:
-          dotnet-version: |
-            8.0.x
+          dotnet-version:
+            - 8.0.x
+            - 10.0.x
       - name: Install deps
         run: |
           dotnet restore
--- a/.github/workflows/test-and-analyze-fork.yml
+++ b/.github/workflows/test-and-analyze-fork.yml
@@ -13,8 +13,9 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
         with:
-          dotnet-version: |
-            8.0.x
+          dotnet-version:
+            - 8.0.x
+            - 10.0.x
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
--- a/.github/workflows/test-and-analyze.yml
+++ b/.github/workflows/test-and-analyze.yml
@@ -17,8 +17,9 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
         with:
-          dotnet-version: |
-            8.0.x
+          dotnet-version:
+            - 8.0.x
+            - 10.0.x
       - name: Set up JDK 11
         uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
         with:
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -24,8 +24,6 @@
     <PackageVersion Include="JWTCookieAuthentication" Version="3.0.1" />
     <PackageVersion Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.23.0" />
     <PackageVersion Include="Microsoft.Build.Tasks.Core" Version="17.10.29" />
-    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="8.0.27" />
-    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="8.0.27" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.11.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.14.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.14.0" />
@@ -79,5 +77,15 @@
     <PackageVersion Include="KS.Fiks.IO.Client" Version="6.1.2" />
     <PackageVersion Include="Polly.Extensions" Version="8.6.6" />
     <PackageVersion Include="Polly.Testing" Version="8.6.6" />
+  </ItemGroup>
+
+  <!-- Separate dependency versions for net8 and net10-->
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="8.0.27" />
+    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="8.0.27" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' != 'net8.0'">
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="10.0.8" />
+    <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="10.0.8" />
   </ItemGroup>
 </Project>
--- a/benchmark/Altinn.App.Benchmarks/Altinn.App.Benchmarks.csproj
+++ b/benchmark/Altinn.App.Benchmarks/Altinn.App.Benchmarks.csproj
@@ -2,7 +2,7 @@
     <PropertyGroup>
         <OutputType>Exe</OutputType>
         <LangVersion>latest</LangVersion>
-        <TargetFramework>net8.0</TargetFramework>
+        <TargetFramework>net10.0</TargetFramework>
         <Nullable>disable</Nullable>
         <DebugType>pdbonly</DebugType>
         <DebugSymbols>true</DebugSymbols>
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "8.0.421",
+        "version": "10.0.300",
         "rollForward": "latestFeature",
         "allowPrerelease": false
     }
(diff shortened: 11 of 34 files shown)


ivarne-agent:
Copilot's review flagged that six workflows pass `actions/setup-dotnet`'s `dotnet-version` as a YAML list instead of the documented multiline-string form, plus that `codeql.yml` only installs the net10.0 SDK. I verified both are real (confirmed `dotnet-version` parses as a Python list vs. plain string via PyYAML before/after), but I don't have a token with `workflow` scope in my sandbox to push a commit that touches `.github/workflows/*`, so here's the diff to apply directly:

```diff
diff --git a/.github/workflows/codeql.yml b/.github/workflows/codeql.yml
index 5786d708..edbf9f9e 100644
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -36,6 +36,7 @@ jobs:
       uses: actions/setup-dotnet@9a946fdbd5fb07b82b2f5a4466058b876ab72bb2 # v5.3.0
       with:
         dotnet-version: |
+          8.0.x
           10.0.x
         include-prerelease: true
 
diff --git a/.github/workflows/dotnet-format.yml b/.github/workflows/dotnet-format.yml
index 37feef81..b22cbaea 100644
--- a/.github/workflows/dotnet-format.yml
+++ b/.github/workflows/dotnet-format.yml
@@ -12,9 +12,9 @@ jobs:
     - name: Setup .NET
       uses: actions/setup-dotnet@9a946fdbd5fb07b82b2f5a4466058b876ab72bb2 # v5.3.0
       with:
-        dotnet-version:
-         - 8.0.x
-         - 10.0.x
+        dotnet-version: |
+          8.0.x
+          10.0.x
     - name: Install csharpier
       run: dotnet tool restore
     - name: Run csharpier
diff --git a/.github/workflows/dotnet-test.yml b/.github/workflows/dotnet-test.yml
index 57a84679..d3593869 100644
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -31,9 +31,9 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@9a946fdbd5fb07b82b2f5a4466058b876ab72bb2 # v5.3.0
         with:
-          dotnet-version:
-            - 8.0.x
-            - 10.0.x
+          dotnet-version: |
+            8.0.x
+            10.0.x
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of analysis
diff --git a/.github/workflows/pr-actions.yml b/.github/workflows/pr-actions.yml
index 7dd156d8..7e9687af 100644
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -28,9 +28,9 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@9a946fdbd5fb07b82b2f5a4466058b876ab72bb2 # v5.3.0
         with:
-          dotnet-version:
-            - 8.0.x
-            - 10.0.x
+          dotnet-version: |
+            8.0.x
+            10.0.x
 
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
diff --git a/.github/workflows/publish-release.yml b/.github/workflows/publish-release.yml
index e195ad90..be806f27 100644
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -31,9 +31,9 @@ jobs:
       - name: Install dotnet8
         uses: actions/setup-dotnet@9a946fdbd5fb07b82b2f5a4466058b876ab72bb2 # v5.3.0
         with:
-          dotnet-version:
-            - 8.0.x
-            - 10.0.x
+          dotnet-version: |
+            8.0.x
+            10.0.x
       - name: Install deps
         run: |
           dotnet restore
diff --git a/.github/workflows/test-and-analyze-fork.yml b/.github/workflows/test-and-analyze-fork.yml
index 3c5aa70d..09da7db8 100644
--- a/.github/workflows/test-and-analyze-fork.yml
+++ b/.github/workflows/test-and-analyze-fork.yml
@@ -13,9 +13,9 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@9a946fdbd5fb07b82b2f5a4466058b876ab72bb2 # v5.3.0
         with:
-          dotnet-version:
-            - 8.0.x
-            - 10.0.x
+          dotnet-version: |
+            8.0.x
+            10.0.x
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
diff --git a/.github/workflows/test-and-analyze.yml b/.github/workflows/test-and-analyze.yml
index a627be05..a3db2457 100644
--- a/.github/workflows/test-and-analyze.yml
+++ b/.github/workflows/test-and-analyze.yml
@@ -17,9 +17,9 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@9a946fdbd5fb07b82b2f5a4466058b876ab72bb2 # v5.3.0
         with:
-          dotnet-version:
-            - 8.0.x
-            - 10.0.x
+          dotnet-version: |
+            8.0.x
+            10.0.x
       - name: Set up JDK 11
         uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5.2.0
         with:
```

.github/workflows/dotnet-test.yml
```diff
@@ -31,8 +31,9 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@9a946fdbd5fb07b82b2f5a4466058b876ab72bb2 # v5.3.0
         with:
-          dotnet-version: |
-            8.0.x
+          dotnet-version:
+            - 8.0.x
+            - 10.0.x
```
ivarne-agent: This one wasn't included in the batch of Copilot suggestions you applied — same fix as the other five workflows.

```suggestion
          dotnet-version: |
            8.0.x
            10.0.x
```

.github/workflows/codeql.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       uses: actions/setup-dotnet@9a946fdbd5fb07b82b2f5a4466058b876ab72bb2 # v5.3.0
       with:
         dotnet-version: |
-          8.0.x
+          10.0.x
```
ivarne-agent: Copilot's other finding here: this workflow only installs .NET 10, but the repo still multi-targets net8.0 too — CodeQL's autobuild can fail without the net8 targeting pack.

```suggestion
          8.0.x
          10.0.x
```

ivarne-agent:
Found and root-caused the Api.Tests failures from the last CI run - it's a real bug, not flakiness, and it's separate from the Analyzers.Tests issue (see #1850).

**Root cause**: `Altinn.App.Api.Tests` now multi-targets `net8.0;net10.0`. Its test fixtures resolve on-disk test data via `[CallerFilePath]` (`TestData.cs`), which always points at the same absolute source path regardless of which TFM built the test assembly. `dotnet test solutions/All.sln` runs every project/TFM's tests in parallel by default, so the net8.0 and net10.0 copies of `Api.Tests` end up mutating (writing/moving/deleting) the *same* on-disk instance JSON fixtures concurrently - causing the `FileNotFoundException` / unexpected `409 Conflict` failures on tests named around "mutates data" that showed up in the CI run.

I confirmed this locally: running `dotnet test solutions/All.sln` without `-m:1` produces dozens of sporadic failures, including in `Altinn.App.Clients.Fiks.Tests` which has no net8/net10 split at all (pure resource contention from too many parallel test hosts). With `-m:1` added, only one pre-existing, already-flaky, unrelated symlink-watcher test remains, reproduced twice for stability.

I don't have `workflow` scope to push this (same limitation as the earlier `dotnet-version` fix), so here's the diff - it just adds `-m:1` to every `dotnet test solutions/All.sln` invocation, forcing sequential project/TFM test execution. It doesn't touch the `dotnet build` steps, which stay parallel since building doesn't mutate shared runtime fixtures:

```diff
diff --git a/.github/workflows/dotnet-test.yml b/.github/workflows/dotnet-test.yml
index d3593869..XXXXXXXX 100644
--- a/.github/workflows/dotnet-test.yml
+++ b/.github/workflows/dotnet-test.yml
@@ -46,12 +46,12 @@ jobs:
       - name: Test
         if: matrix.os != 'ubuntu-latest'
         run: |
-          dotnet test solutions/All.sln -v m --no-restore --no-build --filter "Category!=Integration"
+          dotnet test solutions/All.sln -v m -m:1 --no-restore --no-build --filter "Category!=Integration"
       - name: Test (main tests)
         if: matrix.os == 'ubuntu-latest'
         run: |
-          dotnet test solutions/All.sln -v m --no-restore --no-build --filter "Category!=Integration"
+          dotnet test solutions/All.sln -v m -m:1 --no-restore --no-build --filter "Category!=Integration"
       - name: Install studioctl
         if: matrix.os == 'ubuntu-latest'
         env:
@@ -69,4 +69,4 @@ jobs:
           STUDIOCTL_HOME: ${{ runner.temp }}/studioctl-home
           TEST_STUDIOCTL_COMMAND: ${{ runner.temp }}/studioctl-bin/studioctl
         run: |
-          dotnet test solutions/All.sln --no-restore --no-build --logger "console;verbosity=detailed" --filter "Category=Integration"
+          dotnet test solutions/All.sln -m:1 --no-restore --no-build --logger "console;verbosity=detailed" --filter "Category=Integration"
diff --git a/.github/workflows/test-and-analyze.yml b/.github/workflows/test-and-analyze.yml
index a3db2457..XXXXXXXX 100644
--- a/.github/workflows/test-and-analyze.yml
+++ b/.github/workflows/test-and-analyze.yml
@@ -56,6 +56,7 @@ jobs:
 
           dotnet build solutions/All.sln
           dotnet test solutions/All.sln `
+          -m:1 `
           --no-restore --no-build `
           --results-directory TestResults/ `
           --collect:"XPlat Code Coverage" `
diff --git a/.github/workflows/test-and-analyze-fork.yml b/.github/workflows/test-and-analyze-fork.yml
index 09da7db8..XXXXXXXX 100644
--- a/.github/workflows/test-and-analyze-fork.yml
+++ b/.github/workflows/test-and-analyze-fork.yml
@@ -24,7 +24,7 @@ jobs:
         run: dotnet build solutions/All.sln -v m
 
       - name: dotnet test
-        run: dotnet test solutions/All.sln --no-restore --no-build --results-directory TestResults/ --collect:"XPlat Code Coverage" --filter "Category!=Integration" -v m
+        run: dotnet test solutions/All.sln -m:1 --no-restore --no-build --results-directory TestResults/ --collect:"XPlat Code Coverage" --filter "Category!=Integration" -v m
```

ivarne-agent:
Continuing from where I left off — two updates:

**1. Directory.Packages.props audit**

Ran a full `dotnet restore` and `dotnet build -c Release` of `AppLibDotnet.sln` against the current branch tip (`ceddfa2d`). Both net8.0 and net10.0 come back clean — zero `NU16xx` version-conflict/downgrade warnings, zero build errors on either TFM. The only warning present is the pre-existing, unrelated `NU1902` advisory on the transitive `Scriban.Signed` 7.2.0 package (via `Verify.Xunit`), which predates this PR.

The earlier batch of fixes already merged into this branch (Humanizer pin to 3.0.1, `Microsoft.Extensions.Diagnostics.Testing`/`Microsoft.Extensions.TimeProvider.Testing` → 10.9.0, `Microsoft.FeatureManagement.AspNetCore` → 4.7.0, and the net8/net10 conditional `ItemGroup` split for `Microsoft.AspNetCore.Mvc.Testing`/`Microsoft.AspNetCore.TestHost`) appear to have closed the gap the PR description flags ("Directory.Pacages.props needs some more work"). I didn't find any further TFM-conditional rules needed — I don't think there's anything left to fix here. Happy to take another pass if you know of a specific package/version pair that's still wrong.

**2. Instance test-data race — fixed, opened as #1851**

The CI failures on this PR (`FileNotFoundException` on `Instances/*.json`, tests named around "mutates data") are the race I described earlier: two TFMs' `Api.Tests` processes mutating the same on-disk fixture path concurrently. Rather than the `-m:1` CI-serialization workaround I posted earlier (still blocked — needs `workflow` scope I don't have), I implemented the underlying fix: each test process now gets its own private copy of the `Instances` tree. Verified locally: before the fix, running both TFMs concurrently produced 101 failures; after, only the 1 pre-existing, already-flaky, unrelated symlink-watcher test remains (on each TFM).

Opened as #1851 against this branch, since I only have push access via my own fork (no direct write access to this repo, and my fork's history doesn't have `workflow` scope either, so it's based a few commits back to avoid carrying workflow-file diffs — the resulting diff is just the one test file).

**Still open / unresolved (from earlier in this session):**
- CI workflows use `dotnet-version:` as a YAML list in six workflow files instead of the documented multiline-string form (only `codeql.yml` uses the multiline form, and only installs net10.0, missing net8.0 entirely). I posted the diff as a comment above; still needs someone with `workflow` scope to apply it directly.
- SonarCloud Quality Gate is failing on "Maintainability Rating on New Code" — mostly pre-existing complexity/param-count issues in `LayoutEvaluator.cs`, `DataTagsController.cs`, and `FiksArkivHost.cs` that get pulled into Sonar's "new code" scope because this PR touches those files. Not introduced by this PR's actual diff as far as I can tell, but will keep failing the gate until addressed or the gate is overridden.
- `Altinn.App.Analyzers.Tests` failures are a separate, already-tracked issue (#1850), unrelated to the fixes above.

ivarne-agent:
Reviewed the two Copilot comments from this morning — both check out:

- `Altinn.App.Api.csproj:60` — correct, the `Exists()` guard is check-then-act and doesn't actually make the rename safe under MSBuild's default parallel inner-TFM builds.
- `TestData.cs:64` — correct, the "disposable" copy from #1851 was never actually deleted.

Fixed both in #1852 (also keyed the temp `Instances` copy by TFM instead of a fresh Guid per run, per feedback, so repeated local runs reuse/reset a bounded set of directories instead of accumulating — a better fix than just deleting on exit, since it also caps disk usage from failed runs where teardown might not run).

ivarne-agent:
Following up on the [`TestData.cs:64`](https://github.com/Altinn/app-lib-dotnet/pull/1763#discussion_r3901952699) Copilot comment (the other half of #1852, which now only covers the `Api.csproj` race - see that PR for why I split it out).

Verified: with this change, running both TFMs concurrently repeatedly leaves exactly 2 directories (`Instances-net8.0`, `Instances-net10.0`) under the temp dir instead of growing unbounded, and a stale/mutated leftover file from a prior (e.g. failed) run gets wiped on the next run rather than persisting. Same pass rate as #1851 otherwise (only the pre-existing flaky symlink-watcher test fails).

I can't push this one directly - my fork-based token can only reach commits far enough back in history to avoid `.github/workflows/*`-touching commits, and this fix builds directly on top of #1851's actual merge commit (which sits after those). Diff, ready to apply on top of the current branch tip:

```diff
diff --git a/test/Altinn.App.Api.Tests/Data/TestData.cs b/test/Altinn.App.Api.Tests/Data/TestData.cs
index 5db29bfc..1011086a 100644
--- a/test/Altinn.App.Api.Tests/Data/TestData.cs
+++ b/test/Altinn.App.Api.Tests/Data/TestData.cs
@@ -59,14 +59,29 @@ public static class TestData
     // Instance data is mutated in place by tests (created, patched, deleted). Altinn.App.Api.Tests now
     // multi-targets net8.0;net10.0, and both builds share the exact same [CallerFilePath]-derived source
     // path, so running both TFMs concurrently (as `dotnet test solutions/All.sln` does by default) would
-    // otherwise race on the same on-disk files. Give each test *process* its own private, disposable copy
-    // of the Instances tree instead of mutating the checked-in fixtures directly.
+    // otherwise race on the same on-disk files. Give each TFM its own private copy of the Instances tree,
+    // keyed by TFM instead of a fresh Guid per run, and reset it on every run, so repeated (including
+    // failed) local runs reuse the same directory instead of filling up the disk with orphaned copies.
     private static readonly Lazy<string> _isolatedInstancesDirectory = new(CreateIsolatedInstancesDirectory);
 
+#if NET8_0
+    private const string TargetFrameworkMoniker = "net8.0";
+#else
+    private const string TargetFrameworkMoniker = "net10.0";
+#endif
+
     private static string CreateIsolatedInstancesDirectory()
     {
         string sourceDirectory = Path.Join(GetTestDataRootDirectory(), "Instances");
-        string targetDirectory = Path.Join(Path.GetTempPath(), "AltinnAppApiTests", $"Instances-{Guid.NewGuid()}");
+        string targetDirectory = Path.Join(
+            Path.GetTempPath(),
+            "AltinnAppApiTests",
+            $"Instances-{TargetFrameworkMoniker}"
+        );
+        if (Directory.Exists(targetDirectory))
+        {
+            Directory.Delete(targetDirectory, recursive: true);
+        }
         CopyDirectory(sourceDirectory, targetDirectory);
         return targetDirectory;
     }

```

danielskovli (approved):

PR #1772, opened 2026-05-27
Deprecate sender and use get binary data stream

MagnusHAa:
## Description
The sender field is no longer used in the correspondence api as this is now set determined from the resource registry. The WithSender methods and the Sender field has been marked as obsolete. Additionally the CorrespondenceBuilder no longer sends the sender parameter to the correspondence api. Using the WithSender method is still possible but it maps to nothing in the end. Made WithSender extend WithSendersReference so that integrations can skip the now deprecated Sender when building a request.

Additionally, I made the SigningReceiptService use GetBinaryDataStream instead of GetDataBytes so that the data isn't loaded into memory in the app, but streamed all the way through from Storage.

## Related Issue(s)
- #https://github.com/Altinn/altinn-correspondence/issues/1367

## Verification
- [x] All tests run green

## Documentation

## Summary by CodeRabbit

* **Changes**
  * Sender is now resolved automatically from the resource registry via the resource ID; explicit sender configuration is deprecated.
  * Correspondence payloads and notifications no longer include or rely on an explicitly set sender; all other correspondence behavior (references, recipients, content, attachments, notifications) remains unchanged.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Altinn/app-lib-dotnet/pull/1772?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceRequestBuilder.cs
```diff
@@ -44,16 +43,14 @@ public ICorrespondenceRequestBuilderSender WithResourceId(string resourceId)
     /// <inheritdoc/>
     public ICorrespondenceRequestBuilderSendersReference WithSender(OrganisationNumber sender)
```
ivarne, line 47: Should preferably copy the `[Obsolete]` message here as well, so that it doesn't get missed when we clean this up.

ivarne: Or maybe an explicit implementation is better, so that it definitely gets removed.

ivarne (approved):

src/Altinn.App.Core/Features/Correspondence/Builder/CorrespondenceRequestBuilder.cs
MagnusHAa: I'll add the Obsolete message to the builder so that end-users know not to use WithSender. This also keeps it consistent with the other Obsolete methods for correspondence.

ivarne (approved):

PR #1788, opened 2026-06-01
Support all data model bindings in expressions

TomasEng:
## Description
Currently, the `component` expressions supports components with `simpleBinding` only. Now that we support lists and objects, it makes sense to support `group` bindings and components with several bindings as well.

This pull requests adds the following bahaviour to the `component` expression:
- When referring to a group component, the value referred in the `group` binding of the component of is returned
- When referring to a component that is not a group and that does not have a `simpleBinding`, an object containing all the bindings is returned

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/issues/4297

## Verification
- Manual testing not possible since there is no use case for this until we add functions that accept objects as parameters; the automated unit tests should be sufficient
- [x] All tests run green

## Documentation


ivarne:
> When referring to a group component, the value referred in the group binding of the component of is returned
When referring to a component that is not a group and that does not have a simpleBinding, an object containing all the bindings is returned

I think this is the wrong semantics for `["component"]`. The main benefit of using `["component"]` is that it evaluates `"hidden"` expressions, and for groups (at a minimum), that would include `"hiddenRow"` on the component.

The `"simpleBinding"` restriction looks somewhat silly in retrospect, and the thinking was that we might do support an explicit binding with something like `["component", "componentId", "binding"]`.

I think groups deserve a special handling where we actually first filter by `"hiddenRow"` and then return a list of objects that represent each component in the group, where each component is an object of the bindings and values of that component (or null if the component is hidden)
```json
// return value from `["component", "groupComponentId"]`
[
  {
      "component1":{
         "simpleBinding": "value in component 1"
      },
      "component2":{
        "simpleBinding": true,
        "label": "ja"
      }
  },
  {
     "component1": {...}
     "component2": {...}
  }
]
```
A few questions we need to decide on

1. When a row is hidden with "hiddenRow", should the resulting list contain `null`s (to keep indexes consistent), or just be filtered out? (probably filter, because it is important for correct count/average, and consistent indexes are less important here.)
2. When a component is hidden (in a visible row), should it's ID be removed, it binding object be empty, or each binding evaluate to `null`?
3. Should it be optional to get an object for `{"simpleBinding": "value"}` or just the direct `"value"` to follow the current semantics or is group support so different that we now can do just objects?
4. How should recursive groups look?

test/Altinn.App.Core.Tests/LayoutExpressions/CommonTests/shared-tests/functions/component/lookup-group.json
```diff
@@ -0,0 +1,65 @@
...
+      "alderIAar": 40
+    },
+    {
+      "navn": "Kari",
+      "alderIAar": 35
+    },
+    {
+      "navn": "Knut",
+      "alderIAar": 17
+    }
```
ivarne, line 16: Litt uheldig for denne testen at datamodellen bruker samme nøkler som id i komponenter. For å teste at det brukes rett id, bør de være ulike.

Altså bør ikke komponenten som binder til `mennesker.navn` ha id `navn`.

Tester av `component` bør også verifisere oppførselen til
* `hidden` på komponenter
* `hiddenRow` på rader,
* flere `dataModelBindings` (adresse/options) 
* rekursiv repeterende gruppe gruppe, 
* ikke repeterende gruppe (accordion, table..)

TomasEng: > Litt uheldig for denne testen at datamodellen bruker samme nøkler som id i komponenter. For å teste at det brukes rett id, bør de være ulike.

Dette blir verifisert med `alder`-komponenten, som peker på `mennesker.alderIAar`. Men testen må uansett endres siden den går ut fra at vi skal returnere hele datamodellobjektet som `group` refererer til, og ikke et objekt med komponent-ID-er slik vi nå har blitt enige om.

PR #1804, opened 2026-06-10
Fix 500 in payment status read when a concurrent webhook locks the data element

Magnusrm:
### Problem

`GET …/payment` (`GetPaymentInformation`) intermittently returned **500** after a successful payment.

The endpoint persists the latest payment status on read when the requested task is the current task. This races with the payment webhook, which advances the process and **locks** the payment data element. The persist `PUT` to storage then fails with `409 - "data element <id> is locked and cannot be updated"`, which surfaced to the user as a 500 — typically right after returning from the hosted payment page.

### Why the existing mitigation missed it

`GetPaymentInformation` already guarded this race: on a persist failure it called `CurrentTaskMovedAwayFrom` and only degraded to a read-only result if it could **confirm the current task had moved**. But the data-element lock can become visible *before* the `CurrentTask` change does, so in that window the re-check still saw the payment task as current → the 409 was rethrown → 500. This made it intermittent ("some users").

### Fix

Treat a `409/Conflict` from t back to read-only paymentstatus, **without** requiring the task-moved check — the lock itself is the   authoritative signal that a c) already finalized things.Non-409 exceptions keep their previous behavior (still propagate when the taskmoved).
                                                                              
### Testing
Adds a regression test,`GetPaymentInformation_PersistFailsWithLockedDataElement_FallsBackToReadOnly`,reproduces the production scel storage needed): the persist throws `PlatformHttpException(409, "…is locked…")` with the current task left **unchanged**, and the endpoithe read-only status. Theexisting `PersistFailsButTaskUnchanged_PropagatesException` test confirms non-409 errors
still propagate.

Frontend PR adding retries in order to further harden the process-flow:
https://github.com/Altinn/app-frontend-react/pull/4259

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


ivarne (approved):
Looks very reasonable!

Maybe @vxkc could also have a look?

vxkc (approved):

PR #1812, opened 2026-06-16
Fix signing notification for signees with multiple identifiers

danielskovli:
## Description

Fixes [#19153](https://github.com/Altinn/altinn-studio/issues/19153). A signee with more than one notification identifier (e.g. SSN + email + mobile) caused the Correspondence API to reject the call with `Custom recipient with multiple identifiers is not allowed`, so **no notification was sent** and signees saw _"Varsling mislyktes"_.

The Correspondence API now requires **exactly one identifier per custom recipient**, with multiple recipients forming a list (`customRecipients`); the signing code packed email + mobile + the org/person identifier into a single recipient. This PR fixes that and brings the correspondence client up to date with the `customRecipients` / `overrideRegisteredContactInformation` surface (also addresses [#17407](https://github.com/Altinn/altinn-studio/issues/17407)).

## Notification semantics: override, restored correctly

The signing `Email.EmailAddress` / `Sms.MobileNumber` fields have always been documented as **overrides** — an app-supplied address *replaces* the signee's registered (KRR) contact. That behavior never actually worked: the original mechanism (singular `customRecipient` + registry enrichment, no override flag) couldn't express "replace" against the Correspondence API, and custom recipients are additive by default — so the signee was effectively notified twice.

This PR restores the **intended** override behavior using the API's purpose-built `overrideRegisteredContactInformation` flag ([altinn-correspondence#1325](https://github.com/Altinn/altinn-correspondence/issues/1325)):

- When the app supplies explicit contact info, we set `overrideRegisteredContactInformation = true` → only the custom recipients are notified (KRR contact is suppressed). The signee is notified **once**, at the override address.
- When no explicit contact is supplied, no override is sent and the notification falls back to the default correspondence recipient (resolved via KRR) — exactly as before.

## Changes

**Correspondence client / builder / model**
- Add `CustomRecipients` (list) and `OverrideRegisteredContactInformation` to `CorrespondenceNotification` and the request DTO (`customRecipients` / `overrideRegisteredContactInformation` wire fields).
- Builder methods, each setting both the recipient list and the override flag so the last call fully wins:
  - `WithCustomRecipients(...)` — additive: KRR contact **and** the custom recipients are notified (`overrideRegisteredContactInformation = false`).
  - `WithRecipientOverrides(...)` — override: **only** the custom recipients are notified (`overrideRegisteredContactInformation = true`).
  - `*IfConfigured` variants for both (null/empty is a no-op, so the API's "flag requires recipients" constraint is never violated).
- `CorrespondenceClient` **explodes** every custom recipient (new plural `CustomRecipients` and legacy singular `CustomRecipient`) into one wire entry per identifier, so the multiple-identifier error is structurally impossible for any caller — not just the signing path. The deprecated singular `customRecipient` wire field is no longer emitted.
- Add `EmailAndSms` to `CorrespondenceNotificationChannel` (matches the API enum).
- Deprecate the singular `CustomRecipient`, `CustomNotificationRecipients`, and the `WithRecipientOverride` family in favour of the new surface, formalizing the `CustomRecipients` term per #17407.

**Signing**
- Emit one single-identifier recipient per channel and override the registered contact information.
- Map the `SmsAndEmail` choice to the new `EmailAndSms` channel. Previously it mapped to `EmailPreferred` ("prefer email, fall back to SMS"), which only delivered on one channel instead of both.
- Registry contact-info enrichment in `SigneeContextsManager` is **intentionally retained** — it is established, published behaviour and lets an app declare an `Email`/`Sms` block without an address and have the contact resolved from the registry. Combined with the override flag, the signee is still notified exactly once.

## Backwards compatibility

All changes are additive — new members plus `[Obsolete]` annotations, no removals. Binary compatible; verified via the public-API snapshot test.

## Related
- Closes https://github.com/Altinn/altinn-studio/issues/19153
- Closes https://github.com/Altinn/altinn-studio/issues/17407

## Tests

- Regression test for #19153: a signee with both email and mobile produces one single-identifier recipient per channel, and `SmsAndEmail` maps to `EmailAndSms`.
- Theory covering `Email` / `Sms` / `SmsPreferred` / `EmailPreferred` recipient counts and the override flag; `None` asserts no custom recipients and no override.
- Mapping tests for `WithRecipientOverrides` (override flag true), `WithCustomRecipients` (additive, flag false), and the legacy singular override being folded + exploded into `customRecipients`.

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying multiple custom recipients for correspondence notifications.
  * Added an option to control whether registered contact information is overridden when custom recipients are used.
  * Introduced a new `EmailAndSms` notification channel option (send via both email and SMS).

* **Deprecations**
  * Deprecated single-recipient recipient-override methods in favor of the multiple-recipients API.
  * Marked legacy recipient override builders as deprecated and removed the older wrapper-based overloads.

* **Bug Fixes**
  * Updated correspondence recipient override mapping to consistently use the plural recipients representation.

* **Tests**
  * Expanded/updated recipient override and channel mapping coverage to match the new contract.


HauklandJ (approved):
LGTM

